PR #7, opened 2017-07-20
Updates to setup script

jessieay:
* Now, running setup script does not result in any errors when running citizen
  doctor script.


bin/setup
```diff
@@ -10,13 +10,15 @@ chdir APP_ROOT do
...
     system('bundle check') || system!('bundle install')
+    system('which direnv') || system!('brew install direnv')
   end
 
   step 'Copying sample files' do
     ['.envrc'].each do |file|
       unless File.exist?(file)
         cp "#{file}.sample", file
         puts "Copied #{file}.sample to #{file}"
+        system! "ln -s #{file} .env"
```
jessieay: question for Citizen Code folks: why do we have an `envrc.sample` file rather than `env` ?

eahanson: Some people like to use `direnv` and run `rails s` which requires `.envrc`, and some people like to use `heroku local` which reads `.env`, so the compromise is to have one be symlinked to the other.

jessieay: ahh I see. `citizen doctor` told me that I needed `direnv`, so if we want to stick with that perhaps we stay with `.envrc.sample` but also change the README to tell people to run `rails s` instead of `heroku local`?

adarsh:
Looks good to me! Nice changes. 👍

adarsh:
[Update setup script](https://trello.com/c/xeC95jv3/14-update-setup-script)

PR #10, opened 2017-07-24
Remove form-related code

jessieay:
**WHY**: A long long time ago, we thought this web application would create a PDF
and email it in order to submit an app. Now, we are going to use
selenium to fill in web forms instead.


jessieay:
[Remove the /form stuff](https://trello.com/c/BCRn7TMs/16-remove-the-form-stuff)

jessieay:
@adarsh in light of this morning's meeting, should I update this to remove the mailer stuff but keep the PDF generation part? Or just remove all and re-build as we create the fax behavior.

adarsh:
@jessieay 

> should I update this to remove the mailer stuff but keep the PDF generation part? Or just remove all and re-build as we create the fax behavior.

I'm going to guess you're right here - remove mailer stuff and keep PDF generation but I trust your judgement.

jessieay:
@adarsh alrighty, mailer stuff is now gone but form stuff stays! see 710418b for diff

PR #12, opened 2017-07-24
Remove unnecessary dependencies

jessieay:
**WHY**: This is mostly dev-only stuff that Cylinder folks won't use.
Plus, better not to have dev-specific stuff in the Gemfile when
possible.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ group :development, :test do
   gem 'haml-lint', require: false
   gem 'poltergeist'
   gem 'pry-rails'
+  gem 'rspec_junit_formatter'
   gem 'rspec-rails'
   gem 'rubocop-rspec', require: false
 end
--- a/Gemfile
+++ b/Gemfile
@@ -4,12 +4,9 @@ source 'https://rubygems.org'
 
 ruby '2.4.1'
 
-gem 'awesome_print'
 gem 'aws-sdk'
 gem 'bourbon', '~> 4.2.0' # to keep in sync with getcalfresh
-gem 'coffee-rails'
 gem 'haml', '~> 5.0'
-gem 'jbuilder', '~> 2.0'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -28,48 +24,23 @@ gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
...
   gem 'capybara-accessible'
-  gem 'capybara-screenshot'
   gem 'climate_control'
   gem 'haml-lint', require: false
-  gem 'launchy'
   gem 'poltergeist'
   gem 'pry-rails'
   gem 'rspec-rails'
-  gem 'rspec_junit_formatter'
-  gem 'rubocop', require: false
```
adarsh: I'm assuming we don't need both this and `rubocop-rails`? 🚔

Gemfile
```diff
@@ -28,48 +24,23 @@ gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
...
 group :development do
   gem 'citizen-scripts', git: 'https://github.com/citizencode/citizen-scripts'
-  gem 'guard', require: false
-  gem 'guard-rspec', require: false
-  gem 'listen', '~> 3.0.5'
-  gem 'meta_request'
-  gem 'pivotal_git_scripts'
-  gem 'pry-rescue'
-  gem 'pry-stack_explorer'
-  gem 'spring'
```
adarsh: 🌸

Gemfile
jessieay: yep! `rubocop-rails` includes this

jessieay:
@adarsh I think that if we add the hakiri webook, which getcalfresh uses, that will run `bundler-audit` (or something similar). We have a card for that here: https://trello.com/c/ZPEvNFUi/30-add-hakiri-for-security

PR #15, opened 2017-07-24
Green Rubocop build

jessieay:
**WHY**: So we can run rubocop as part of the default Rake task.

NOTE that I removed the line length cop because there were 185
violations and that didnt seem like a good use of time. We can fix those
as we go and add back in later?


jessieay:
[Add rubocop to default rake task](https://trello.com/c/EUkGAvSP/36-add-rubocop-to-default-rake-task)

.rubocop.yml
```diff
@@ -209,6 +210,7 @@ Metrics/LineLength:
   Description: 'Limit lines to 80 characters.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#80-character-limits'
   Max: 80
+  Enabled: false
```
adarsh: What about a longer hard limit like 100 chars?

jessieay: that still resulted in many errors. I am totally on board with doing it, just felt like a lot of work for today when there were other cleanup tasks to tackle. 

I think a good strategy that ensures we actually do this eventually would be to add this back in as its own task: https://trello.com/c/8MHQSLAV

adarsh: Sounds good, I like it. 👍

PR #16, opened 2017-07-24
Re-add listen gem

jessieay:
**WHY**: used to look for file changes in the dev environment


adarsh:
Thanks @jessieay.

For posterity, I was seeing this error on my machine:

```
LoadError: Could not load the 'listen' gem. Add `gem 'listen'` to the development group of your Gemfile
```
https://stackoverflow.com/questions/38663706/loaderror-could-not-load-the-listen-gem-rails-5

PR #17, opened 2017-07-25
Remove form class and controller

jessieay:
**WHY**: We are going to be filling out an existing PDF, not creating
our own.


jessieay:
Circle is failing here for reasons fixed in https://github.com/codeforamerica/michigan-benefits/pull/18

PR #18, opened 2017-07-25
Update circle config for codeclimate test reporter

jessieay:
**WHY**: Was getting "Coverage results not found" error


jessieay:
I think this is how code climate references env vars...but am not 100% sure. Trying to copy setup from GCF.

PR #22, opened 2017-07-26
Create views and flow for minimal SNAP application

jessieay:
**WHY**: This is an MVP. Next step will be to hook up filling out of PDF
and fax this.

To test:
* Visit root path
* Click "Apply now"
* Fill in all fields (all are required)
* Submit
* See success message on root path
* There should be a "MbSnapApp" record in the DB based on your entry


Gemfile
```diff
@@ -32,7 +32,6 @@ end
 
 group :development, :test do
   gem "capybara"
-  gem "capybara-accessible"
```
jessieay: I removed this because I copied form helper methods from GCF which were raising accessibility issues. I think we should re-add this and fix this, but it felt out of scope for this. Created a card here:  https://trello.com/c/4zn71DZM

jessieay:
[When I apply for SNAP via the prototype, I want the agency to receive and register my application, so that I don't have to start over.](https://trello.com/c/0LcvwkPP/43-when-i-apply-for-snap-via-the-prototype-i-want-the-agency-to-receive-and-register-my-application-so-that-i-dont-have-to-start-ov)

app/helpers/mb_form_builder.rb
```diff
@@ -0,0 +1,71 @@
+class MbFormBuilder < ActionView::Helpers::FormBuilder
+  def mb_input_field(method, label_text, type: "text", notes: [], options: {}, classes: [], prefix: nil, autofocus: nil)
```
jessieay: these were copied from https://github.com/codeforamerica/gcf-backend/blob/master/app/helpers/gcf_form_builder.rb FYI

app/views/shared/_flashes.html.erb
```diff
@@ -0,0 +1,17 @@
+<% if notice %>
```
jessieay: copied this from GCF, too

config/routes.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
-  root "users#new"
```
jessieay: kept `users#new` so that the existing feature spec will still pass, but if we are going to continue on this path I will remove it

app/controllers/application_controller.rb
```diff
@@ -56,6 +56,6 @@ def current_path(*); end
   def next_path(*); end
 
   def current_app
-    @current_app ||= App.find_or_create_by!(user: current_user)
+    @current_app ||= MbSnapApp.find_or_create_by!(user: current_user)
```
adarsh: Naming: Mb is "Michigan Bridges"? The whole application is for Michigan, so perhaps it's not necessary to specify that here?

Other ideas: `SnapApplication`, `SnapApp`, `MichiganSnapApp`

app/controllers/introduction_introduce_yourself_controller.rb
```diff
@@ -3,15 +3,15 @@
 class IntroductionIntroduceYourselfController < StepsController
```
adarsh: Naming: `IntroductionIntroduceYourselfController` is a little unwieldy, although I don't have enough context to suggest something clearer.

app/controllers/sign_and_submit_controller.rb
```diff
@@ -5,17 +5,11 @@ def update
     @step = step_class.new(step_params)
 
     if @step.valid?
-      current_app.update!(step_params)
-      send_submission_sms
-      redirect_to(next_path)
+      current_app.update(step_params)
+      flash[:notice] = "Your application has been successfully submitted."
```
adarsh: Probably not in this PR, but we should come up with an I18n plan at some point. (e.g. "I18n every string we touch" or "don't bother until we get to internationalization, when we can do all of it at once")

adarsh: Created a card here: https://trello.com/c/bC3XF2Eu/3-think-about-internationalization-strategy

app/helpers/mb_form_builder.rb
```diff
@@ -0,0 +1,71 @@
+class MbFormBuilder < ActionView::Helpers::FormBuilder
```
adarsh: Naming: Prefixing classes with `Mb` seems reasonable, although the acronym might not be clear to the next developer. Also, isn't everything in the app is for MichiganBridges? I'm probably missing something but just trying to understand the scheme.

app/helpers/mb_form_builder.rb
```diff
@@ -0,0 +1,71 @@
...
+    <<-HTML.html_safe
+      <fieldset class="form-group#{error_state(object, method)}">
+        #{label(method, label_contents(label_text, notes))}
+        <div class="input-group--inline">
+          <div class="select">
+            #{date_select(method, { autofocus: autofocus, date_separator: '</div><div class="select">' }.merge(options), class: 'select__element')}
+          </div>
+        </div>
+        #{errors_for(object, method)}
+      </fieldset>
```
adarsh: Interesting to have this much markup in the form helper. I'm sure it's fine for now but we should keep an eye on this.

app/models/concerns/multiparameter_attribute_assignment.rb
```diff
@@ -0,0 +1,202 @@
...
+  def initialize(params = {})
+    assign_attributes(params)
+  end
+
+  def assign_attributes(new_attributes)
+    multi_parameter_attributes = []
+
+    attributes = sanitize_for_mass_assignment(new_attributes.stringify_keys)
+
+    attributes.each do |k, v|
```
adarsh: Style: single-letter iterator values. Maybe `key` and `value`? or `attribute` and `value`? Not a huge deal.

app/steps/introduction_contact_information.rb
```diff
@@ -1,36 +1,27 @@
 # frozen_string_literal: true
 
 class IntroductionContactInformation < Step
```
adarsh: Naming: I wonder if this could just be `ContactInformation`?

app/steps/introduction_contact_information.rb
```diff
@@ -1,36 +1,27 @@
...
 
-  before_validation :strip_non_digits_from_phone_number
+  validates :state,
+    presence: { message: "Make sure to provide a state" }
 
-  def strip_non_digits_from_phone_number
-    self.phone_number = phone_number.remove(/[^\d]/)
-  end
+  validates :zip,
+    length: { is: 5, message: "Make sure your ZIP code is 5 digits long" },
```
adarsh: We don't want to validate this is an integer, correct? Just wanted to make sure that was part of our permissive data plan.

app/views/introduction_contact_information/edit.html.haml
```diff
@@ -11,48 +11,35 @@
...
+            placeholder: 'MI'
           },
           {
-            title: 'ZIP Code',
+            title: 'ZIP code',
             type: :text_field,
-            method: :mailing_zip,
-            placeholder: 'ZIP'
+            method: :zip,
+            placeholder: '12345'
```
adarsh: Minor: Flint City Hall's zip code is `48502` - could use that instead

app/views/sign_and_submit/edit.html.haml
```diff
@@ -11,8 +11,7 @@
     .form-card__content
       %p
         By entering your name you agree that you want to apply for
-        FAB and Medicaid, that you have been honest on this application, and
-        that you have read and agreed to the terms on the previous page.
+        FAB, that you have been honest on this application.
```
adarsh: I assume the user will know what "FAB" is (I do not). If that's in question, maybe we can link this text to somewhere that explains it?

app/views/shared/_flashes.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<% end %>
+
+<% if alert %>
+<div class="flash-container" id="dismiss-notice">
+  <div class="flash flash--error">
+    <p class="flash__message"><%= alert %></p>
+    <a href="#dismiss-notice" class="flash__dismiss"><span class="icon icon-close"></span></a>
+  </div>
+</div>
+<% end %>
```
adarsh: Looks good - I forgot if we standardized on ERB or HAML?

app/views/users/new.html.haml
```diff
@@ -12,49 +12,20 @@
 
 #apply-for-programs.slab.slab--white
   .card--narrow
-    %h3.h3--strong 
-      Choose the programs you want to apply for today.
-
-    %p.text--help.text--padded
-      You can choose more than one
+    = render 'shared/flashes'
```
adarsh: Should this go into a layout instead? No worries if that's premature.

spec/controllers/sign_and_submit_controller_spec.rb
```diff
@@ -4,19 +4,11 @@
...
-      ),
-    )
+    MbSnapApp.create!(attributes.merge(user: member))
   end
 
   let(:attributes) do
-    {
-      signature: "Hans Solo",
-    }.with_indifferent_access
+    { signature: "Hans Solo" }.with_indifferent_access
```
adarsh: German Star Wars character?

spec/controllers/sign_and_submit_controller_spec.rb
```diff
@@ -32,23 +24,8 @@
...
-      end
-    end
-
     let(:params) do
-      {
-        step: {
-          signature: "Chiu Baka",
-        },
-      }
+      { step: { signature: "Chiu Baka" } }
```
adarsh: chewbacca

spec/features/snap_app_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "rails_helper"
+
+feature "SNAP appl" do
```
adarsh: Maybe expand `appl` to `application`?

app/helpers/mb_form_builder.rb
jessieay: I think we need to call it something other than `FormBuilder` because that is the name of `ActionView::Helpers::FormBuilder`. Get CalFresh dealth with this by using the `GCF` prefix, which I agree is not awesome. Ideas for other names?

app/controllers/application_controller.rb
jessieay: Yeah I was going back and forth on what to call this. Get Calfresh Calls this class `CalfreshApplication`. Overall, the word `Application` is annoying. So I prefer "App"...but maybe that is confusing. Also, the challenge with using shorthand is it's easy to forget. For example, in GCF they call things "docs" and "documents" in different places. Very maddening. 

I think `SnapApplication` could work. Based on above, what do you think is best?

app/controllers/introduction_introduce_yourself_controller.rb
jessieay: Yeah I think the reason this is named this way is that `Introduction` is kind of a namespace. In the steps architecture. This makes more sense when you see the former `StepNavigation` code: https://github.com/codeforamerica/michigan-benefits/pull/22/files#diff-36e9c654f58b00a97974aba78e06ff24

For now, I am cool with keeping as-is since this PR is pretty large, but I do think we will want to do some renaming.

app/helpers/mb_form_builder.rb
jessieay: Yeah it is a lot! The good news is that I found these helper methods to be very user-friendly when creating forms for GCF.

app/steps/introduction_contact_information.rb
jessieay: I am using the zip code validation that they had in another step class. I think this validation is fine for now, but yeah perhaps we should refine in the future...

app/views/shared/_flashes.html.erb
jessieay: I just copied this from GCF, which uses ERB. I think this flash stuff should be refactored anyway, so I was hesitant to put time into chaging over to HAML. Created a card for this here:  https://trello.com/c/FPMNPuGm

app/views/users/new.html.haml
jessieay: I think we should do it! but separately https://trello.com/c/FPMNPuGm

spec/controllers/sign_and_submit_controller_spec.rb
jessieay: LOL

spec/controllers/sign_and_submit_controller_spec.rb
jessieay: it wasn't me!

app/models/concerns/multiparameter_attribute_assignment.rb
jessieay: sure thing, updated at 67ed12d

app/steps/introduction_contact_information.rb
jessieay: changing the name of this requires also changing the name of the controller and views. do you think we should do it now or wait?

spec/features/snap_app_spec.rb
jessieay: works fr m e01c7a6

app/controllers/application_controller.rb
adarsh: That works!

jessieay: cool! renamed at 3b482d6

PR #23, opened 2017-07-26
Add pdftk to install script

adarsh:
Reason for Change
=================
* When running the application's tests, I errored on the missing `pdftk` library.
* While this required dependency is described in the `README.md`, it would be easier to add to `bin/setup` instead.

Changes
=======
* Add a step to install `pdftk` via Homebrew. (source of solution is [here])

[here]: https://stackoverflow.com/a/41243285/1001350

Minor
-----
* Added a helper to determine if a cli command is installed or not.
* Ignore `bin/stubs` in rubocop.


adarsh:
@jessieay yeah I removed it in aa2bc64.

PR #26, opened 2017-07-26
Add overcommit gem

adarsh:
Reason for Change
=================
* The [overcommit] gem allows for installation of git commit hooks.
* This can provide a way to run checks like Rubocop when committing, as opposed running tests or on CI, which are further down the code lifecycle.

[overcommit]: https://github.com/brigade/overcommit

Changes
=======
* Add the overcommit gem to the Gemfile
* Add overcommit installation to setup script
* Add the default configuration file.

Minor
=====
* Add a helper to detect if a cli is installed or now.

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -1,0 +1,34 @@
+gemfile: Gemfile
+
+PreCommit:
+  BundleCheck:
+    enabled: true
+
+  ExecutePermissions:
+    enabled: true
+    exclude:
+      - 'bin/**/*'
+      - 'libexec/**/*'
+      - 'template-dir/hooks/**/*'
+
+  HardTabs:
+    enabled: true
+
+  MasterHooksMatch:
+    enabled: true
+    quiet: true
+
+  RuboCop:
+    enabled: true
+    command: ['bundle', 'exec', 'rubocop']
+    include:
+      - '**/*.gemspec'
+      - '**/*.rb'
+      - '**/Gemfile'
+      - template-dir/hooks/overcommit-hook
+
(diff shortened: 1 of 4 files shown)


jessieay:
if code climate is failing on this PR, does that mean that overcommit isn't working as expected?

jessieay:
[Overcommit](https://trello.com/c/L9a59XEB/29-overcommit)

adarsh:
> Should we also enforce a PR commit style with this? That's what we did on this project, which I kind of liked: https://github.com/18F/identity-idp/blob/master/.overcommit.yml#L36

I like the idea of having a unified commit message style!

My personal preference is for something a little more detailed using a [`.gitmessage` file][1]. [Here's the template I usually work off of][2]. It's format is generally in the direction of an RFC, like "Here's the reason for this change, here are the changes".

At any rate, this feels like a separate PR discussion.

[1]: https://robots.thoughtbot.com/better-commit-messages-with-a-gitmessage-template
[2]: https://github.com/adarsh/dotfiles/blob/master/gitmessage

adarsh:
> For some reason I thought we could use overcommit to prevent pushes directly to master, but I think I was thinking of this: https://help.github.com/articles/configuring-protected-branches/ -- should we also config master so it is protected?

Yeah I know GitHub has this as a configuration option. I would vote "yes" since `master` deploys to staging automatically.

adarsh:
> if code climate is failing on this PR, does that mean that overcommit isn't working as expected?

It was failing for some other reason, I believe. After rebasing it's green! 🚦

PR #27, opened 2017-07-26
Create PDF with Snap app info after signing

jessieay:
**WHY**: So we can fax it in.

To test:
* visit home page
* click "apply now"
* go through flow
* open file named "test_pdf.pdf" and confirm that data you enterd is
  included on page 9


Gemfile
```diff
@@ -34,6 +34,7 @@ group :development, :test do
   gem "capybara"
   gem "climate_control"
   gem "codeclimate-test-reporter"
+  gem "factory_girl_rails"
```
adarsh: 🔧

app/controllers/sign_and_submit_controller.rb
```diff
@@ -6,10 +6,32 @@ def update
...
       flash[:notice] = "Your application has been successfully submitted."
       redirect_to root_path(anchor: "fold")
     else
       render :edit
     end
   end
+
+  private
+
+  def create_dhs1171_pdf
```
adarsh: This method could almost be it's own object. Maybe something to keep an eye on?

jessieay: yes totally agree! When I was writing the controller specs for this it was a total pain and I wanted to just have an object to stub. But I think we should take care of it separately / when we start adding more fields.

PR #29, opened 2017-07-27
Refactor use of `User`

jessieay:
**WHY**: Before, we were creating a `User` record in the database with
bogus data for each new SNAP application.

This PR moves us toward what GCF does, which is saving the `id` of the
current SNAP application as a session variable and determining relations
based on that information.

The benefit of this is that we are standardizing on one approach across
apps and not creating meaningless database records.


jessieay:
[Refactor use of `user` in code](https://trello.com/c/eKgn37M9/20-refactor-use-of-user-in-code)

jessieay:
NOTE that we may want to use authorization in the future, but since we currently have nothing scoped for only users or admins, it seemed like we could just take it our for now and bring it back when we need it.

spec/support/mom.rb
```diff
@@ -1,34 +0,0 @@
...
-  @mom ||= Mom.new
-end
-
-def build(thing, *args)
-  mom.send(thing, *args)
-end
-
-def create(thing, *args)
-  mom.send(thing, *args).tap(&:save!)
-end
```
adarsh: mom jeans

app/controllers/address_controller.rb
```diff
@@ -5,7 +5,7 @@ def update
     @step = step_class.new(step_params)
 
     if @step.valid?
-      current_app.update!(step_params)
+      current_snap_application.update!(step_params)
```
adarsh: 💕

app/controllers/application_controller.rb
```diff
@@ -5,57 +5,33 @@ class ApplicationController < ActionController::Base
...
-  def previous_path(*); end
-
-  def current_path(*); end
+  def current_or_new_snap_application
+    current_snap_application || SnapApplication.create
+  end
 
-  def next_path(*); end
+  def current_snap_application
+    id = current_snap_application_id
```
adarsh: This tempvar seems unnecessary, unless I'm missing something.

app/controllers/personal_detail_controller.rb
```diff
@@ -2,6 +2,8 @@
 
 class PersonalDetailController < StepsController
   def edit
+    current_app = current_or_new_snap_application
```
adarsh: What do you think about keeping the controller variable as something like `current_snap_application`, so then it leaves the `ApplicationController` to worry about the "find or create" logic?

So, the controllers just ask for the application and get it, but don't worry about where it comes from?

app/controllers/personal_detail_controller.rb
```diff
@@ -11,7 +13,9 @@ def update
     @step = step_class.new(step_params)
 
     if @step.valid?
-      current_app.update!(step_params)
+      app = current_or_new_snap_application
+      app.update!(step_params)
+      set_current_snap_application(app)
```
adarsh: I'm guessing we'll be doing this a bunch - what do you think about some kind of `SnapApplicationUpdater.new(step_params).update_and_set_current` collaborator? Premature?

spec/controllers/address_controller_spec.rb
```diff
@@ -2,28 +2,23 @@
...
-      user: member,
-      street_address: "123 Fake St",
-      city: "Springfield",
-      county: "Genesee",
-      zip: "12345",
-      state: "MI",
-    )
-  end
-
+RSpec.describe AddressController, type: :controller do
```
adarsh: Minor: I believe we don't need to specify the spec type (e.g. `type: :controller`) because we've set `config.infer_spec_type_from_file_location` [here][1].

[1]: https://github.com/codeforamerica/michigan-benefits/blob/6fc3d9f77d619a12c19b7c7db4dbf01cbe939a93/spec/rails_helper.rb#L65

spec/controllers/address_controller_spec.rb
```diff
@@ -2,28 +2,23 @@
...
   let(:step) { assigns(:step) }
 
   describe "#edit" do
     it "assigns the correct step" do
+      session[:snap_application_id] = current_app.id
```
adarsh: I'm not convinced myself, but this might be a place (maybe the only situation) worth considering moving this into a `before` block. Dunno. What do you think?

spec/controllers/personal_detail_controller_spec.rb
```diff
@@ -2,31 +2,42 @@
...
+      current_app = SnapApplication.create
+      session[:snap_application_id] = current_app.id
+
       get :edit
+
       expect(assigns(:step)).to be_an_instance_of PersonalDetail
     end
 
     it "assigns the name to the step" do
+      current_app = SnapApplication.create(name: "bob")
```
adarsh: This is a model so we could also use a factory here, I believe.

spec/controllers/sign_and_submit_controller_spec.rb
```diff
@@ -29,16 +23,29 @@
...
     it "updates attributes" do
+      session[:snap_application_id] = current_app.id
+
       expect do
         put :update, params: params
       end.to change {
-        current_app.reload.attributes.slice(*attributes.keys)
-      }.from("signature" => "Hans Solo").to("signature" => "Chiu Baka")
+        current_app.reload.signature
+      }.from("Hans Solo").to("Chiu Baka")
```
adarsh: Still hilarious.

spec/models/household_member_spec.rb
```diff
@@ -5,17 +5,21 @@
...
 
     it "does not return HouseholdMembers with consistent income" do
-      described_class.create!(income_consistent: true)
+      described_class.create!(income_consistent: true, app: app)
       expect(described_class.with_inconsistent_income).to be_empty
     end
   end
+
+  def app
+    @_app ||= App.create
```
adarsh: Wait, shouldn't this be `SnapApplication`?

app/controllers/application_controller.rb
jessieay: agreed! removed at  5640fad

app/controllers/personal_detail_controller.rb
jessieay: We only need to set it once (for this controller) - equivalent of "logging in". Do you still think we should create a class for this?

spec/controllers/address_controller_spec.rb
jessieay: Nice 👁  - removed at efb6ffc

spec/controllers/address_controller_spec.rb
jessieay: In general I want to keep this app as similar to GCF as possible, but the use of `let` and `before` in the GCF codebase is probably something I would not choose to duplicate.

All that being said, I agree that this is a place where it makes sense. Done at 8b9527e 😉

spec/models/household_member_spec.rb
jessieay: this is for a model that we are not using (From prototype)

spec/controllers/personal_detail_controller_spec.rb
jessieay: MOM MOM MOM

done at d3e6429

spec/models/household_member_spec.rb
jessieay: (which should be deleted eventually, too. I am just hesitating because I feel like we will most definitely be using this class in The Future)

PR #33, opened 2017-07-27
Ignore pdf files created by specs

adarsh:
Reason for Change
=================
* Running the test suite leaves a new `test_pdf.pdf` file in the repository.

Changes
=======
* `.gitignore` the test file.

Questions
=========
* I wonder if we shouldn't just save the file in `/tmp` or alternatively clean up the file after the spec?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@
 !/log/.keep
 /tmp
 /coverage
+
+test_pdf.pdf


jessieay:
I think I agree with your suggestion to save in `/tmp` rather than just gitignoring. I also think it would be wise to include some type of random string in the pdf name so we don't overwrite the PDF on each run?

adarsh:
This is no longer necessary because of #37, which moves all the file creation into `/tmp`.

Closing without merging.

PR #34, opened 2017-07-27
Add view and controller logic for "email me" feature

jessieay:
* Next step is to actually send the email

screen shot 2017-07-27 at 11 56 50 am


app/controllers/email_application_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class EmailApplicationController < StandardStepsController
```
jessieay: Good idea. I think that this naming is confusing because of the word `Email`. Changed the Step name and related controllers to `SendApplication` at f319808

app/controllers/skip_email_applications_controller.rb
```diff
@@ -0,0 +1,6 @@
+class SkipEmailApplicationsController < ApplicationController
```
jessieay: Renamed at 23685e3

PR #35, opened 2017-07-27
Send email to client after entering email address

jessieay:
* Next steps:
   - make file name dynamic
   - delete file after email is sent (or take advantage of tempfile)
   - send email in background job


app/controllers/send_application_controller.rb
```diff
@@ -14,7 +15,35 @@ def update
...
+      birth_month: current_snap_application.birthday.strftime("%m"),
+      birth_year: current_snap_application.birthday.strftime("%Y"),
+      street_address: current_snap_application.street_address,
+      city: current_snap_application.city,
+      county: current_snap_application.county,
+      state: current_snap_application.state,
+      zip: current_snap_application.zip,
+      signature: current_snap_application.signature,
+      signature_date: current_snap_application.signed_at,
+    }
```
adarsh: I wonder if this data translation/adapter hash could move into the `Dhs1171Pdf` class instead of here in the controller...

app/mailers/application_mailer.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
-  default from: "no-reply@codeforamerica.org"
+  default from: "admin@#{ENV['EMAIL_DOMAIN']}"
   layout "mailer"
+
+  def snap_application_notification(file_name:, recipient_email:)
+    attachments["snap_application.pdf"] = File.read("#{Rails.root}/#{file_name}")
```
adarsh: Should this be in `Rails.root/tmp/`?

app/controllers/send_application_controller.rb
jessieay: YES that is so much better! done at 46f6bc6

app/mailers/application_mailer.rb
jessieay: We are not yet saving it in `/tmp` but I am guessing you will fix that :) 

Will this be problematic on Heroku?

PR #37, opened 2017-07-28
Add a coversheet to the PDF

adarsh:
Reason for Change
=================
* When we fax applications in, we want to provide enough context for the recipient to know where it came from and how to ask questions about it.
* Also, this is a good opportunity to solicit feedback.

Changes
=======
* Change the PDF-creation object interface to accept an optional filename
* Create the PDFs in the `/tmp` directory so as not to clutter up the repository.
* Also, ask for the desired filename when creating the final PDF.

Questions
=========
* Ask rubocop to ignore `/bin` files


app/controllers/sign_and_submit_controller.rb
```diff
@@ -32,6 +32,6 @@ def create_dhs1171_pdf
       signature_date: current_snap_application.signed_at,
     }
 
-    Dhs1171Pdf.new(data).save("test_pdf.pdf")
+    Dhs1171Pdf.new(client_data: data).save
```
jessieay: 👍  to named args

app/models/dhs1171_pdf.rb
```diff
@@ -14,20 +14,25 @@ class Dhs1171Pdf
     "zip",
   ].freeze
 
-  SOURCE_PDF = "DHS_1171.pdf".freeze
+  PDF_DIRECTORY = "lib/pdfs".freeze
+  SOURCE_PDF = "#{PDF_DIRECTORY}/DHS_1171.pdf".freeze
+  COVERSHEET_PDF = "#{PDF_DIRECTORY}/michigan_snap_fax_cover_letter.pdf".freeze
 
-  def initialize(client_data)
+  def initialize(client_data:, output_filename: "test_output.pdf")
```
jessieay: Was thinking about it a few mins ago, and I think it's right for `save` to return `true` but for us to know the file name because of dependency injection. No need for `save` to return the file name. 

This is what you already have here, just wanted to write out my thinking because before I was unsure

app/models/dhs1171_pdf.rb
```diff
@@ -14,20 +14,25 @@ class Dhs1171Pdf
...
     @client_data = client_data
+    @working_filename = "tmp/working_#{output_filename}"
```
jessieay: I wonder if there is a better name for this than `working` -- maybe `filled_out_file_name` and then the output can be called `filled_out_file_with_coversheet_name` ...too verbose?

db/schema.rb
```diff
@@ -68,9 +68,9 @@
     t.integer "income_other"
     t.boolean "has_home"
     t.boolean "has_vehicle"
-    t.text "financial_accounts", default: [], array: true
     t.boolean "sms_reminders"
     t.boolean "email_reminders"
+    t.text "financial_accounts", default: [], array: true
```
jessieay: these seem unrelated?

jessieay: might be my fault. I had some database issues earlier and re-ran all migrations rather than doing a schema load.

spec/models/dhs1171_pdf_spec.rb
```diff
@@ -45,8 +55,8 @@ def filled_in_values(file:)
     end
   end
 
-  def generated_pdf
-    @_generated_pdf ||= "tmp/dhs1171_test_output.pdf"
+  def new_pdf_filename
+    @_new_pdf_filename ||= "dhs1171_test_output.pdf"
```
jessieay: should this be in `tmp` ?

app/models/dhs1171_pdf.rb
adarsh: There is always a better name! I was thinking that I want the final PDF with the PDF to have the same name that was passed into the constructor. That's why I created some kind of interim "working" file.

What about renaming to `completed_form` or similar?

db/schema.rb
adarsh: I'm not sure what this is, I'm happy to remove this from the PR - just let me know.

app/controllers/send_application_controller.rb
```diff
@@ -19,13 +19,16 @@ def update
 
   def send_email
     ApplicationMailer.snap_application_notification(
-      file_name: new_file_name,
+      file_name: "tmp/#{new_file_name}",
```
jessieay: should we just include `/tmp` in the `new_file_name` method?

spec/models/dhs1171_pdf_spec.rb
adarsh: yes, I believe so... let me doublecheck.

app/controllers/send_application_controller.rb
```diff
@@ -19,13 +19,16 @@ def update
...
       recipient_email: current_snap_application.email,
     ).deliver
   end
 
   def create_dhs1171_pdf
-    Dhs1171Pdf.new(current_snap_application).save(new_file_name)
+    Dhs1171Pdf.new(
+      snap_application: current_snap_application,
+      output_filename: new_file_name,
```
jessieay: I think we want this output to `/tmp` so perhaps putting that in the `new_file_name` method will help standardize?

app/models/dhs1171_pdf.rb
```diff
@@ -1,17 +1,41 @@
...
+  ].freeze
 
-  def initialize(snap_application)
+  PDF_DIRECTORY = "lib/pdfs".freeze
+  SOURCE_PDF = "#{PDF_DIRECTORY}/DHS_1171.pdf".freeze
+  COVERSHEET_PDF = "#{PDF_DIRECTORY}/michigan_snap_fax_cover_letter.pdf".freeze
+
+  def initialize(snap_application:, output_filename: "test_output.pdf")
     @snap_application = snap_application
+    @working_filename = "tmp/working_#{output_filename}"
```
jessieay: I can see why you had issues with this -- adding `tmp` all over is confusing! wonering if it would be simpler to just include it as part of the file name string rather than interpolating

app/models/dhs1171_pdf.rb
```diff
@@ -1,17 +1,41 @@
...
+    "full_name",
+    "signature",
+    "signature_date",
+    "state",
+    "street_address",
+    "zip",
+  ].freeze
 
-  def initialize(snap_application)
+  PDF_DIRECTORY = "lib/pdfs".freeze
```
jessieay: nice!

app/models/dhs1171_pdf.rb
```diff
@@ -1,17 +1,41 @@
 class Dhs1171Pdf
-  SOURCE_PDF = "DHS_1171.pdf".freeze
+  FIELDS = [
```
jessieay: this is no longer used

spec/models/dhs1171_pdf_spec.rb
jessieay: ?

adarsh: Oh no - this is just the filename which is passed into the constructor without a path. So this is correct.

jessieay: ah I see. So I think my confusion here underscores the confusion of passing in a variable but not having it reflect the full file path. Maybe rather than calling the argument `filename` we should call it `file_path` and include `tmp/`?

app/models/dhs1171_pdf.rb
```diff
@@ -29,4 +53,10 @@ def client_data
       signature_date: snap_application.signed_at,
     }
   end
+
+  def add_cover_sheet_to_completed_form
+    system(
+      "pdftk #{COVERSHEET_PDF} #{completed_filename} cat output #{output_filename}",
```
jessieay: ugh I am kind of regretting my comments from before because it feels like it is not necessarily clear out of `completed` and `output` which is the more final version. Can always be renamed another time, though...

db/schema.rb
jessieay: nah I think it's fine

app/models/dhs1171_pdf.rb
adarsh: Whoops removed in d31772d

app/models/dhs1171_pdf.rb
adarsh: Yeah I agree. This one is tough.

app/controllers/send_application_controller.rb
adarsh: Yeah it is a good suggestion but is breaking the interim file thing where I append `working_` or equivalent to the filename. It either mangles the path or the extension. I'm going to leave this for now and revisit later.

jessieay:
[As an MDHHS intake worker, I want to know why faxed applications are blank, and what information they contain](https://trello.com/c/bgqY9s7O/51-as-an-mdhhs-intake-worker-i-want-to-know-why-faxed-applications-are-blank-and-what-information-they-contain)

PR #39, opened 2017-07-28
Do PDF creation and email sending in async processes

jessieay:
**WHY**:
- so doesn't have to happen in request, which has a timeout risk
- so they can be retried if they fail

Will probably want these to be in job classes next, but for now we are
just calling `delay` (we are using DelayedJob)


jessieay:
[Set up delayed job](https://trello.com/c/76KWt72V/58-set-up-delayed-job)

jessieay:
@adarsh good call, I moved both into a job class so they can be done async TOGETHER. Please check it out at 
Author

PR #40, opened 2017-07-28
Call Job properly

jessieay:
* Was not sending keyword arg before - oops!

### DIFF
--- a/spec/controllers/send_application_controller_spec.rb
+++ b/spec/controllers/send_application_controller_spec.rb
@@ -34,6 +34,15 @@ RSpec.describe SendApplicationController do
           current_app.reload.email
         }.from("test@example.com").to("new_email@example.com")
       end
+
+      it "creates a SendApplicationJob" do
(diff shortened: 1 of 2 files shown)


spec/controllers/send_application_controller_spec.rb
```diff
@@ -34,6 +34,15 @@
           current_app.reload.email
         }.from("test@example.com").to("new_email@example.com")
       end
+
+      it "creates a SendApplicationJob" do
```
jessieay: this is kind of a lame test because it's all stubs. Thoughts on a better way to test that this is called?

adarsh: It's fine, I believe. All you want to know is that the job is created by the controller, so this is clear to me.

Is the job itself in another PR coming down the pipe?

jessieay:
@adarsh this job is already there, but was failing to run on staging because we weren't calling it properly!

PR #41, opened 2017-07-28
Ignore brakeman warning (via code climate)

jessieay:
**WHY**: we are not worried about this because we determine input

* Output file is programmatically ignored via:
```
gem install brakeman
brakeman --interactive-ignore
```

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,0 +1,26 @@
+{
+  "ignored_warnings": [
+    {
+      "warning_type": "Command Injection",
+      "warning_code": 14,
+      "fingerprint": "c7c9b4de9c22b2c60bc74a8246bae70fe99146bae901fc6559dd983fe02ce6b1",
+      "check_name": "Execute",
+      "message": "Possible command injection",
+      "file": "app/models/dhs1171_pdf.rb",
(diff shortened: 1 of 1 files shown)


config/brakeman.ignore
```diff
@@ -0,0 +1,26 @@
...
+        "method": "add_cover_sheet_to_completed_form"
+      },
+      "user_input": "\"lib/pdfs\".freeze",
+      "confidence": "Medium",
+      "note": ""
+    }
+  ],
+  "updated": "2017-07-28 13:56:10 -0700",
+  "brakeman_version": "3.7.0"
+}
```
adarsh: This is good but hold off on merging because the location of that method call will change a bit before end of the day.

jessieay: ok sounds good - we can also always run `brakeman --interactive-ignore` again and it will auto-generate this file differently if you select `i` when this error pops up

jessieay:
This circle error is confusing me

adarsh:
I added this `brakeman.ignore` file in https://github.com/codeforamerica/michigan-benefits/pull/43 so I believe this can be closed without merging.

jessieay:
cool! closing

PR #43, opened 2017-07-28
Use Ruby's tempfiles for temporary files

adarsh:
Reason for Change
=================
* Much of what we were doing manually with files (creating them, randomizing names, cleaning them up after use) is managed by Ruby's [tempfile] library.

[tempfile]: https://docs.ruby-lang.org/en/2.4.0/Tempfile.html

Changes
=======
* In the generation of the completed `Dhs1171Pdf` file, use `tempfile`s in the pdf generation and return the resulting `tempfile`.
* Be sure to close the `tempfile` explicitly in this class.
* In the job class, generate and close the tempfile after sending is complete.

Minor
-----
* No need to profile all specs on every run.
* gitignore .rspec because those are personal preferences, not team ones

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,4 +1,0 @@
---color
---require spec_helper
---order random
---profile
--- a/app/jobs/send_application_job.rb
+++ b/app/jobs/send_application_job.rb
@@ -1,31 +1,28 @@
 class SendApplicationJob < ApplicationJob
   def perform(snap_application:)
-    create_pdf(snap_application)
-    send_pdf(snap_application)
-    destroy_pdf
+    @snap_application = snap_application
+
+    create_pdf
+    send_pdf
+  ensure
+    complete_form_with_cover.try(:close)
+    complete_form_with_cover.try(:unlink)
   end
 
   private
 
-  def create_pdf(snap_application)
-    Dhs1171Pdf.new(
+  attr_reader :complete_form_with_cover, :snap_application
+
+  def create_pdf
+    @complete_form_with_cover = Dhs1171Pdf.new(
       snap_application: snap_application,
-      output_filename: new_file_name,
-    ).save
+    ).completed_file
   end
 
-  def send_pdf(snap_application)
+  def send_pdf
     ApplicationMailer.snap_application_notification(
-      file_name: "tmp/#{new_file_name}",
+      file_name: complete_form_with_cover.path,
       recipient_email: snap_application.email,
     ).deliver
-  end
-
-  def destroy_pdf
-    # delete pdf
-  end
-
-  def new_file_name
-    "test_pdf.pdf"
   end
 end
--- a/app/models/dhs1171_pdf.rb
+++ b/app/models/dhs1171_pdf.rb
@@ -3,23 +3,29 @@ class Dhs1171Pdf
   SOURCE_PDF = "#{PDF_DIRECTORY}/DHS_1171.pdf".freeze
   COVERSHEET_PDF = "#{PDF_DIRECTORY}/michigan_snap_fax_cover_letter.pdf".freeze
 
-  def initialize(snap_application:, output_filename: "test_output.pdf")
+  def initialize(snap_application:)
     @snap_application = snap_application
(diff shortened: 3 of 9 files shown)


.rspec
```diff
@@ -1,4 +1,3 @@
 --color
```
jessieay: from PR description it looks like you want to ignore this whole file -- 100% cool me! So let's do it? And move the `random` config to `spec_helper` ?

app/jobs/send_application_job.rb
```diff
@@ -1,31 +1,28 @@
...
-      output_filename: new_file_name,
-    ).save
+    ).completed_file
   end
 
-  def send_pdf(snap_application)
+  def send_pdf
     ApplicationMailer.snap_application_notification(
-      file_name: "tmp/#{new_file_name}",
+      file_name: complete_form_with_cover.path,
```
jessieay: niiiice - so much cleaner!

app/models/dhs1171_pdf.rb
```diff
@@ -39,9 +45,11 @@ def client_data
...
-  def add_cover_sheet_to_completed_form
-    system(
-      "pdftk #{COVERSHEET_PDF} #{completed_filename} cat output #{output_filename}",
-    )
+  def prepend_cover_sheet_to_completed_form
+    system("pdftk #{COVERSHEET_PDF} #{filled_in_form.path} cat output #{complete_form_with_cover.path}")
+  end
+
+  def complete_form_with_cover
+    @_with_cover ||= Tempfile.new(["snap_app_with_cover", ".pdf"], "tmp/")
```
jessieay: this is super nit picky but I almost always create memo-ized ivars with the same name as the method so I never have two ivars with the same name (so here would be `@_completed_form_with_cove` )

app/models/dhs1171_pdf.rb
```diff
@@ -3,23 +3,29 @@ class Dhs1171Pdf
...
+  attr_reader :snap_application
 
-  def fill_in_template_form
-    PdfForms.new.fill_form(SOURCE_PDF, completed_filename, client_data)
+  def complete_template_pdf_with_client_data
+    PdfForms.new.fill_form(SOURCE_PDF, filled_in_form.path, client_data)
+  end
+
+  def filled_in_form
+    @file_filled_in ||= Tempfile.new(["snap_app", ".pdf"], "tmp/")
```
jessieay: same nit picky comment: what do you think about making the ivar have the same name as the method?

app/models/dhs1171_pdf.rb
```diff
@@ -3,23 +3,29 @@ class Dhs1171Pdf
...
   end
 
-  def save
-    fill_in_template_form
-    add_cover_sheet_to_completed_form
+  def completed_file
+    complete_template_pdf_with_client_data
+    prepend_cover_sheet_to_completed_form
+    complete_form_with_cover
+  ensure
```
jessieay: oooh I am not familiar with `ensure` - what will happen if the file is not closed or unlinked properly? should we have a unit test for that?

.rspec
adarsh: Yeah I changed my mind on it, but since you agree I removed and gitignored it as well as moved the config to `spec_helper` as you suggested in 9490ad7.

app/models/dhs1171_pdf.rb
adarsh: If the `tempfile` is not explicitly closed, it will hang around until the next Garbage Collection. Not a huge deal but best practice is to clean up after yourself.

`ensure` is a block modifier which ensures whatever is inside will get run regardless of any exceptions or interruptions.

app/models/dhs1171_pdf.rb
adarsh: Oh I like it! Done in 23d1aeb

app/models/dhs1171_pdf.rb
adarsh: Done in 23d1aeb

PR #45, opened 2017-07-28
Remove more unused stuff / duplication

jessieay:
**WHY**: why not?

This was inspired by our low test coverage score in code climate. I
removed stuff we do not use / have tests for


app/controllers/address_controller.rb
```diff
@@ -1,14 +1,4 @@
 # frozen_string_literal: true
 
 class AddressController < StandardStepsController
```
adarsh: Do we need to keep this file?

app/controllers/sign_and_submit_controller.rb
```diff
@@ -1,14 +1,4 @@
 # frozen_string_literal: true
 
 class SignAndSubmitController < StandardStepsController
```
adarsh: Same, do we need to keep this still?

app/controllers/address_controller.rb
jessieay: Yes we are using this to collect address info, but `update` is defined in ` StandardStepsController` so we don't need to duplicate the logic here

app/controllers/sign_and_submit_controller.rb
jessieay: same as above! We are using it, but `update` does not have special logic so we can fall back on `StandardStepsController` -- thanks citizen code, for that!

PR #51, opened 2017-07-31
Add sfax, corresponding fax adapter and test

jayroh:
The sfax gem has been taken over by another project so pointing to the
CFA fork that the GCF team has set up is the best bet.

This commit adds an adapter class to interface with the SFax gem and
tests that it is being called correctly, using the credentials places in
the Rails secret yaml file.

NOTE: Adding faraday is necessary, for some reason to get things to
function reliably.

### DIFF
--- a/app/models/fax.rb
+++ b/app/models/fax.rb
@@ -21,11 +21,9 @@ class Fax
     sfax.send_fax(number, file)
   end
 
-  protected
+  private
 
   attr_reader :file, :number
-
-  private
 
   def sfax
     SFax::Faxer.new(
--- a/app/models/fax.rb
+++ b/app/models/fax.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+class Fax
+  delegate \
+    :sfax_api_key,
+    :sfax_encryption_key,
+    :sfax_init_vector,
+    :sfax_username,
+    to: :"Rails.application.secrets"
+
+  def self.send_fax(number:, file:)
+    new(number, file).send_fax
+  end
+
+  def initialize(number, file)
+    @number = number
+    @file = file
+  end
+
+  def send_fax
+    sfax.send_fax(number, file)
+  end
+
+  protected
+
+  attr_reader :file, :number
+
+  private
+
+  def sfax
+    SFax::Faxer.new(
+      sfax_username,
+      sfax_api_key,
+      sfax_init_vector,
+      sfax_encryption_key,
+    )
+  end
+end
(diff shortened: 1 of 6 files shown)


app/models/fax.rb
```diff
@@ -0,0 +1,38 @@
...
+  def initialize(number, file)
+    @number = number
+    @file = file
+  end
+
+  def send_fax
+    sfax.send_fax(number, file)
+  end
+
+  protected
```
adarsh: I forget the canonical use cases - why `protected` and not `private` here?

config/secrets.yml
```diff
@@ -15,8 +15,19 @@ development:
 
 test:
   secret_key_base: a4601e0fb037b7d1ed534a5d5dd114f47b753c9945757ebe0f9fedff98d41b12741f7dd631e0d9a69cd518623384d871b0387ec5645ccccbbb5d9336ff1cf800
+  sfax_username: sfax_username
+  sfax_api_key: sfax_api_key
+  sfax_encryption_key: sfax_encryption_key
+  sfax_init_vector: sfax_init_vector
+
+from_environment: &deployable_settings
+  sfax_username: <%= ENV['SFAX_USERNAME'] %>
```
adarsh: I wonder - is there any downside to just referencing `ENV['WHATEVER']` in the code itself? We have other keys which we are managing directly from the environment variables. ([here's an example of the SMTP credentials][1])

[1]: https://github.com/codeforamerica/michigan-benefits/blob/3ece19a001d1ea57cf4d228067cc2841e279a397/config/smtp.rb

app/models/fax.rb
jayroh: It's mostly out of habit at this point but the reason is generally for purposes like equality checking. A protected attribute allows an instance of a class to inquire about another instance's values ... which, for things like value objects is a necessity.

Instead of thinking about it, I've just gotten used to putting `attr_*` in the protected object space.

config/secrets.yml
jayroh: I don't think there's downside, nah. However, I took the lead from the approach used in GCF for this - and this is how they did it.

Another data point - this sort of implementation *would* allow for a pretty quick refactor to use the new encrypted secret feature in Rails 5.1+!

app/models/fax.rb
adarsh: Either way sounds good to me, however we should probably be consistent across the codebase. I'm slightly in favor of `private`, but frankly, I'll be happy if they're not fully public `attr_*`s. 

Ping @jessieay for perspective. She's off this week so maybe merge and we can discuss when she is back? Up to you.

config/secrets.yml
adarsh: Nice! Let's do that then and standardize on that.

Maybe add something quick in `CONTRIBUTING.md`, like a "Secret Keys" section so others know where to put them?

app/models/fax.rb
jayroh: I'm completely cool adjusting for consistency - so I'll move them into the private space :). All good! 👍

config/secrets.yml
jayroh: Done!

PR #52, opened 2017-08-01
Remove possible PII id if exiting application process

jayroh:
If clicks out of the application process in the middle of it without
submitting their data we want to remove the snap application id from the
session so that subsequent visits to the form(s) are not populated with
their PII.

This is in an effort to keep users' information private and secure.

* Add feature spec that tests that data is persisted across steps


app/controllers/application_controller.rb
```diff
@@ -23,7 +23,8 @@ def set_current_snap_application(application)
   end
 
   def current_or_new_snap_application
-    current_snap_application || SnapApplication.create
+    @_current_or_new_snap_applicatin ||=
```
adarsh: Typo, "applicatin"

jayroh:
>  Is it worth encrypting the session ID or is that overkill?

@adarsh - we can do that with `session.signed`, is that right? I'm not 100% positive

jayroh:
Actually, per [this SO post]:

> If you have secret_key_base set, your cookies will be encrypted. This goes a step further than signed cookies in that encrypted cookies cannot be altered or read by users. This is the default starting in Rails 4.

[this SO post]: https://stackoverflow.com/a/32720274/547300

PR #54, opened 2017-08-03
Serve only Genesee County

adarsh:
Reason for Change
=================
* Users often are not aware of their county name.
* As such, asking them to fill in a form with the county name may not be productive.

Changes
=======
* Lock the usage to users of Genesee county only.

Minor
-----
* Add the `awesome_print` gem for colorful debugging (use `ap` instead of `puts`)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ group :test do
 end
 
 group :development, :test do
+  gem "awesome_print", require: false
   gem "capybara"
   gem "climate_control"
   gem "codeclimate-test-reporter"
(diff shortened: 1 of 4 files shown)


adarsh:
[When I am filling out the form and applying for SNAP, I don't always know my county name](https://trello.com/c/kUFdnyIe/66-when-i-am-filling-out-the-form-and-applying-for-snap-i-dont-always-know-my-county-name)

adarsh:
image

jayroh:
@adarsh I think this looks good. The only suggestion I might give, which is totally a 6-to-1/half-dozen thing, is to set the value of the county form field to Genesee and set `disable: true` on that same field.

The small win you'd get with that is that you wouldn't have to merge the county in on the controller side of things. It's purely a view-level change at that point.

jayroh:
👍

PR #56, opened 2017-08-03
Add "Documents" step to the application

jayroh:
There is a need for an additional page in the application for collecting
all necessary supporting documents. This commit adds that page and all
of the dependencies around it. This includes, but it not limited to, the
following:

* Documents controller, step, and navigation update.
* Adding a new field to the SnapApplication model - an string array that
  contains the URL's to the uploaded documents.
* Using Shubox(.io) to facilitate the uploading of photos to S3.
* Application feature is updated to run w/the JS driver so that images
  can be dynamically addded to the documents page with javascript.
* Updates to the step base class to allow a collection as a value
  instead of JUST a single string.

Needs further work:

* Design and UX for the documents page

## On/after uploading on a previous visit to Documents page 
michigan_assistance_programs


app/steps/step.rb
```diff
@@ -8,7 +8,12 @@ class Step
...
-    self.attribute_names = attribute_names.map(&:to_s)
-    attr_accessor(*attribute_names)
+    self.attribute_names = attribute_names
+
+    attributes_or_keys = attribute_names.map do |attr|
+      attr.class == Symbol ? attr : attr.keys
+    end
+
+    attr_accessor(*attributes_or_keys.flatten.map(&:to_s))
   end
```
jayroh: The changes in here are to allow for not only what we've had up until now:

`"key" => "value"`

but also an array of values, which is what we're receiving when we submit N documents:

`"key" => ["value1", "value2"]`

app/assets/javascripts/form_exit.js.erb
```diff
@@ -1,9 +1,14 @@
 ;(function (w, d) {
   exitHandler = function () {
     var links = d.querySelectorAll('A');
+    var inSteps = window.location.href.indexOf('/steps/') >= 0;
+    if(!inSteps) return;
```
jayroh: Add a guard clause here that bails out if we're not ***within*** the steps pages. If we're not inside of those form pages, we really don't care where we click out to.

app/assets/javascripts/form_exit.js.erb
```diff
@@ -1,9 +1,14 @@
...
 
     for(var i=0; i<links.length; i++){
-      if(links[i].href.indexOf('steps') < 0){
+      var notStepsLink = links[i].href.indexOf('steps') < 0
+      var notSkipLink = links[i].href.indexOf('skip_') < 0
+
+      if(notStepsLink && notSkipLink){
```
jayroh: Take into account the situation where, on the very last page, we're clicking the "skip" link. We don't want to be alerted if we click the "skip" link,

Gemfile
```diff
@@ -31,6 +31,7 @@ gem "uglifier", ">= 1.3.0"
 group :test do
   gem "database_cleaner"
   gem "faker"
+  gem "launchy"
```
jayroh: Add launchy to allow for `save_and_open_(page|screenshot)`

app/assets/stylesheets/organisms/_form-card.scss
```diff
@@ -120,3 +120,82 @@
...
+
+  i.icon-file_upload {
+
+  }
+}
+
+.form-card__documents {
+  &:empty { display: none; }
+
+  .dz-preview {
```
adarsh: What does the `dz` prefix here stand for?

app/controllers/standard_steps_controller.rb
```diff
@@ -3,7 +3,9 @@
 class StandardStepsController < StepsController
   def edit
     @step = step_class.new(
-      current_snap_application.attributes.slice(*step_attrs),
+      HashWithIndifferentAccess.
```
adarsh: What do you think about extracting a private method like

```ruby
def application_attributes
  HashWithIndifferentAccess.new(current_snap_application.attributes)
end
```

on which you could call `.slice(*step_attrs)`?

Not sure if that's better, just feels like this one controller action is doing a good amount of work.

app/steps/step.rb
```diff
@@ -8,7 +8,12 @@ class Step
   class_attribute :attribute_names
 
   def self.step_attributes(*attribute_names)
-    self.attribute_names = attribute_names.map(&:to_s)
-    attr_accessor(*attribute_names)
+    self.attribute_names = attribute_names
+
+    attributes_or_keys = attribute_names.map do |attr|
+      attr.class == Symbol ? attr : attr.keys
```
adarsh: Style: I wonder if we could expand the ternary operator to clarify the branching logic? Probably just me but I often have trouble following it when it's all in the one line.

app/steps/step.rb
adarsh: Nice!

db/migrate/20170802203515_add_documents_to_snap_applications.rb
```diff
@@ -0,0 +1,5 @@
+class AddDocumentsToSnapApplications < ActiveRecord::Migration[5.1]
+  def change
+    add_column :snap_applications, :documents, :string, array: true, default: []
```
adarsh: Nice array usage 🦑

jayroh:
> My only question - how does this file upload work on mobile? Does Safari/Chrome Mobile handle uploading phone pics easily? I remember hearing this was the dominant use case, so probably good to think at some point (need not be this PR)

Great question! Big part of what I did with shubox was make sure this was 💯 . TLDR: it's great.

It's better to show it off so I would recommend visiting http://shubox.io/demos/ on your phone and giving those a shot. Let me know what you think :)

app/steps/step.rb
jayroh: True true true - agreed.

app/controllers/standard_steps_controller.rb
jayroh: I like it. Deconstructs the complexity in there well

app/assets/stylesheets/organisms/_form-card.scss
jayroh: It stands for `dropzone`. What's Dropzone? [Dropzone.js] is the OSS JS library that shubox wraps around to get a lot of the upload plumbing to work. It's pretty dope.

[Dropzone.js]: http://www.dropzonejs.com/

Instead of painting over it w/the shubox libs and namespacing it on my own I've just let it be, let it maintain it's rightful identity.

adarsh: Makes sense, although I wonder if we can alias `dz` to `shubox` somehow to make this clearer? Not a huge deal, just speculating.

PR #58, opened 2017-08-04
Fix typo for disabling a field

adarsh:
Reason for Change
=================
* The HTML directive is `disabled: true` with a "d" at the end. I had it as `disable: true` accidentally.
* See [reference].

[reference]: https://www.w3schools.com/jsref/prop_text_disabled.asp

Changes
=======
* Fix typo.

Minor
-----
* Add note specifying we are currently only serving Genesee County, so people aren't upset about the disabled text box.


jayroh:
@adarsh found the culprit and pushed a commit up to fix it all up. Turns out `readonly` WAS the best way to handle that 😮

adarsh:
Ah thanks @jayroh!

PR #60, opened 2017-08-07
Update docs upload UI to be more like GCF

jayroh:
Using a UI that's more analogous to GetCalFresh gets us a little further
along in the development without re-inventing another solution to a
design they've already tackled.

This commit re-uses some of their (s)css and extracts some of the views
out to partials that can be used with a templating language like
handlebars, which is also included here.

Using handlebars allows us to render some html we already use for the
documents collection - therefore not needing to escape into a large JS
string.

Demo:

doc upload demo


app/views/documents/edit.html.erb
```diff
@@ -33,14 +32,32 @@
 
 <script src="https://js.shubox.io/v1/<%= Rails.application.secrets.shubox_js_id %>.js"></script>
 
+<script id="template" type="text/x-handlebars-template">
+<%= render "document", url: nil %>
+</script>
+
 <script>
 new Shubox("#document-drop", {
```
jessieay: would it make sense to include this JS in its own js file that is included here?

app/views/documents/_document.html.erb
```diff
@@ -1,4 +1,9 @@
-<div class="uploaded_document">
-  <%= f.mb_input_field :document, nil,
-    options: { value: File.basename(document), disabled: true } %>
+<div class="doc-preview">
+  <div class="doc-preview__info">
+    <h4><%= File.basename(url || '{{filename}}') %></h4>
```
jessieay: where this `filename` is coming from feels kind of mysterious. Any way you can think of that would make it more obvious?

app/assets/stylesheets/organisms/_document-preview.scss
```diff
@@ -1,36 +1,28 @@
-.document-previewer {
-  display: inline-block;
-  margin-bottom: 1em;
+.doc-preview-container {
```
jessieay: one think I noticed on GCF is that we used `doc` and `document` interchangeably and I never knew which one to use where. What do you think about attempting to use one or the other in this project so there is less guesswork?

app/views/documents/edit.html.erb
```diff
@@ -33,14 +32,32 @@
 
 <script src="https://js.shubox.io/v1/<%= Rails.application.secrets.shubox_js_id %>.js"></script>
 
+<script id="template" type="text/x-handlebars-template">
```
jessieay: would it make sense to use an id or class that somehow indicates we are using this for js purposes? Not sure how you feel about js prefixed. Alternatively, maybe a name more specific than `template` would be helpful.

app/views/documents/edit.html.erb
```diff
@@ -33,14 +32,32 @@
...
+<%= render "document", url: nil %>
```
jessieay: why do we render this once with a `nil` value for URL?

app/views/documents/_document.html.erb
jayroh: Yeah, it's weird. Best way to describe its purpose is that it's a placeholder for handlebars.js to use to interpolate.

So, if we have a url, get it's `basename`, otherwise output `'{{filename}}'` so handlebars can replace it.

What do you think about putting it up at the top in some ruby like:

```ruby
<%
doc_title = if url
              File.basename(url)
            else
              "{{filename}}"
            end
%>

... ✂ snip ✂ ...

<h4><%= doc_title %></h4>
```

app/views/documents/edit.html.erb
jayroh: This is in order to trigger the usage of the handlebars `"{{variable}}"` inside the partial. Like:

> "Url is nil so expect that this will be used as a handlebars template - spit this out with the `'{{...}}'` placeholders."

app/views/documents/edit.html.erb
jayroh: After writing a big comment on the pros and cons, then deleting it, I think you're right. It can/should be in its own thing.

I'll work on that.

jayroh:
> One final question: do we want an integration test for this behavior? Assume we have none existing since these updates did not require changes to our test suite

I included *some* within the context of the end to end application steps - tests that the documents form submits if several hidden form fields with image URL's are added to the form. It's why those feature specs are now set to be JS instead of using rack-test.

Now, if we want to test the *shubox* functionality, that's a whole other ball game, and we should discuss the implications of that.

jayroh:
Code climate says 19 issues to fix. 
Then I click over to https://codeclimate.com/github/codeforamerica/michigan-benefits/pull/60 
and it says everything's good. What the heck?

jayroh:
@jessieay All good w/Rubocop. The errors it was complaining about were related to eslint. Looking into how I can turn that off bc we're not exactly writing a SPA here.

jayroh:
@jessieay could I get a review on this one more time?

app/assets/stylesheets/organisms/_document-preview.scss
jessieay: tough call. It seems like in some places in the app we'd have to call it "document" since "doc" can have many meanings and might confuse people. Would be happy to have less typing, though. We can punt for now :)

app/assets/javascripts/upload_setup.js
```diff
@@ -0,0 +1,28 @@
+/*eslint no-unused-vars: ["error", { "vars": "local" }]*/
+/*global uploadSetup */
```
jessieay: dumb question: don't we have to include this in the manifest or a view file in order for it to be available?

app/views/documents/edit.html.erb
```diff
@@ -1,4 +1,7 @@
 <% content_for :header_title, "Documents" %>
+<% content_for :header_javascript do %>
+  <%= javascript_include_tag "https://js.shubox.io/v1/#{Rails.application.secrets.shubox_js_id}.js", defer: true %>
```
jessieay: 😍

app/assets/javascripts/upload_setup.js
jayroh: `//= require_tree .` will pull it in :)

PR #66, opened 2017-08-07
As an MDHHS worker, I want to know the clients phone number and email

jessieay:
Adds phone number, sms subscribed (boolean), and email fields to a new 
"contact information" page, which comes before address page.

screen shot 2017-08-07 at 4 15 00 pm

### DIFF
--- a/spec/controllers/address_controller_spec.rb
+++ b/spec/controllers/address_controller_spec.rb
@@ -4,9 +4,7 @@ require "rails_helper"
 
 RSpec.describe AddressController, type: :controller do
   let(:step) { assigns(:step) }
-  before do
-    session[:snap_application_id] = current_app.id
-  end
+  before { session[:snap_application_id] = current_app.id }
 
   describe "#edit" do
     it "assigns the correct step" do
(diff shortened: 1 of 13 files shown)


jessieay:
[As an MDHHS worker, I want to know the clients phone number and email so I can follow up with them](https://trello.com/c/Wcz2Uf6z/81-as-an-mdhhs-worker-i-want-to-know-the-clients-phone-number-and-email-so-i-can-follow-up-with-them)

jessieay:
@jayroh CI failure seems like it might be related to radio? In any case, would be neat if doc upload worked without js --possible? (like a fallback or something)

spec/controllers/address_controller_spec.rb
```diff
@@ -79,7 +77,8 @@
   end
 
   def current_app
-    @_current_app ||= SnapApplication.create!(
+    @_current_app ||= FactoryGirl.create(
```
jayroh: Hmm, do we not include the factory girl helper methods?

Feel like that's something we should/could do in the near future.

jessieay: yes we for sure should! they were not using FG when we first started, so we added it to work alongside their custom factory thingy. Could not use FG helper methods here because there would be clashing namespaces. will add soon!

PR #67, opened 2017-08-07
Add demo banner, replace image on home page

jessieay:
screen shot 2017-08-07 at 4 25 33 pm


jessieay:
[As a client, I want to understand that MichiganBenefits.org is an application assister, be confident using the service, and have on off-ramp to the official government site in case I change my mind.](https://trello.com/c/vYE0RlCX/49-as-a-client-i-want-to-understand-that-michiganbenefitsorg-is-an-application-assister-be-confident-using-the-service-and-have-on)

PR #72, opened 2017-08-09
Add residential address step

jessieay:
* This is skipped if the client indicates mailing and residential are
the same
* To do in follow up PRs:
   - save mailing address in the "addresses" table instead of on the snap
   app
  - make sure "back" works with "skip" functionality

screen shot 2017-08-09 at 9 42 48 am

### DIFF
--- a/app/controllers/residential_address_controller.rb
+++ b/app/controllers/residential_address_controller.rb
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
 
 class ResidentialAddressController < StandardStepsController
-  def edit
-    @step = step_class.new(snap_application_attributes.slice(*step_attrs))
-  end
-
   def update
     @step = step_class.new(step_params)
 
--- a/app/controllers/residential_address_controller.rb
+++ b/app/controllers/residential_address_controller.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+class ResidentialAddressController < StandardStepsController
+  def edit
+    @step = step_class.new(snap_application_attributes.slice(*step_attrs))
+  end
+
+  def update
+    @step = step_class.new(step_params)
+
+    if @step.valid?
+      current_snap_application.update!(unstable_housing: unstable_housing?)
+      residential_address.update!(step_params.except(:unstable_housing))
+      redirect_to(next_path)
+    else
+      render :edit
+    end
+  end
+
+  private
+
+  def snap_application_attributes
+    HashWithIndifferentAccess.new(residential_address_attributes)
+  end
+
+  def residential_address_attributes
+    residential_address.attributes.merge(unstable_housing: current_snap_application.unstable_housing)
+  end
+
+  def residential_address
+    current_snap_application.addresses.where.not(mailing: true).first || current_snap_application.addresses.create(mailing: false)
+  end
+
+  def step_params
+    super.merge(state: "MI", county: "Genesee")
+  end
+
+  def unstable_housing?
+    step_params[:unstable_housing] == "1"
+  end
+
+  def skip?
+    current_snap_application.mailing_address_same_as_residential_address?
+  end
+end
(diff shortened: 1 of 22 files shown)


jessieay:
[As an MDHHS worker, I want to know the clients mailing address and residential address, so I can follow up with them](https://trello.com/c/WgULFMVE/80-as-an-mdhhs-worker-i-want-to-know-the-clients-mailing-address-and-residential-address-so-i-can-follow-up-with-them)

app/controllers/steps_controller.rb
```diff
@@ -44,4 +46,14 @@ def next_path(params = nil)
   def step_navigation
     @step_navigation ||= StepNavigation.new(self)
   end
+
+  def maybe_skip
```
jessieay: oooh I can already see that this is going to cause an issue with going back

config/initializers/inflections.rb
```diff
@@ -13,14 +13,4 @@
...
-  inflect.uncountable %w[
-    expenses_additional_sources
-    household_health_situations
-    household_personal_details
-    household_situations
-    income_additional_sources
-    income_other_assets
-    maybe_submit_documents
-    preferences_reminders
-  ]
```
jayroh: Just out of curiosity - how come these are getting removed?

jessieay: we aren't using these anymore! Were deleted two weeks ago. :)

Back story on why I even found these: I initially forgot that we had a non-AR class named `Address` so when I added the addresses table and tried to make a AR class it wasn't working and I was baffled. One theory I had was that somehow the inflection rules for the app weren't letting Rails see `Address` as the AR class for the `addresses` table...but then I realized we had two classes :)

Can remove in a separate PR, if you like!

jayroh: Naaaahhhh leave it on in here.  No biggie 😁

app/controllers/residential_address_controller.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class ResidentialAddressController < StandardStepsController
+  def edit
+    @step = step_class.new(snap_application_attributes.slice(*step_attrs))
+  end
```
jayroh: Correct me if I'm wrong, but isn't this action's contents already in `StandardStepsController`?

app/controllers/steps_controller.rb
jayroh: If `skip?` is set to false, isn't this always going to end up with the same behavior? If that's the case I would imagine the `skip?` method will end up, in the future, being changed to result in a nondeterministic bool.

Could this work be deferred until that's needed?

db/migrate/20170808225329_create_addresses.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :addresses do |t|
+      t.timestamps null: false
+      t.string :street_address
+      t.string :city
+      t.string :county
+      t.string :state
+      t.string :zip
+      t.boolean :mailing
+      t.belongs_to :snap_application, index: true
+    end
```
jayroh: Is there anything here we should enforce with some `null: false` constraints? Looking at `app/steps/mailing_address.rb` it seems as if some of these could use some validation on the db side for data integrity purposes.

app/controllers/residential_address_controller.rb
jessieay: you are right! I initially had this looking pretty different from `StandardStepsController` but then realized I could just overwrite `snap_application_attributes` and it would work. 

Removed at  ba02d66

app/controllers/steps_controller.rb
jessieay: `skip?` has some logic in `ResidentialAddressController`-- are you suggesting I only define this `before_action` there and then include in the base class later?

FWIW, I got the idea for this behavior from the prototype app for michigan benefits. This is how they handled skipping steps.

db/migrate/20170808225329_create_addresses.rb
jessieay: great idea! updated at bf8dc30

PR #73, opened 2017-08-09
Add documents step after sign / submit step

jessieay:
**WHY**: want people to submit before worrying about docs


jessieay:
[As a client, I want to sign by typing my full legal name right after I agree to terms, but BEFORE I have to think about documents](https://trello.com/c/oXGsPHkI/83-as-a-client-i-want-to-sign-by-typing-my-full-legal-name-right-after-i-agree-to-terms-but-before-i-have-to-think-about-documents)

PR #74, opened 2017-08-09
Client can go "Back" if home and residential addresses are same

jessieay:
* Before, `skip` logic would always send the user forward

### DIFF
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -30,12 +30,10 @@ class StepsController < ApplicationController
   end
 
   def maybe_skip
-    if skip?
-      if going_backwards?
-        redirect_to previous_path(rel: "back")
-      else
-        redirect_to next_path
-      end
+    if skip? && going_backwards?
+      redirect_to previous_path(rel: "back")
+    elsif skip?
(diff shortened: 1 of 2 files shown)


app/controllers/steps_controller.rb
```diff
@@ -29,13 +25,31 @@ def step_attrs
...
+
+  def maybe_skip
+    if skip?
+      if going_backwards?
+        redirect_to previous_path(rel: "back")
+      else
+        redirect_to next_path
+      end
+    end
+  end
```
jayroh: Nested conditionals can sometimes feel a little smelly so what do you think about breaking this down to ...

```ruby
def maybe_skip
  if going_backwards?
    redirect_to previous_path(rel: "back")
  elsif skip?
    redirect_to next_path
  end
end

# ... ✂

def going_backwards?
  skip? && params["rel"] == "back"
end
```

I'm looking at this and haven't *quite* decided which approach is better. I'll leave it up to you to decide how you feel about it and will defer to your best judgement. Just presenting that as an option 😄

jessieay: love it!! done at 1193211

PR #75, opened 2017-08-09
Add success resource/page with boilerplate copy

jayroh:
As a client, after the final step they should see a confirmation page
telling them that they are FINISHED! In addition, give them guidance on
what to expect next.

Also:

* Rescue and provide a NullStep for instances where a form object class
  can not be inferred from the controller path.

Trello:

https://trello.com/c/DVStptea/82-as-a-client-after-i-uploading-docs-or-choosing-not-to-i-want-to-see-a-confirmation-page-telling-me-i-am-finished-and-giving-guid


jayroh:
[As a client, after I uploading docs or choosing not to, I want to see a confirmation page telling me I am FINISHED! and giving guidance on what to do next](https://trello.com/c/DVStptea/82-as-a-client-after-i-uploading-docs-or-choosing-not-to-i-want-to-see-a-confirmation-page-telling-me-i-am-finished-and-giving-guid)

app/controllers/send_application_controller.rb
```diff
@@ -7,8 +7,7 @@ def update
     if @step.valid?
       current_snap_application.update(step_params)
       create_and_send_pdf
-      flash[:notice] = "You will receive an email with your filled out application attached in a few minutes."
-      redirect_to root_path(anchor: "fold")
+      redirect_to(next_path)
```
jessieay: no flash to let them know it was sent?

jessieay: Per Alan's comment in The Trello:

> If you email a copy to yourself, let's take you back to this page. We should give some confirmation that the email worked.

jayroh:
Updated this to resemble the following (please excuse the weird offset/sizing):

michigan_assistance_programs
michigan_assistance_programs

app/views/documents/edit.html.erb
```diff
@@ -44,7 +44,7 @@
           Add documents
         </button>
 
-        <%= link_to send_application_steps_path, class: "button" do %>
+        <%= link_to next_path, class: "button" do %>
```
jessieay: 😸

spec/controllers/success_controller_spec.rb
```diff
@@ -15,6 +15,14 @@
     end
   end
 
+  context "in order to not allow going back" do
```
jessieay: ✔️

jayroh:
> Just to confirm: is the flash message actually showing up after an email is sent? I was having issues getting it to show up on a "step" page.

It is, yeah. Here it is on the homepage after the redirect:



And I think maybe the flash not showing up might have to do with these being set on `flash` instead of `flash.now`. I'll take a look and confirm

jayroh:
Looks like the only place we're setting the flash is right before we redirect to the homepage. Should we be setting a message of some sort here?

1__tmux

In the else block?

jessieay:
@jayroh did you mean to close without merge?

PR #76, opened 2017-08-09
Refactor db storage of mailing address

jessieay:
* Use `addresses` table


app/models/snap_application.rb
```diff
@@ -1,3 +1,7 @@
 class SnapApplication < ApplicationRecord
   has_many :addresses
+
+  def mailing_address
+    addresses.where(mailing: true).first || NullAddress.new
```
jayroh: 🎉 NullAddress 🎉

jessieay:
@jayroh yeah, I was going to leave it with tons of duplication but rubocop was mad so I was shamed into creating the inheritance piece. could def be improved in future

PR #80, opened 2017-08-10
Set max width for step layout

jessieay:
* Same as GCF

screen shot 2017-08-10 at 10 05 54 am

Minor:

* Makes it easier to hide "debug" yellow bar in dev mode


jessieay:
[Adapt Michigan to include a max-width similar to GetCalFresh, but incorporating our blue navigation bar](https://trello.com/c/dIGuz6kh/95-adapt-michigan-to-include-a-max-width-similar-to-getcalfresh-but-incorporating-our-blue-navigation-bar)

config/secrets.yml
```diff
@@ -13,6 +13,7 @@
 development:
   secret_key_base: 04c0aa396262a7530a9f0ba4c979376fea7f43761a3e980cb1397eae7152f33ec41c3b118ce9da26542e19f02bedc89e4f2c06a861f8bda96bb42e2f626c6fa0
   shubox_js_id: '4711fd96-9be2-46d5-8fb9-6eef6fd24a4b'
+  debug: <%= ENV['DEBUG'] %>
```
jayroh: I don't know if I would put this here as it doesn't feel quite "secret"-y.

Thoughts on putting this in the base application controller and exposing it as a helper? That feels like a more representative location as it can then be used in both the controller and view layer.

app/assets/stylesheets/templates/_step.scss
```diff
@@ -6,4 +6,15 @@
...
+  .form-card {
+    max-width: 72rem;
+
+    @include media($tablet-up) {
+      margin: {
+        left: auto;
+        right: auto;
+      }
+    }
+  }
```
jayroh: What do you think about making this a more generic class that we can then apply to more elements than just the `.form-card`? I say this because I know we'll need it for the navigation, and eventually the flash messages

config/secrets.yml
jessieay: sure! I like that and I think it's much clearer than a secret.

 done at e43ce0e

app/assets/stylesheets/templates/_step.scss
jessieay: You mean create a new class named `max-width` or something and having these styles on that? 

I am totally down with that. In general, my only hesitation is that this is exactly how GCF does their max width and I'd like to copy the as much as possible: https://github.com/codeforamerica/gcf-backend/blob/master/app/assets/stylesheets/templates/_question.scss

what do you think?

jayroh: I feel you on that, but they have only one element that that rule applies to. Since we'll have at least 2 I think it's a worthwhile endeavor :)

jessieay: 🆒  added `max-container` at 9eaa2c6

jayroh:
This is good. ta. GO!

PR #82, opened 2017-08-10
Refactor flash-partial code

jayroh:
The existing display/dismiss flash partial code is tricky within the context of
the new michigan benefits app so I've made some changes to allow things to fit
in, for the time being, until there's a more robust visual treatment.

This commit changes things to loop through all flash keys and values and
display the corresponding html. It keeps the ability to dismiss the flash
notices albeit on a slightly superficial level (read: using the `:target`
pseudo-class in css).

flashes

Trello: https://trello.com/c/FPMNPuGm/53-refactor-flashes-code

### DIFF
--- a/app/views/shared/_flashes.html.erb
+++ b/app/views/shared/_flashes.html.erb
@@ -1,17 +1,10 @@
-<% if notice %>
-<div class="flash-container" id="dismiss-notice">
-  <div class="flash">
-    <p class="flash__message"><%= notice %></p>
-    <a href="#dismiss-notice" class="flash__dismiss"><span class="icon icon-close"></span></a>
+<% if flash.any? -%>
+  <div class="flash-container max-container" id="dismiss-notice">
+    <% flash.each do |key, value| %>
+      <div class="flash flash--<%= key %>">
+        <p class="flash__message"><%= value %></p>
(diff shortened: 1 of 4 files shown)


jayroh:
[Refactor "flashes" code](https://trello.com/c/FPMNPuGm/53-refactor-flashes-code)

app/views/shared/_flashes.html.erb
```diff
@@ -1,17 +1,10 @@
-<% if notice %>
-<div class="flash-container" id="dismiss-notice">
-  <div class="flash">
-    <p class="flash__message"><%= notice %></p>
-    <a href="#dismiss-notice" class="flash__dismiss"><span class="icon icon-close"></span></a>
+<% if flash.any? -%>
```
jessieay: YES so much better

jayroh:
> One question about our CSS approach (maybe something to discuss in retro?): if we are trying to copy and modify GCF styles, might it make sense to start writing all of our custom CSS in our own files and including them above the GCF copies? That is what I did when creating this Rails gem for assets that needed some overrides to work with Rails: https://github.com/18F/uswds-rails-gem

After yesterday's retro I think it can fall within the scope of us flowing in the entire GCF project's css when we get to do that. I imagine that's as soon as next week?

PR #83, opened 2017-08-11
Add personal detail page for app members

jessieay:
* Page asks for sex, marital status, SSN
* Introducing concept of "members" since we will soon be asking this
  for multiple people.

screen shot 2017-08-10 at 6 50 02 pm

### DIFF
--- a/app/controllers/introduce_yourself_controller.rb
+++ b/app/controllers/introduce_yourself_controller.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class IntroduceYourselfController < StepsController
+  def edit
+    current_app = current_or_new_snap_application
+
+    @step = step_class.new(
+      HashWithIndifferentAccess.
+        new(current_app.attributes).
+        slice(*step_attrs),
+    )
+  end
+
+  def update
+    @step = step_class.new(step_params)
+
+    if @step.valid?
+      app = current_or_new_snap_application
+      app.update!(step_params)
+      set_current_snap_application(app)
+      redirect_to(next_path)
+    else
+      render :edit
+    end
+  end
+end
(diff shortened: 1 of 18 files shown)


jessieay:
[As an MDHHS worker, I want to know more details about the person applying so I can identify them in my system.](https://trello.com/c/eP065iH9/96-as-an-mdhhs-worker-i-want-to-know-more-details-about-the-person-applying-so-i-can-identify-them-in-my-system)

jayroh:
We should consider bumping up the inclusion and setup of the `attr_encrypted` gem so items like SSN get encrypted on disk. Just a note to get that on our radar before we start taking on real applicants.

app/controllers/introduce_yourself_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class IntroduceYourselfController < StepsController
+  def edit
+    current_app = current_or_new_snap_application
+
+    @step = step_class.new(
+      HashWithIndifferentAccess.
+        new(current_app.attributes).
+        slice(*step_attrs),
+    )
```
jayroh: what do you think about extracting `HashWithIndifferentAccess.new(current_app.attributes).slice(*step_attrs)` out to something like?

```ruby
def existing_attributes  # or, like below, `snap_application_attributes`
  HashWithIndifferentAccess.
    new(current_app.attributes).
    slice(*step_attrs)
end
```

app/helpers/mb_form_builder.rb
```diff
@@ -1,10 +1,11 @@
 class MbFormBuilder < ActionView::Helpers::FormBuilder
-  def mb_input_field(method, label_text, type: "text", notes: [], options: {}, classes: [], prefix: nil, autofocus: nil)
+  def mb_input_field(method, label_text, type: "text", notes: [], options: {}, classes: [], prefix: nil, autofocus: nil, append_html: "")
     classes = classes.append(%w[text-input])
     <<-HTML.html_safe
       <fieldset class="form-group#{error_state(object, method)}">
         #{label_and_field(method, label_text, text_field(method, { autofocus: autofocus, type: type, class: classes.join(' '), autocomplete: 'off', autocorrect: 'off', autocapitalize: 'off', spellcheck: 'false' }.merge(options)), notes: notes, prefix: prefix)}
         #{errors_for(object, method)}
+        #{append_html}
```
jayroh: this is handy!

app/controllers/introduce_yourself_controller.rb
jessieay: ok I will do this in a follow up so we can have Alan accept today :)

PR #85, opened 2017-08-11
Split documents page in two

jayroh:
..., update copy, and enhance the usability of the actual document
upload page.

1. Do not allow submission of the form unless there have been files
   uploaded and saved to the form.
2. Do not allow submission of the form until files are done being
   uploaded.
3. Change button to say "Upload *more* documents" so that the client
   knows they CAN upload more files

documents guide page

* * *

document upload

* * *

documents uploaded

### DIFF
--- a/app/assets/javascripts/upload_setup.js
+++ b/app/assets/javascripts/upload_setup.js
@@ -11,6 +11,12 @@ var uploadSetup = function (previewTemplate) {
       window.alert(msg)
     },
     success: function (f) {
+      $('.button--next').removeAttr('disabled')
+
+      var button = $('#click-to-upload')
+      var buttonHtml = button.html()
+      button.html(buttonHtml.replace('Upload documents', 'Upload more documents'))
+
       var source = $('#form-card__documents__handlebars_template').html()
       var template = window.Handlebars.compile(source)
       var context = {
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -57,6 +57,11 @@ class StepsController < ApplicationController
     step_path(next_step, params) if next_step
   end
 
+  def skip_path(params = nil)
+    skip_step = step_navigation.skip
+    step_path(skip_step, params) if skip_step
+  end
+
   def step_navigation
     @step_navigation ||= StepNavigation.new(self)
   end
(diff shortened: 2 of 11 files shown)


jayroh:
[Tweak design / UX of "doc upload" page](https://trello.com/c/QGGiXPsv/93-tweak-design-ux-of-doc-upload-page)

app/controllers/steps_controller.rb
```diff
@@ -57,6 +57,11 @@ def next_path(params = nil)
     step_path(next_step, params) if next_step
   end
 
+  def skip_path(params = nil)
```
jessieay: neat!

app/assets/javascripts/upload_setup.js
```diff
@@ -11,6 +11,12 @@ var uploadSetup = function (previewTemplate) {
       window.alert(msg)
     },
     success: function (f) {
+      $('.button--next').removeAttr('disabled')
+
+      var button = $('#click-to-upload')
+      var buttonHtml = button.html()
+      button.html(buttonHtml.replace('Upload documents', 'Upload more documents'))
```
jessieay: this js isn't terribly complex, but it would be nice if we could test this behavior somehow so specs would fail it we mess up the js in the future. thoughts?

jessieay:
Pulled down the branch to get the user experience and I still think the 3 buttons on the doc upload page is kind of confusing. 

Perhaps an Alan question, but I wonder if we really need the "I'll do this later" button. 

Seems like the two actions on that page should be "upload document" and "done uploading documents". If someone wants to go back, they always have the back button in the nav (which GCF does not have). What do you think?

jessieay:
Ok, chatted with @alanjosephwilliams on Slack and I think we are in agreement:

On the page where doc upload is happening, we only want two options:

1. “Upload docs” (which becomes “Upload more docs” after first upload) 
2. “Done uploading documents” (same behavior as current "Continue" button)

These buttons can be next to each other similar to GCF's "Upload" and "Skip this step" buttons.

app/assets/javascripts/upload_setup.js
jayroh: Yeah I think it'd be valuable to add some JS testing tooling. Jasmine, chai, etc.

jayroh:
I think it's worth getting into staging and all of us taking a look at it together.

jessieay:
it looks like this was maybe merged to master? I am confused  https://github.com/codeforamerica/michigan-benefits/commit/0ec95069a1d5d5ffac546aae488eb28157b16ace

PR #88, opened 2017-08-11
Cleaner calls to action on the documents step

jayroh:
Reason for Change
===================
* A more clear collection of calls to action on the documents step
* Clear up copy

michigan_benefits

### DIFF
--- a/app/views/documents/edit.html.erb
+++ b/app/views/documents/edit.html.erb
@@ -18,11 +18,8 @@
       not a good time.
     </p>
     <p>
-      Clicking the "Upload documents button" below, or dragging and dropping
+      Clicking the "Upload documents" button below, or dragging and dropping
       photos or documents onto this area of the page, will upload them.
-    </p>
-    <p>
-      Clicking "Continue" will attach these documents to your application.
     </p>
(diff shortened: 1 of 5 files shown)


jayroh:
[Tweak design / UX of "doc upload" page](https://trello.com/c/QGGiXPsv/93-tweak-design-ux-of-doc-upload-page)

app/views/documents/edit.html.erb
```diff
@@ -48,12 +45,10 @@
           Upload documents
         </button>
 
-        <%= link_to next_path, class: "button" do %>
-          I'll do this later
-        <% end %>
+        <button class="button button--next button--cta button--full-width" type="submit">
+          Done uploading documents
```
jessieay: do we want to disable this before there are docs? or is that too annoying?

jayroh: If we disable it there's no way for them to continue without uploading docs (eg: "Skip")

jessieay: ah ok. I guess they could click the "back" button in the nav, but I see your point

PR #94, opened 2017-08-15
Use stylesheet framework from GCF

jayroh:
1. Remove current MB stylesheets
2. Add assets from GCF

   * Rename main.scss to application.scss
   * Clean all trailing spaces
   * Clean hard tabs, change to soft tabs (2 spaces)

3. To get things to look just right, we can override some of the existing
   styles and colors with some app-specific styles, and some updates in the
   variables to make it super easy to customize.

   * Some of the component specific colors have been abstracted one layer
     deep so that the variables are specific to the component, and the
     variables can be changed to suit the "theme".  Instead of telling a
     button it is blue, we tell it to use the defined "button color" and
     assign the right color to it in the variables.
   * michigan benefits specific styles have been namespaced within a
     `michigan-benefits` directory
   * Add components and organisms to get new michigan-benefits-specific
     elements to display correctly

mb-home

mb-intro


app/assets/stylesheets/application.scss
```diff
@@ -3,71 +3,75 @@
...
-@import 'organisms/pagination';
 @import 'organisms/legal';
-@import 'organisms/document-preview';
-@import 'organisms/demo-banner';
+@import 'organisms/main-footer';
+@import 'organisms/main-header';
+@import 'organisms/pagination';
+@import 'organisms/statistic-card';
 
+/* Michigan Benefits Overrides */
```
jessieay: should these below all of the other stuff in this file so it overrides in case we ever add override templates?

app/assets/stylesheets/atoms/_base.scss
```diff
@@ -6,7 +6,7 @@
 html {
   position: relative;
   background-color: $color-background;
-  font-size: $em-base;
```
jessieay: 🤑

app/assets/stylesheets/michigan-benefits/atoms/_base.scss
```diff
@@ -0,0 +1,4 @@
+.page-wrapper {
+  padding-top: $step-header-height;
+}
+
```
jessieay: extra newline

app/assets/stylesheets/michigan-benefits/atoms/_icons.scss
```diff
@@ -0,0 +1,7 @@
+.icon-below-the-fold {
+  @extend .icon-keyboard_arrow_down;
+  color: white;
+  font-size: $font-size-h1;
+  text-decoration: none;
+}
+
```
jessieay: extra newline

app/assets/stylesheets/michigan-benefits/organisms/_debug.scss
```diff
@@ -0,0 +1,9 @@
+.debug {
```
jessieay: 🐛

app/assets/stylesheets/michigan-benefits/organisms/_form-card.scss
```diff
@@ -0,0 +1,10 @@
+%max-container {
```
jessieay: do the base styles not have a `max-container` type thing we can use wholesale?

app/assets/stylesheets/michigan-benefits/organisms/_main-footer.scss
```diff
@@ -0,0 +1,14 @@
+.main-footer {
+  font-weight: 300;
+}
+
+.main-footer {
+  .illustration--cfa, .illustration--civilla {
+    float: left;
+    margin-top: 46px;
```
jessieay: would be ideal to not use such a specific margin height if we can (prefer rem or referencing a variable)

app/assets/stylesheets/michigan-benefits/organisms/_form-card.scss
jayroh: They don't! As far as I can tell.

app/assets/stylesheets/michigan-benefits/organisms/_main-footer.scss
jayroh: The specificity here  is justifiable, in my opinion, because it's referencing a very specific component. If this was something that was used across pages and across contexts, I would totally agree with you.

If this was extracted out to a variable it would be specific to this component - which would push the value down one level into a variable named *for* this component.

PR #95, opened 2017-08-15
Fix deprecation re: encrypted attribute

jayroh:
```
DEPRECATION WARNING: ssn is not an attribute known to Active Record.
This behavior is deprecated and will be removed in the next version of
Rails. If you'd like ssn to be managed by Active Record, add `attribute :ssn to your class.
```

Adding the attribute to the model fixes and removes this deprecation
warning.


jayroh:
WOOOO!

PR #97, opened 2017-08-15
Add employment status for household members

jessieay:
screen shot 2017-08-15 at 12 08 38 pm


jessieay:

app/controllers/many_member_steps_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+        attributes.
+        slice(*step_attrs).
+        merge(members: current_snap_application.members),
+    )
+  end
+
+  def member_attrs
+    raise NotImplementedError
+  end
+end
```
jayroh: If only one controller is subclassing this could we get away with placing the logic in this parent class into the child class?

app/helpers/steps_helper.rb
```diff
@@ -11,4 +11,9 @@ def icon(name)
   def data_md5(str)
     Digest::MD5.hexdigest(str ? str.squish : "")
   end
+
+  def header_name(household_member)
+    name = household_member.first_name.titleize
+    household_member.primary_member? ? "#{name} (that’s you!)" : name
+  end
```
jayroh: It me

app/views/shared/_question.html.erb
```diff
@@ -1,3 +0,0 @@
-(line 1, column 17):
-unexpected ":"
-expecting white space or "=>"
```
jayroh: Whoa!

spec/controllers/income_employment_status_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+require "rails_helper"
+
+RSpec.describe IncomeEmploymentStatusController do
+  let(:step) { assigns(:step) }
+  let(:member) do
+    create(:member, sex: "male", relationship: "Child", ssn: "12345")
+  end
+  let(:invalid_params) do
+    {
+      step: { members: { member.id => { employment_status: "" } } },
```
jayroh: Thought about breaking this up to multiple lines? Nested hashes, at least to me, are easier to scan when expanded out

app/views/income_employment_status/edit.html.erb
```diff
@@ -0,0 +1,32 @@
...
+        includes odd jobs, home businesses, online businesses, etc.
+      </p>
+    </div>
+  </header>
+
+  <div class="form-card__content">
+    <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
+      <% @step.members.each do |member| %>
+        <%= f.fields_for('members[]', member, hidden_field_id: true) do |ff| %>
+          <div class="household-member-group" data-md5="<%= data_md5(header_name(member))%>">
```
jayroh: What's this data attribute for/do? Just curious.

jessieay: it's basically just for the tests :) I was trying for about 30 min to target these elements based on the header name rather than some random attr, but it was not working. This is how the prototype had this implemented, so I just went with the "copy what they did" approach.

That being said, I can try to get it to work by selecting a section based on the h4 element again if you think this is too janky

app/views/shared/_question.html.erb
jessieay: I think that was the output from haml1erb not understanding the original file ;)

spec/controllers/income_employment_status_controller_spec.rb
jessieay: good idea! done at 168cbef

app/controllers/many_member_steps_controller.rb
jessieay: yeah for sure -- this is kind of premature optimization, but I was basing this code off of what we had in the prototype and I know for a fact that you are going to be subclassing off of this in your next feature so figured why not just keep as is?

jayroh: Ah-ha! Well put. If this is on the docket for quite a bit going forward, leave it as it. 👍

app/views/income_employment_status/edit.html.erb
jayroh: Would it work if you targeted `[data-member-name="John Doe"]` instead of the hashed string? I think it would?

so 

```
<div class="household-member-group" data-md5="<%= header_name(member) %>">
```

app/views/shared/_question.html.erb
jayroh: I figured as much. Just jumped out at me! Scared me!

👻  boo!

app/views/income_employment_status/edit.html.erb
jessieay: done at 8a1dfac 🍾

PR #98, opened 2017-08-16
Collect employment details for employed and self employed

jessieay:
screen shot 2017-08-15 at 5 32 04 pm

### DIFF
--- a/app/controllers/income_details_per_member_controller.rb
+++ b/app/controllers/income_details_per_member_controller.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class IncomeDetailsPerMemberController < ManyMemberStepsController
+  private
+
+  def skip?
+    current_snap_application.
+      members.
+      none? { |member| member.employed? || member.self_employed? }
+  end
+
+  def member_attrs
+    %i[
+      employed_employer_name
+      employed_hours_per_week
+      employed_pay_interval
+      employed_pay_quantity
+      self_employed_monthly_expenses
+      self_employed_monthly_income
+      self_employed_profession
+    ]
+  end
+end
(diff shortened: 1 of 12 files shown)


jessieay:

jessieay:
needs some design love, but figure we can ship as-is

app/controllers/income_details_per_member_controller.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class IncomeDetailsPerMemberController < ManyMemberStepsController
+  private
+
+  def skip?
+    current_snap_application.
+      members.
+      none? { |member| member.employed? || member.self_employed? }
```
jayroh: > skip if no members are employed or self-employed

Is that right?

I'm asking because I'm wondering if there's any conditional that might be the opposite approach to the `none?`. For some reason unpacking `none?`, like `unless`, makes my brain bend a little bit.

spec/controllers/income_details_per_member_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  describe "#update" do
+    context "when valid" do
+      it "updates the member attributes if they are present" do
+        params = {
+          step: {
+            members: {
+              member.id => { self_employed_profession: "Baseball hat" },
+            },
+          },
+        }
```
jayroh: 😍

app/controllers/income_details_per_member_controller.rb
jessieay: yeah, we only show questions for employed or self employed people, so this could also be phrased as "if all members are unemployed"

jessieay: updated logic at  7cdd9e1 ✔️

jayroh:
the little refactor to `&:not_employed` is *perfect*

Thanks Jessie :)

PR #99, opened 2017-08-16
Add "additional sources of income" page

jessieay:
screen shot 2017-08-16 at 10 16 20 am

screen shot 2017-08-16 at 11 08 18 am

### DIFF
--- a/app/controllers/income_additional_controller.rb
+++ b/app/controllers/income_additional_controller.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+class IncomeAdditionalController < StandardStepsController
+  def edit
+    @additional_income = current_snap_application.additional_income.map do |key|
+      "income_#{key}"
+    end
+
+    @income_fields = income_fields(@additional_income)
+
+    @step = step_class.new(
+      @additional_income.map { |k| [k, current_snap_application[k]] }.to_h,
+    )
+  end
+
+  private
+
+  def skip?
+    no_additional_income_sources?
+  end
+
+  def no_additional_income_sources?
+    current_snap_application.additional_income.empty?
+  end
+
+  def income_fields(additional_income)
+    {
+      income_unemployment_insurance: "Unemployment Insurance",
+      income_ssi_or_disability: "SSI or Disability",
+      income_workers_compensation: "Worker's Compensation",
+      income_pension: "Pension",
+      income_social_security: "Social Security",
+      income_child_support: "Child Support",
+      income_foster_care: "Foster Care or Adoption Subsidies",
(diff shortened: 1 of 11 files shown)


jessieay:

app/views/income_additional/edit.html.erb
```diff
@@ -0,0 +1,33 @@
...
+      income_child_support: "Child Support",
+      income_foster_care: "Foster Care or Adoption Subsidies",
+      income_other: "Other Income"
+    }.stringify_keys.slice(*@additional_income) %>
+
+    <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
+
+      <% income_fields.each do |method, label| %>
+        <%= f.mb_input_field method, label, notes: ["Monthly amount"] %>
+      <% end %>
```
jayroh: My personal opinion is to explicitly put the input field tags here - which is what I did in my current branch.

On the "feels strongly about" scale, from 1 to 10 (10 being "feel VERY strongly about") I'd probably put that at a 5.

So I'll defer to you here, @jessieay. What do you think? Whichever way we land is how I'll handle this on my side too.

I do think putting the variable `income_fields` into the controller as an ivar would be a nice change though. I do feel pretty strongly that views should have view code, and not a whole lot of ruby/logic.

app/controllers/income_additional_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  private
+
+  def skip?
+    step_attrs.none? do |attr|
+      current_snap_application.
+        additional_income.
+        include?(attr.to_s.remove(/^income_/))
+    end
+  end
```
jayroh: What's your take on delegating to an additional, descriptive, well-named method to make this even easier to read?

Something like how I'm doing it over here: https://github.com/codeforamerica/michigan-benefits/pull/100/files#diff-7a0580538493197e22bd6c89261ac0a1R76

I had to do some digging on my branch over there to figure out *why* something necessitated skipping and it would have been helpful for me if I had known what the original developer in that controller was looking for.

app/views/income_additional/edit.html.erb
jessieay: Moved the ruby code to the controller here: 6322f2a

still iterating since the fields we are showing is dynamic (so hardcoding would require a lot of ruby conditionals in the view) but I think this is a clarifying improvement nonetheless!

app/controllers/income_additional_controller.rb
jessieay: great idea! delegated and simplified at 11e176a

jayroh:
WOO!

PR #100, opened 2017-08-16
Collect information about expenses

jayroh:
Work in progress branch. More to come!

### DIFF
--- a/app/controllers/expenses_additional_controller.rb
+++ b/app/controllers/expenses_additional_controller.rb
@@ -1,0 +1,126 @@
+# frozen_string_literal: true
+
+class ExpensesAdditionalController < StepsController
+  PREVIOUS_SECTION_ATTRIBUTES = %w[
+    court_ordered
+    dependent_care
+    medical
+    tax_deductible
+  ].freeze
+
+  ATTRIBUTES = %w[
+    monthly_care_expenses
+    monthly_medical_expenses
+    monthly_court_ordered_expenses
+    monthly_tax_deductible_expenses
+  ].freeze
+
+  ARRAY_ATTRIBUTES = {
+    care_expenses: %w[
+      childcare
+      elderly_care
+      disabled_adult_care
+    ],
+    medical_expenses: %w[
+      health_insurance
+      co_pays
+      prescriptions
+      dental
+      in_home_care
+      transportation
+      hospital_bills
+      other
+    ],
+    court_ordered_expenses: %w[
+      child_support
+      alimony
+    ],
+    tax_deductible_expenses: %w[
+      student_loan_interest
+      other_tax_deductible
+    ],
+  }.freeze
+
+  def edit
+    set_sections
+
+    @step = step_class.new(
+      current_snap_application.
+        attributes.
+        slice(*ATTRIBUTES).
+        merge(array_to_checkboxes(attribute_array)),
+    )
+  end
+
+  def update
+    update_params = step_params.
+      slice(*ATTRIBUTES).
+      merge(attribute_hash)
+    @step = step_class.new(update_params)
+
+    if @step.valid?
+      current_snap_application.update!(update_params)
+      redirect_to next_path
+    else
+      render :edit
+    end
+  end
+
+  private
+
+  def existing_attributes
+    HashWithIndifferentAccess.new(application_attributes)
+  end
+
+  def application_attributes
+    app = current_snap_application
+
+    {
+      monthly_care_expenses: app.monthly_care_expenses,
+      monthly_medical_expenses: app.monthly_medical_expenses,
+      monthly_court_ordered_expenses: app.monthly_court_ordered_expenses,
+      monthly_tax_deductible_expenses: app.monthly_tax_deductible_expenses,
+      care_expenses: app.care_expenses,
+      medical_expenses: app.medical_expenses,
+      court_ordered_expenses: app.court_ordered_expenses,
+      tax_deductible_expenses: app.tax_deductible_expenses,
+    }
+  end
+
+  def attribute_array
+    ARRAY_ATTRIBUTES.keys.flat_map do |key|
+      current_snap_application[key]
+    end
+  end
+
+  def attribute_hash
+    ARRAY_ATTRIBUTES.map do |k, c|
+      [k, checkboxes_to_array(c)]
+    end.to_h
+  end
+
+  def array_to_checkboxes(array)
+    array.map { |k| [k, true] }.to_h
+  end
+
+  def checkboxes_to_array(checkboxes)
+    checkboxes.select { |k| step_params[k].in?(["1", 1, true]) }.map(&:to_s)
+  end
+
+  def skip?
+    @_skip ||= previous_page_all_false?
+  end
+
+  def previous_page_all_false?
+    previous_section.values.all?(&:!)
+  end
+
+  def previous_section
+    @previous_section ||= current_snap_application.
+      attributes.
+      slice(*PREVIOUS_SECTION_ATTRIBUTES).
+      symbolize_keys
+  end
+
+  alias set_sections previous_section
+end
--- a/app/controllers/expenses_additional_sources_controller.rb
+++ b/app/controllers/expenses_additional_sources_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class ExpensesAdditionalSourcesController < StandardStepsController
+  private
+
+  def step_params
+    super.tap do |params_hash|
+      params_hash.each do |key, val|
+        params_hash[key] = if val == "true"
+                             true
+                           else
+                             false
+                           end
+      end
+    end
+  end
+end
--- a/app/controllers/expenses_introduction_controller.rb
+++ b/app/controllers/expenses_introduction_controller.rb
@@ -1,7 +1,4 @@
 # frozen_string_literal: true
 
 class ExpensesIntroductionController < StepsController
-  def update
-    redirect_to(next_path)
-  end
 end
--- a/app/steps/expenses_additional.rb
+++ b/app/steps/expenses_additional.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class ExpensesAdditional < Step
+  step_attributes(
+    :alimony,
+    :care_expenses,
+    :child_support,
+    :childcare,
+    :co_pays,
+    :court_ordered_expenses,
+    :dental,
+    :disabled_adult_care,
+    :elderly_care,
+    :health_insurance,
+    :hospital_bills,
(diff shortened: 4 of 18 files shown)


jayroh:
[As a client, I want to provide information about my expenses to maximize the benefits I will receive](https://trello.com/c/MrLrwn3O/109-as-a-client-i-want-to-provide-information-about-my-expenses-to-maximize-the-benefits-i-will-receive)

app/controllers/expenses_additional_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+  def checkboxes_to_array(checkboxes)
+    checkboxes.select { |k| step_params[k].in?(["1", 1, true]) }.map(&:to_s)
+  end
+
+  def skip?
+    @_skip ||= previous_page_all_false?
+  end
+
+  def previous_page_all_false?
+    sections.values.all?(&:!)
```
jessieay: oooh fancy

app/controllers/expenses_additional_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  def sections
+    @sections ||= current_snap_application.
+      attributes.
+      slice(
+        "court_ordered",
```
jessieay: should these go into a constant too?

app/controllers/expenses_additional_sources_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ExpensesAdditionalSourcesController < StandardStepsController
+  private
+
+  def step_params
+    super.tap do |params_hash|
+      params_hash.each do |key, val|
+        params_hash[key] = val == "true" ? true : false
```
jessieay: this might be easier to read in a non-ternary format.

app/controllers/expenses_introduction_controller.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+class ExpensesIntroductionController < StepsController
```
jessieay: when I created this controller and the other "static" pages controllers I overrode `update` but I guess I didn't have to! *magic* ✨

app/steps/expenses_additional.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class ExpensesAdditional < Step
+  step_attributes \
```
jessieay: super tiny stylistic thing: in other step classes we've been putting parens around step_attributes attributes rather than using `\` -- also, alpha order?

app/views/expenses_additional/edit.html.erb
```diff
@@ -0,0 +1,78 @@
...
+      <div class="form-card__content">
+        <h4 class="step-section-header__headline">
+          Care expenses
+        </h4>
+
+        <%= f.mb_input_field :monthly_care_expenses,
+          "In total, how much do you pay in care expenses each month?" %>
+
+        <%= f.mb_checkbox_set [
+          { method: :childcare,           label: 'Childcare'           },
```
jessieay: not sure I'm on board with this type of spacing for hashes, but I could be convinced. 💟

app/views/expenses_additional/edit.html.erb
```diff
@@ -0,0 +1,78 @@
...
+          { method: :elderly_care,        label: 'Elderly Care'        },
+          { method: :disabled_adult_care, label: 'Disabled adult care' },
+          ], label_text: 'What kind of care expenses do you have?' %>
```
jessieay: seems like there's some inconsistency in single/double quotes in this file. Think we're doing double?

db/migrate/20170816143355_add_expense_columns_to_snap_application.rb
```diff
@@ -0,0 +1,29 @@
...
+    add_column :snap_applications, :utility_heat, :boolean
+    add_column :snap_applications, :utility_cooling, :boolean
+    add_column :snap_applications, :utility_electrity, :boolean
+    add_column :snap_applications, :utility_water_sewer, :boolean
+    add_column :snap_applications, :utility_trash, :boolean
+    add_column :snap_applications, :utility_phone, :boolean
+    add_column :snap_applications, :utility_other, :boolean
+
+    add_column :snap_applications, :dependent_care, :boolean
+    add_column :snap_applications, :medical, :boolean
```
jessieay: I feel like a bool field called `medical` is not super clear -- thoughts on a more descriptive name for this column? (feel similarly about `dependent_card` `court_ordered` and `tax_deductible`)

app/steps/expenses_additional_source.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class ExpensesAdditionalSource < Step
+  step_attributes \
+    :dependent_care,
+    :medical,
+    :court_ordered,
+    :tax_deductible
+
+  validates \
```
jessieay: should we have a controller spec to test this validation?

app/controllers/expenses_additional_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    tax_deductible_expenses: %w[
+      student_loan_interest
+      other_tax_deductible
+    ],
+  }.freeze
+
+  def edit
+    set_sections
+
+    attribute_array = ARRAY_ATTRIBUTES.keys.flat_map do |key|
```
jessieay: maybe this should happen in a well-named private method? there is enough happening in these controller methods that I am getting a but turned around while reading them

app/controllers/expenses_additional_sources_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ExpensesAdditionalSourcesController < StandardStepsController
+  private
+
```
jessieay: we've been defining the `existing_attributes` method to ensure that attrs are filled in on the page when someone returns. Is that functionality working here? Perhaps a controller spec would be a good way to confirm

db/migrate/20170816143355_add_expense_columns_to_snap_application.rb
jessieay: even `medical_expenses` as a boolean would be clearer! not sure we need `has`

app/controllers/expenses_additional_sources_controller.rb
jayroh: @jessieay what's the rule for the `existing_attributes` method? Is the goal to provide a hash with the existing model's current state as it pertains to *this* action?  So for here I would be adding?

```ruby
  def existing_attributes
    HashWithIndifferentAccess.new(application_attributes)
  end

  def application_attributes
    app = current_snap_application

    {
      monthly_care_expenses: app.monthly_care_expenses,
      monthly_medical_expenses: app.monthly_medical_expenses,
      monthly_court_ordered_expenses: app.monthly_court_ordered_expenses,
      monthly_tax_deductible_expenses: app.monthly_tax_deductible_expenses,
      care_expenses: app.care_expenses,
      medical_expenses: app.medical_expenses,
      court_ordered_expenses: app.court_ordered_expenses,
      tax_deductible_expenses: app.tax_deductible_expenses,
    }
  end
```

Is that right?  And is that tested with that shared example?

app/steps/expenses_additional_source.rb
jessieay: yeah unit test works, too! just wanted to make sure we have *something* testing this validation

app/controllers/expenses_additional_sources_controller.rb
jessieay: yeah, we've been testing this at the controller level so far. Example: https://github.com/codeforamerica/michigan-benefits/blob/13c20027a88425746d695df75abee0040a158da2/spec/controllers/household_add_member_controller_spec.rb#L14

db/migrate/20170816143355_add_expense_columns_to_snap_application.rb
jayroh: There's already a `medical_expenses` column (an array of the actual medical expenses they're incurring) versus `medical` (the bool)

app/views/expenses_introduction/edit.html.erb
```diff
@@ -1,36 +1,35 @@
...
-<div class='form-card'>
-  <header class='form-card__header'>
-    <div class='step-section-header'>
-      <div class="step-section-header__icon illustration illustration--section-3"></div>
-      <div class='step-section-header__headline'>You're halfway there!</div>
-      <div class='step-section-header__subhead'>
+<div class="form-card">
+  <header class="form-card__header">
+    <div class="step-section-header">
+      <% icon 'section-3' %>
```
jessieay: this `icon` helper does not exist in this codebase, needs to be `<div class="step-section-header__icon illustration illustration--section-3"></div>` per diff

db/migrate/20170816143355_add_expense_columns_to_snap_application.rb
jessieay: hmmm ok, let's potentially revisit this soon, then. I am just noticing as I fill in the PDF that it is hard to sometimes tell which DB fields match to which PDF fields. Hopefully we can get to a place naming-wise where it is more obvious without having to dig through the views and controllers.

spec/controllers/expenses_additional_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ExpensesAdditionalController, type: :controller do
+  let(:step) { assigns(:step) }
+  let(:step_class) { ExpensesAdditional }
+  before {}
```
jessieay: surprised we don't need a `current_snap_apploication` set for this page to render?

jessieay: oh I see -- we are setting in each spec. In other controller specs we've been doing this via a `before` action. I generally don't love `before` but it kinda makes sense here since they all need it

spec/steps/expenses_additional_source_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    )
+
+    valid_falses_step = ExpensesAdditionalSource.new(
+      court_ordered: false,
+      dependent_care: false,
+      medical: false,
+      tax_deductible: false,
+    )
+
+    invalid_step = ExpensesAdditionalSource.new(
```
jessieay: could also test for `nil` values if you thought that was valuable

spec/controllers/expenses_additional_controller_spec.rb
jayroh: d'oh. Shoulda removed that 😊

PR #104, opened 2017-08-17
Add google analytics code to application layout

jayroh:


jayroh:
[Set up Google Analytics before we acquire our first client](https://trello.com/c/0kZcazLa/110-set-up-google-analytics-before-we-acquire-our-first-client)

app/views/shared/_google_analytics.html.erb
```diff
@@ -0,0 +1,9 @@
+<script>
```
jessieay: should we put an `if` around this so it only runs in prod? 

Also, staging and prod are both `production` so maybe we could just set the `property Id` via an env var so we aren't accidentally tracking metrics for both under the same account?

jayroh: Excellent suggestion!

PR #105, opened 2017-08-17
Fill in more fields in DC 1171 field

jessieay:
* Mailing / residential address
* Personal details for primary member
* Next up: details for other household members


jessieay:
[Fill in DC 1171 with data](https://trello.com/c/W9Cjeuyb/116-fill-in-dc-1171-with-data)

app/controllers/introduce_yourself_controller.rb
```diff
@@ -6,7 +6,7 @@ def update
 
     if @step.valid?
       app = current_or_new_snap_application
-      app.update!(snap_application_update_params)
+      app.save!
```
jayroh: So this no longer needs to have any new values assigned to it in this step?

spec/models/dhs1171_pdf_spec.rb
```diff
@@ -39,7 +70,8 @@
     end
 
     it "returns the tempfile" do
-      snap_application = create(:snap_application)
+      member = create(:member)
+      snap_application = create(:snap_application, members: [member])
```
jayroh: Could you use `:with_member` here (and in other spots) where you are assigning it explicitly?

app/controllers/introduce_yourself_controller.rb
jessieay: yeah, before we were assigning the birthday to the `snap_application`, but now we want it on the `member` instead. This didn't become clear until I started filling out the PDF and realized we were being inconsistent about birthday. For regular members we were assigning birthday to the `member`, but for the primary member we were getting it from the `snap_application`

jayroh: ya-haaaaa! That's a nice change :)

spec/models/dhs1171_pdf_spec.rb
jessieay: sure! done at b7bd776

jayroh:
Yeah this is great to go, @jessieay  :)

jessieay:
merged at 20ab7be938a99e3e19f1d2efbb5f53046345dc65

PR #108, opened 2017-08-18
Add household member data to 1171 PDF

jessieay:
* For up to 6 members.

screen shot 2017-08-17 at 9 04 09 pm


app/models/dhs1171_pdf_member_attributes.rb
```diff
@@ -0,0 +1,49 @@
+class Dhs1171PdfMemberAttributes
```
jayroh: I can surmise that Dhs21171 is the form name - haha, that's a wild guess! But I'm wondering what you think about a more generic name that encompasses what the form *is*. It's the snap application for our purposes, so could we name these classes after that instead of the formal, esoteric, Dhs 1171?

spec/models/dhs1171_pdf_member_attributes_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        primary_member_marital_status_widowed: nil,
+        primary_member_new_mom_no: nil,
+        primary_member_new_mom_yes: "Yes",
+        primary_member_relationship: nil,
+        primary_member_requesting_food_assistance: "Yes",
+        primary_member_sex_female: "Yes",
+        primary_member_sex_male: nil,
+        primary_member_ssn: "222-22-2222",
+      )
+    end
```
jayroh: would it be helpful to test an additional (non-primary) members' results in the result hash?

jessieay: yeah! will do that this morning as I add *even more* fields to this lovely PDF

PR #112, opened 2017-08-18
Updates to expenses section

jessieay:
* Remove fields for "Tax deductible"
* Only show follow up questions for expenses indicated (Eg: if no
medical expenses, don't ask for types of medical expenses)

### DIFF
--- a/app/controllers/expenses_additional_controller.rb
+++ b/app/controllers/expenses_additional_controller.rb
@@ -5,14 +5,12 @@ class ExpensesAdditionalController < StepsController
     court_ordered
     dependent_care
     medical
-    tax_deductible
   ].freeze
 
   ATTRIBUTES = %w[
     monthly_care_expenses
     monthly_medical_expenses
     monthly_court_ordered_expenses
-    monthly_tax_deductible_expenses
   ].freeze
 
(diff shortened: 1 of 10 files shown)


app/controllers/expenses_additional_controller.rb
```diff
@@ -79,11 +73,9 @@ def application_attributes
       monthly_care_expenses: app.monthly_care_expenses,
       monthly_medical_expenses: app.monthly_medical_expenses,
       monthly_court_ordered_expenses: app.monthly_court_ordered_expenses,
-      monthly_tax_deductible_expenses: app.monthly_tax_deductible_expenses,
       care_expenses: app.care_expenses,
       medical_expenses: app.medical_expenses,
       court_ordered_expenses: app.court_ordered_expenses,
-      tax_deductible_expenses: app.tax_deductible_expenses,
     }
```
adarsh: I'm probably way off but I wonder: could you call something like `current_snap_application.attributes` and get the same hash?

jessieay: Realized when taking a second look per your feedback that this is unused code so deleted at 86d3ab4 📉  -- good eye!

PR #113, opened 2017-08-18
Refactor PDF filling code

jessieay:
* Rename class for member attributes
* Add class for snap app attributes
* Add spec for multiple household members

### DIFF
--- a/app/models/dhs1171_pdf.rb
+++ b/app/models/dhs1171_pdf.rb
@@ -29,32 +29,19 @@ class Dhs1171Pdf
   end
 
   def client_data
-    {
-      applying_for_food_assistance: "Yes",
-      birth_day: primary_member.birthday.strftime("%d"),
-      birth_month: primary_member.birthday.strftime("%m"),
-      birth_year: primary_member.birthday.strftime("%Y"),
-      phone_number: snap_application.phone_number,
-      mailing_address_street_address:
-        snap_application.mailing_address.street_address,
-      mailing_address_city: snap_application.mailing_address.city,
-      mailing_address_county: snap_application.mailing_address.county,
-      mailing_address_state: snap_application.mailing_address.state,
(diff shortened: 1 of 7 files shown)


app/models/dhs1171_pdf.rb
```diff
@@ -70,21 +57,9 @@ def complete_form_with_cover
...
-      if attrs[:member].present?
-        Dhs1171PdfMemberAttributes.new(attrs).to_h
-      end
-    end.compact.reduce({}, :merge)
-  end
-
   def first_six_members
     [
-      { member: primary_member, position: "primary" },
+      { member: application_members[0], position: "primary" },
```
jayroh: 👍

app/models/snap_application_attributes.rb
```diff
@@ -0,0 +1,77 @@
...
+  def buy_food_with
+    snap_application.members.map(&:buy_food_with?)
+  end
+
+  def members_not_buy_food_with
+    snap_application.
+      members.
+      where.
+      not(buy_food_with: true).
+      map(&:full_name).
```
jayroh: Instead of using `map` here, if you used `pluck` it would do a `select full_name from ...` instead of `select * from ...` and then pulling them out with the `map`. Especially useful in a model with so many columns like `snap_applications`

app/models/snap_application_member_attributes.rb
```diff
@@ -1,4 +1,4 @@
-class Dhs1171PdfMemberAttributes
+class SnapApplicationMemberAttributes
```
jayroh: Niiiice

app/models/snap_application_attributes.rb
jessieay: great suggestion! done at  7da0460

jessieay: and done for realiz at f3f7063

PR #115, opened 2017-08-21
Initial setup of admin (Administrate) rubygem

jayroh:
Get the admin set up to view snap applications. This is the first step
to getting an admin spun up and configured as a foundation.

### DIFF
--- a/app/controllers/admin/snap_applications_controller.rb
+++ b/app/controllers/admin/snap_applications_controller.rb
@@ -1,0 +1,21 @@
+module Admin
+  class SnapApplicationsController < Admin::ApplicationController
+    # To customize the behavior of this controller,
+    # you can overwrite any of the RESTful actions. For example:
+    #
+    # def index
+    #   super
+    #   @resources = SnapApplication.
+    #     page(params[:page]).
+    #     per(10)
+    # end
+
+    # Define a custom finder by overriding the `find_resource` method:
+    # def find_resource(param)
+    #   SnapApplication.find_by!(slug: param)
+    # end
+
+    # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
+    # for more information
+  end
+end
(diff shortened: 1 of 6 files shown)


app/controllers/admin/snap_applications_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Admin
+  class SnapApplicationsController < Admin::ApplicationController
+    # To customize the behavior of this controller,
```
jessieay: remove commented scaffolding?

app/dashboards/snap_application_dashboard.rb
```diff
@@ -0,0 +1,195 @@
+require "administrate/base_dashboard"
+
+class SnapApplicationDashboard < Administrate::BaseDashboard
+  # ATTRIBUTE_TYPES
```
jessieay: same comment as above -- do we need these? on the one hand it's a nice shortcut for looking at docs, but it also makes the code harder to read because the file is longer. Your call!

config/routes.rb
```diff
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
+  namespace :admin do
```
jessieay: should we somehow make sure that only admins can see this route? (not sure if we are going to add an auth system or not)

app/controllers/admin/snap_applications_controller.rb
jayroh: For the controller, I think it's worthwhile to keep these in. It's the one layer in this gem that I end up touching more often than the others so it's helpful to know exactly what/where changes are often made.

app/dashboards/snap_application_dashboard.rb
jayroh: Most of these can go, yeah!

config/routes.rb
jayroh: In the future - absolutely. For now since we have no concept of an admin user (just the SNAP applicants) we have no way to tell whether someone is an admin.

As a stop-gap I've put some ENV variables that we can use for basic auth username and password. They've also been added to heroku's config on both staging AND production. So if/when you would like to fire up the admin in either of those two environments, do a `heroku config | grep ADMIN_` (or along those lines) to get that info :)

PR #116, opened 2017-08-21
Three small but distinct changes

jessieay:
1. Set default to be "mailing address same as residential" (true)
2. Change FAB to FAP (typo)
3. Reverse order of contents on success page so "what to expect next"
  is more prominent (email section below)

success page:

screen shot 2017-08-21 at 10 41 29 am


jessieay:
[Alan needs to create three cards](https://trello.com/c/6ocjV8VP/112-alan-needs-to-create-three-cards)

PR #117, opened 2017-08-21
Conditional validation of residential address

jayroh:
If the client has an unstable or volatile home address the application
should *not* validate the presence of their home address. If they don't
have a stable address, allow them to continue without having to fill in
those fields.

ALSO:

* Small formatting refactor in the mb_form_builder to make the text
  input a little easier to read.
* Limit the number of chars in zip code form field as we validate it's
  length in the form object. We can match that restriction on the view
  layer.
* When validating presence of fields in the address form object, do not
  allow empty strings

### DIFF
--- a/app/helpers/mb_form_builder.rb
+++ b/app/helpers/mb_form_builder.rb
@@ -11,9 +11,20 @@ class MbFormBuilder < ActionView::Helpers::FormBuilder
     append_html: ""
   )
     classes = classes.append(%w[text-input])
+    text_field_options = {
+      autofocus: autofocus,
+      type: type,
+      class: classes.join(" "),
+      autocomplete: "off",
+      autocorrect: "off",
+      autocapitalize: "off",
+      spellcheck: "false",
+    }.merge(options)
+    text_field_html = text_field(method, text_field_options)
+
     <<-HTML.html_safe
       <fieldset class="form-group#{error_state(object, method)}">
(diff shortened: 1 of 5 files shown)


app/steps/residential_address.rb
```diff
@@ -10,18 +10,35 @@ class ResidentialAddress < Step
     :zip,
   )
 
-  validates :street_address,
-    presence: { message: "Make sure to provide a street address" }
+  with_options if: :stable_housing? do |address|
+    address.validates :street_address,
+      presence: {
+        message: "Make sure to provide a street address",
+        allow_blank: false,
```
jessieay: do you need an `allow_blank` for it to reject empty strings?

app/helpers/mb_form_builder.rb
```diff
@@ -11,9 +11,20 @@ def mb_input_field(
     append_html: ""
   )
     classes = classes.append(%w[text-input])
+    text_field_options = {
```
jessieay: nice!

jayroh:
[If the client chooses that they do not have a stable address, do not validate/require a home address that includes Address, City, ZIP](https://trello.com/c/Bbt1Wt9p/106-if-the-client-chooses-that-they-do-not-have-a-stable-address-do-not-validate-require-a-home-address-that-includes-address-city-z)

app/steps/residential_address.rb
jayroh: It defaults to `true` if you leave this out. You need to set this to false if you want to reject empty strings. Noticed this when I was going back and forth between this page, and the next page. Hard to explain what I saw - but consider this, the first visit to this page before submitting the values will be `nil`. Subsequent visits will be `""` so there's a clear difference in behavior unless you tell this here to NOT allow blanks.

PR #118, opened 2017-08-21
Update a host of outdated gems

jayroh:
Noticed there were a number of outdated gems, Rails being one of them,
so this PR is a collection of commits that, before merging, will be squashed.

Full test suite runs verify that things remain passing.


greysteil:
Would you be up for automating dependency updates? We built and used [Dependabot](https://dependabot.com) at the startup I used to work at - it puts in PRs each morning for anything out-of-date.

PR #119, opened 2017-08-21
Redirect home if no application is in session

jayroh:
If someone tries to go to one of the steps (steps that are NOT the very
first step) then they should be redirected to the homepage. Once they
start the application process and create a new application, they will be
able to go back and forth along all of the steps.

### DIFF
--- a/app/controllers/introduce_yourself_controller.rb
+++ b/app/controllers/introduce_yourself_controller.rb
@@ -17,6 +17,10 @@ class IntroduceYourselfController < StandardStepsController
 
   private
 
+  def ensure_application_present
+    _application_not_created = true
+  end
+
   def existing_attributes
     HashWithIndifferentAccess.new(member.attributes)
   end
--- a/app/controllers/residential_address_controller.rb
+++ b/app/controllers/residential_address_controller.rb
@@ -21,6 +21,6 @@ class ResidentialAddressController < AddressController
   end
 
   def skip?
-    current_snap_application.mailing_address_same_as_residential_address?
+    current_snap_application&.mailing_address_same_as_residential_address?
   end
 end
--- a/spec/support/shared_examples/step_controller.rb
+++ b/spec/support/shared_examples/step_controller.rb
@@ -9,6 +9,18 @@ RSpec.shared_examples "step controller" do
 
       expect(step).to be_an_instance_of step_class
     end
+
+    context "no application is currently under way" do
+      it "redirects to the homepage" do
+        if subject.class != IntroduceYourselfController
+          session[:snap_application_id] = nil
+
+          get :edit
+
+          expect(response).to redirect_to(root_path)
+        end
+      end
+    end
   end
 
   describe "#update" do
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -4,6 +4,7 @@ class StepsController < ApplicationController
   layout "step"
 
   before_action :maybe_skip, only: :edit
+  before_action :ensure_application_present, only: :edit
 
   def self.to_param
     controller_path.dasherize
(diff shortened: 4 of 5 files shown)


app/controllers/residential_address_controller.rb
```diff
@@ -21,6 +21,6 @@ def snap_application_param
   end
 
   def skip?
-    current_snap_application.mailing_address_same_as_residential_address?
+    current_snap_application&.mailing_address_same_as_residential_address?
```
jayroh: use the little lonely operator to check whether or not `current_snap_application` is nil or not before sending that next method.

app/controllers/introduce_yourself_controller.rb
```diff
@@ -17,6 +17,10 @@ def update
 
   private
 
+  def ensure_application_present
+    _application_not_created = true
+  end
```
jayroh: Instead of just placing `true` here I put the assignment to a descriptive variable instead. COULD delegate to another method but felt this was more direct and to the point. Up for discussion :)

spec/support/shared_examples/step_controller.rb
```diff
@@ -9,6 +9,18 @@
...
+
+    context "no application is currently under way" do
+      it "redirects to the homepage" do
+        if subject.class != IntroduceYourselfController
+          session[:snap_application_id] = nil
+
+          get :edit
+
+          expect(response).to redirect_to(root_path)
+        end
```
jayroh: On the fence about wrapping this conditional around here to exclude just that one controller.

jayroh:
[Redirect the client when the session is cleared.](https://trello.com/c/piFTSltK/102-redirect-the-client-when-the-session-is-cleared)

app/controllers/residential_address_controller.rb
jessieay: should we use this in other places where we're skipping?

app/controllers/introduce_yourself_controller.rb
jessieay: it would be possible to go `back` and get here with a current snap app, so maybe it doesn't need its own implementation?

spec/support/shared_examples/step_controller.rb
jessieay: per my comment above, not sure we need to! Think the logic should be the same everywhere

app/controllers/residential_address_controller.rb
jayroh: I'm guessing yes(?), but the tests surfaced this as the only place where this caused an issue.

app/controllers/introduce_yourself_controller.rb
jayroh: It's not a problem a current snap application exists, that's fine. It's when it doesn't exist yet (which is only on the first visit to this page).

jayroh: It's not a problem a current snap application exists, that's fine. It's when it doesn't exist yet (which is only on the first visit to this page).

spec/support/shared_examples/step_controller.rb
jayroh: If I remove this:

```
def ensure_application_present
  _application_not_created = true
end
```

... then you'll never be able to get into the first step. It'll always redirect you back.

app/controllers/residential_address_controller.rb
jessieay: yeah, unfortunately we haven't been testing all `skip`s consistently :/

app/controllers/introduce_yourself_controller.rb
jessieay: ahh ok -- so I think the var here was actually creating more confusion for me. Just return `true`?

jayroh: 👍

PR #121, opened 2017-08-22
Mask all money-related input fields

jayroh:
When we're asking for money-like answers we can now use a special form
input field called `mb_money_field` that will pass along the necessary
options to `mb_input_field` so that the jquery.maskMoney plugin will
format and mask accordingly.


jayroh:

jessieay:
%s/helper text/placeholder text

PR #123, opened 2017-08-22
Use headless chrome instead of poltergeist

jayroh:
When our JS-enabled tests run we should use (headless) chrome and
chromedriver as the driver. This removes the dependency of poltergeist
but adds the need for chromedriver. As a result:

1. Add the installation of chromedriver to our setup script
2. Add the installation of chromedriver to circle.yml

Because the platforms we test on install chrome(driver) in different
locations this commit adds a `FindChrome` class that will locate chrome
and chromedriver in either the most common locations, or a location we
can specify with the ENV variables `CHROME_PATH` or `CHROMEDRIVER_PATH`.
If they're not found - raise an exception.

Default JS driver is `:headless_chrome` but can be changed to `:chrome`
in `rails_helper.rb` if we want to watch those tests in a real chrome
browser.

Also:

* Update feature_helper to work with this driver. For some reason
capybara was complaining about the previous api usage - something about
how that wasn't implemented. ¯\_(ツ)_/¯

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,11 +44,11 @@ group :development, :test do
   gem "factory_girl_rails"
   gem "listen"
   gem "mailcatcher", "~> 0.2"
-  gem "poltergeist"
   gem "pry-rails"
   gem "rspec-rails"
   gem "rspec_junit_formatter"
   gem "rubocop-rspec", require: false
+  gem "selenium-webdriver"
   gem "simplecov"
 end
 
--- a/spec/support/chrome_headless.rb
+++ b/spec/support/chrome_headless.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "find_chrome"
+
+Capybara.register_driver :headless_chrome do |app|
+  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
+    chromeOptions: {
+      args: %w[headless disable-gpu],
+      binary: FindChrome.binary,
+    },
+  )
+
+  Capybara::Selenium::Driver.new(
+    app,
+    browser: :chrome,
+    desired_capabilities: capabilities,
+    driver_path: FindChrome.driver,
+  )
+end
--- a/spec/support/feature_helper.rb
+++ b/spec/support/feature_helper.rb
@@ -53,7 +53,7 @@ module FeatureHelper
 
   def js_select_radio(question:, answer_id:)
     within(find(:fieldset, text: question)) do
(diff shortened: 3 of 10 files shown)


Gemfile
```diff
@@ -44,11 +44,11 @@ group :development, :test do
   gem "factory_girl_rails"
   gem "listen"
   gem "mailcatcher", "~> 0.2"
-  gem "poltergeist"
   gem "pry-rails"
   gem "rspec-rails"
   gem "rspec_junit_formatter"
   gem "rubocop-rspec", require: false
+  gem "selenium-webdriver"
```
jessieay: do we want this outside of the dev/test group?

spec/support/chrome_headless.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require "find_chrome"
+
+Capybara.register_driver :headless_chrome do |app|
+  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
+    chromeOptions: {
+      args: %w[headless disable-gpu],
```
jessieay: seems like this is the only difference btwn this file and `chrome` -- probs not worth the mental overhead, but thoughts on sharing logic?

spec/support/feature_helper.rb
```diff
@@ -53,7 +53,7 @@ def select_radio(question:, answer:)
 
   def js_select_radio(question:, answer_id:)
     within(find(:fieldset, text: question)) do
-      find("##{answer_id}").trigger("click")
+      page.execute_script("$('##{answer_id}').trigger('click')")
```
jessieay: 💟

Gemfile
jayroh: We do. We're going to use the hell out of this for app driving! :)

spec/support/chrome_headless.rb
jayroh: Yeah I *could* probably clean this up a little bit. Think it's worth it?

spec/support/feature_helper.rb
jayroh: 🙌

Gemfile
jessieay: ok - move it!

spec/support/chrome_headless.rb
jessieay: probs not since we will likely not touch this code again...but maybe I am wrong

PR #125, opened 2017-08-23
Ask clients how they would like to be interviewed

adarsh:
Reason for Change
=================
* We should ask clients how they would like to be interviewed by the county.

Change
======
* Create a "Preferences" subsection
* Add a step for interview preferences.

Minor
=====
* Add a note in the `README.md` on skipping around the steps.
* Remove migration command related to paperclip. After removing paperclip, the migration using `add_attachment` no longer works. The suggestion [here][1] was to just rename the columns in the migration.

[1]: https://github.com/thoughtbot/paperclip/issues/1351#issuecomment-29522904

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,16 @@ Ask the team for access to the following services to begin contributing:
 
 ## What's Included
 
+### Step Skipping
+
+This application is a long questionaire.
+You will probably want to work on parts of it
+without completing the whole application.
+
+After booting the server and filling out the first step,
+go to `http://localhost:3000/steps`
+to jump around.
+
 ### Document / Photo Uploading Details
 
 * The file uploading service is called [Shubox]. The [dashboard] can be
--- a/app/controllers/preferences_interview_controller.rb
+++ b/app/controllers/preferences_interview_controller.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class PreferencesInterviewController < StandardStepsController
+end
--- a/app/views/preferences_interview/edit.html.erb
+++ b/app/views/preferences_interview/edit.html.erb
@@ -1,0 +1,24 @@
+<% content_for :header_title, "Interview Preferences" %>
+
+<div class="form-card">
+  <header class="form-card__header">
+    <div class="step-section-header">
+      <div class="step-section-header__subhead">
+        The next step after you apply is a brief interview with your county.
+      </div>
+    </div>
+  </header>
+
+  <div class="form-card__content">
+    <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
+      <%= f.mb_radio_set :interview_preference,
+        "What do you prefer?",
+        [
+          { value: :telephone, label: "Telephone Interview" },
+          { value: :in_person, label: "In-Person Interview" },
+        ]
+       %>
+
+    <%= render 'shared/next_step' %>
+  <% end %>
+</div>
--- a/spec/controllers/preferences_interview_controller_spec.rb
+++ b/spec/controllers/preferences_interview_controller_spec.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe PreferencesInterviewController do
+  let(:step) { assigns(:step) }
+  let(:invalid_params) { { step: { interview_preference: "telepathy" } } }
+  let(:step_class) { PreferencesInterview }
+
+  before { session[:snap_application_id] = current_app.id }
+
+  include_examples "step controller"
+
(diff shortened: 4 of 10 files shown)


adarsh:
[As an applicant, I want to be able to choose whether to conduct my required interview over the phone or in-person.](https://trello.com/c/06zqHr3P/133-as-an-applicant-i-want-to-be-able-to-choose-whether-to-conduct-my-required-interview-over-the-phone-or-in-person)

adarsh:
image
image

README.md
```diff
@@ -73,6 +73,16 @@ Ask the team for access to the following services to begin contributing:
 
 ## What's Included
 
+### Step Skipping
+
+This application is a long questionaire.
+You will probably want to work on parts of it
+without completing the whole application.
+
+After booting the server,
```
jayroh: `After booting the server and filling out the first step`

README.md
```diff
@@ -73,6 +73,16 @@ Ask the team for access to the following services to begin contributing:
...
+go to `http://localhost:3000/steps`
+to jump around.
```
jayroh: /gif house of pain

app/controllers/preferences_interview_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class PreferencesInterviewController < StandardStepsController
+  def self.step_class
+    PreferencesInterview
+  end
```
jayroh: I could be wrong but I think it'll infer this without needing to explicitly define it?

db/migrate/20170323225139_add_documents.rb
```diff
@@ -7,7 +7,10 @@ def up
       t.timestamps
     end
 
-    add_attachment :documents, :file
+    add_column :documents, :documents_file_name, :string
+    add_column :documents, :documents_content_type, :string
+    add_column :documents, :documents_file_size, :integer
+    add_column :documents, :documents_updated_at, :datetime
```
jayroh: I think you'll have to add the `documents` table again if we're going to break this out from the []String column in `snap_applications`. This is an old migration that ends up getting dropped in `db/migrate/20170728213954_drop_unused_database_tables.rb`

I'd recommend not doing *this* part until that can all be whipped into shape in one atomic commit.

app/controllers/preferences_interview_controller.rb
jessieay: yes! this works as long as the step does not end with an `s`

app/views/preferences_interview/edit.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      </div>
+    </div>
+  </header>
+
+  <div class="form-card__content">
+    <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
+      <%= f.mb_radio_set :interview_preference,
+        "What do you prefer?",
+        [
+          { value: "telephone", label: "Telephone Interview" },
```
jessieay: I feel like we've been making these `value` values symbols elsewhere. they get saved as strings in the DB so doesn't really matter either way, but would be more consistent to go symbol

spec/controllers/preferences_interview_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  let(:step) { assigns(:step) }
+  let(:invalid_params) { { step: { interview_preference: "telepathy" } } }
+  let(:step_class) { PreferencesInterview }
+
+  before { session[:snap_application_id] = current_app.id }
+
+  include_examples "step controller"
+
+  describe "#edit" do
+    it "assigns the attributes to the step" do
```
jessieay: this would be for a separate PR but I think it would be cool to get something like this into the shared example group

spec/controllers/preferences_interview_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  describe "#update" do
+    context "the preference is for a telephone interview" do
+      it "is valid" do
+        params = { interview_preference: "telephone" }
+
+        put :update, params: { step: params }
+
+        expect(response).to redirect_to("/steps/legal-agreement")
```
jessieay: also test that the preference is updated in the DB?

db/migrate/20170323225139_add_documents.rb
adarsh: Yeah I was conflicted about how to approach this. I'll revert the commit for now.

spec/controllers/preferences_interview_controller_spec.rb
adarsh: Yeah I agree!

PR #126, opened 2017-08-23
Continue driving MI Bridges

jayroh:
* Lock Selenium to the current version and monkey-patch   Selenium::WebDriver::Remote::Bridge#execute so that we may slow it   down a little bit (with `sleep`s) for demo purposes. 
* Remove explicit sleeps from within the page objects 
* Add some new steps - mostly intro and starter steps. 
* Create a `click_id` method that will execute javascript to find an   exact html element by id and send a `click()` to it. This helps with   some ambiguous checkbox and radio button labeling/naming 
* Common patterns can be put in the base page object - like `Page#next`
* Moving things into lib, namespaced under `MiBridges::Driver` class which
    does all of the work. This includes setting up capybara and tearing
    things down at the end of the driving session.


lib/mi_bridges/driver.rb
```diff
@@ -0,0 +1,64 @@
...
+    private
+
+    attr_reader :snap_application
+
+    def setup
+      Capybara.default_driver = ENV.fetch("WEB_DRIVER", "chrome").to_sym
+    end
+
+    def teardown
+      MiBridges::Driver::BasePage.new.close
```
jessieay: did this fix the session issue I was seeing?

lib/mi_bridges/driver.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      privacy_pin_page = PrivacyPinPage.new
+      privacy_pin_page.fill_in_required_fields
+      privacy_pin_page.next
+
+      BenefitsOverviewPage.new.next
+
+      FraudPenaltyAffidavit.new.next
+
+      start_page = Start.new
```
jessieay: why `Start` rather than `StartPage` ?

lib/mi_bridges/driver.rb
```diff
@@ -0,0 +1,64 @@
...
+      start_page.fill_in_required_fields
+      start_page.next
+
+      benefits_selector = BenefitsSelector.new
+      benefits_selector.fill_in_required_fields
+      benefits_selector.next
+      benefits_selector.next
```
jessieay: is this second `next` happening on the benefits selector page or is it a separate page? 

if separate, what do you think of creating a page class that represents that view? not worth it, perhaps. But also perhaps confusing to just call it on `benefits_selector` 🤔

lib/mi_bridges/driver/base_page.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module MiBridges
+  class Driver
+    class BasePage
```
jessieay: much better!

lib/mi_bridges/driver/benefits_selector.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module MiBridges
+  class Driver
+    class BenefitsSelector < BasePage
+      def fill_in_required_fields
+        click_id("#requestFoodShare")
+        click_id("#isAuthorized_N")
```
jessieay: what does this id represent? worth it to wrap it into a private method for explicitness?

lib/selenium/web_driver/remote.rb
```diff
@@ -0,0 +1,38 @@
...
+              path[key.inspect] = escaper.escape(value.to_s)
+            end
+          rescue IndexError
+            raise ArgumentError,
+              "#{opts.inspect} invalid for #{command.inspect}"
+          end
+
+          WebDriver.logger.info("-> #{verb.to_s.upcase} #{path}")
+
+          case ENV.fetch("DRIVER_SPEED", "fast").to_sym
```
jessieay: should we add something in the README about this? Have you tried it yet?

lib/selenium/web_driver/remote.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require "selenium-webdriver"
+
+module Selenium
+  module WebDriver
+    module Remote
+      class Bridge
```
jessieay: should this file be called `bridge.rb` since that is the class name?

jayroh:
You're the best, Jessie. Thanks for the comments. Will get these whipped into shape this morning :)

lib/mi_bridges/driver.rb
jayroh: You're right - it's a completely new page and should be represented with its own object. For sure.

lib/mi_bridges/driver.rb
jayroh: I do believe so! Everything looked good for me throughout the day. I'm going to do some work this morning to mitigate that issue, by the way.

lib/mi_bridges/driver/benefits_selector.rb
jayroh: Yeah, once I drive things again this morning I'll do that :)

lib/selenium/web_driver/remote.rb
jayroh: Yeah, I'll add this to the README now. Also, I did try it and it's A+ awesome. Slows things down *just* the right amount.

jayroh:
[User Agent for Web Driver](https://trello.com/c/WTr8JJL6/1-user-agent-for-web-driver)

PR #129, opened 2017-08-24
Validate length of SSN

jessieay:
**WHY**: So we can know that it is 9 digits when entering it,
digit-by-digit, into the 1171 PDF.

Also adds auto-strip-attributes so that if someone has an extra
whitespace when entering their SSN it is auto-stripped

This approach was inspired by the SSN formatting in GCF, which I had a
hand in.

### DIFF
--- a/spec/controllers/personal_detail_controller_spec.rb
+++ b/spec/controllers/personal_detail_controller_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe PersonalDetailController do
 
       expect(step.sex).to eq("male")
       expect(step.marital_status).to eq("Married")
-      expect(step.ssn).to eq("12345")
+      expect(step.ssn).to eq("123456789")
     end
   end
 
@@ -27,7 +27,7 @@ RSpec.describe PersonalDetailController do
         valid_params = {
           sex: "female",
           marital_status: "Divorced",
-          ssn: "54321",
+          ssn: "987654321",
         }
 
         put :update, params: { step: valid_params }
@@ -52,6 +52,6 @@ RSpec.describe PersonalDetailController do
   end
 
   def member
-    create(:member, sex: "male", marital_status: "Married", ssn: "12345")
+    create(:member, sex: "male", marital_status: "Married", ssn: "123456789")
   end
 end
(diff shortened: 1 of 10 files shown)


jessieay:

spec/support/shared_examples/social_security_number.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    it "does not allow dashes" do
+      subject.ssn = "  123-12-1234  "
+      expect(subject).to be_invalid
+      expect(subject.errors[:ssn]).to include(
+        "Make sure your SSN has 9 digits",
+      )
+    end
+  end
+end
```
jayroh: Does the auto_strip_attributes gem do the job of stripping out ALL non-integers before it's validated? 

If it doesn't do that - what do you think about masking the form input as a way to hold their hands in filling out the SSN? For many people it would ensure, in real-time, that they're entering it in the exact format we require.

jessieay: @jayroh no, it just strips whitespace. https://github.com/holli/auto_strip_attributes

the only thing I don't love about only allowing digits (I think that is what you are referring to when you say masking) is that it can sometimes feel unexpected. For example, our financial fields only allow numbers. Yesterday I was typing some words in there and nothing happened...so I thought my external keyboard was broken. Then I realized that numbers did work.

On the other hand, I can totally see why we'd want to help people not enter invalid info.

I think whether we want to do that is a UX question?

spec/controllers/personal_detail_controller_spec.rb
```diff
@@ -17,7 +17,7 @@
 
       expect(step.sex).to eq("male")
       expect(step.marital_status).to eq("Married")
-      expect(step.ssn).to eq("12345")
+      expect(step.ssn).to eq("123456789")
```
adarsh: I would also add a test for SSNs which start with a zero. I have been bitten by this with Boston-area zipcodes (e.g. 02215).

jessieay: done at 412a648

spec/support/shared_examples/social_security_number.rb
jayroh: Yeah, we can punt on this for now.

jayroh: As I was goign through the app on my phone, by the way, there were many many opportunities to make sure the phone KB was using the right keyboard set (numbers instead of letters, etc)

jessieay: created a card for that https://trello.com/c/8dWt7QB8

PR #130, opened 2017-08-24
Split shared examples

adarsh:
Reason for Change
=================
* Some steps don't have validation associated.
* Combining the validation shared example with the other ones then becomes brittle.

Changes
=======
* Split the shared examples into one for general steps and one for param validation.


adarsh:
> Only change I might make if I were you is to go through all the controller specs and see that they are using at least one of the shared examples. I think we've left it out of some because these example groups were not split.

Yeah this is a good idea but after starting, I realize this is an involved process which might slow me down today, so I'm going to make a GH issue for it instead to not forget.

PR #132, opened 2017-08-24
Include SSN digits individually in PDF

jessieay:
* For legibility!


jessieay:

jayroh:
Whoever was, is, reading these PDFs/faxes would probably hug you if they could.

🚢 it

PR #133, opened 2017-08-24
Allow clients to add additional information

adarsh:
Reason for Change
=================
* Allowing clients to add additional information about their current situation, _in their own words_ has proven valuable for food assistance applications.

Changes
=======
* Add a step in the Preferences section at the end, just before signatures.

Minor
-----
* Remove warning for hash key defined twice.

### DIFF
--- a/app/steps/step_navigation.rb
+++ b/app/steps/step_navigation.rb
@@ -33,8 +33,9 @@ class StepNavigation
       ExpensesAdditionalSourcesController,
       ExpensesAdditionalController,
     ],
-    "Preferences" => [
-      PreferencesInterviewController,
+    "General" => [
+      GeneralInterviewPreferenceController,
+      GeneralAnythingElseController,
     ],
     "Legal" => [
       LegalAgreementController,
--- a/db/migrate/20170824170659_add_anything_else_to_snap_application.rb
+++ b/db/migrate/20170824170659_add_anything_else_to_snap_application.rb
@@ -1,0 +1,5 @@
+class AddAnythingElseToSnapApplication < ActiveRecord::Migration[5.1]
+  def change
+    add_column :snap_applications, :additional_information, :text
+  end
+end
--- a/spec/features/snap_app_spec.rb
+++ b/spec/features/snap_app_spec.rb
@@ -145,8 +145,13 @@ accounts?",
     submit_expense_sources(answer: "yes")
     submit_expenses
 
-    on_page "Interview Preferences" do
+    on_page "General" do
       choose "Telephone Interview"
+      click_on "Continue"
+    end
+
+    on_page "General" do
+      fill_in "Anything else?", with: "This is helpful, thank you!"
       click_on "Continue"
     end
 
@@ -246,8 +251,12 @@ accounts?",
 
     submit_expense_sources(answer: "no")
 
-    on_page "Interview Preferences" do
+    on_page "General" do
       choose "Telephone Interview"
+      click_on "Continue"
+    end
+
+    on_page "General" do
       click_on "Continue"
     end
 
--- a/app/controllers/general_anything_else_controller.rb
+++ b/app/controllers/general_anything_else_controller.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class GeneralAnythingElseController < StandardStepsController
+end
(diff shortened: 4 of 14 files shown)


adarsh:
[As an applicant, I want to be able to share my story and situation in my own words.](https://trello.com/c/yClyAYp2/134-as-an-applicant-i-want-to-be-able-to-share-my-story-and-situation-in-my-own-words)

app/steps/step_navigation.rb
```diff
@@ -35,6 +35,7 @@ class StepNavigation
     ],
     "Preferences" => [
       PreferencesInterviewController,
+      PreferencesAnythingElseController,
```
jessieay: do we think this controller should be considered part of the preferences section? Feels kinda separate to me

app/views/preferences_anything_else/edit.html.erb
```diff
@@ -0,0 +1,20 @@
+<% content_for :header_title, "What else should we know?" %>
```
jessieay: Not sure if Alan suggested nav text. Usually we've been using the same nav text for each page in a section (this is not 100% consistent but it is the case probably 80% of the time)

app/views/preferences_anything_else/edit.html.erb
```diff
@@ -0,0 +1,20 @@
...
+
+      <div class="step-section-header__subhead">
+        Feel free to give other details you think might be helpful for us to know. Otherwise, you can skip this question.
+      </div>
+    </div>
+  </header>
+
+  <div class="form-card__content">
+    <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
+      <%= f.mb_textarea :anything_else, "Anything else?", options: { placeholder: "Write here" } %>
```
jessieay: I think there is an ongoing preference to not use placeholder text  https://trello.com/c/CstrMJ8l/85-design-as-a-client-i-want-the-most-usable-form-fields-possible-which-means-no-placeholder-text

app/views/preferences_anything_else/edit.html.erb
```diff
@@ -0,0 +1,20 @@
+<% content_for :header_title, "What else should we know?" %>
+
+<div class="form-card">
+  <header class="form-card__header">
+    <div class="step-section-header">
+      Is there anything else you'd like us to know about your situation?
+
+      <div class="step-section-header__subhead">
+        Feel free to give other details you think might be helpful for us to know. Otherwise, you can skip this question.
```
jessieay: break at 80 chars?

db/migrate/20170824170659_add_anything_else_to_snap_application.rb
```diff
@@ -0,0 +1,5 @@
+class AddAnythingElseToSnapApplication < ActiveRecord::Migration[5.1]
+  def change
+    add_column :snap_applications, :anything_else, :text
```
jessieay: on GCF they call this field `additional_information` -- thoughts on that name?

spec/features/snap_app_spec.rb
```diff
@@ -251,6 +256,11 @@
       click_on "Continue"
     end
 
+    on_page "What else should we know?" do
```
jessieay: maybe on this one we do not fill in to test at feature level that we do not need data in this field?

spec/models/snap_application_attributes_spec.rb
```diff
@@ -53,7 +53,6 @@
         financial_accounts_mutual_funds_or_stocks: false,
         financial_accounts_four_oh_one_k_or_iras: true,
         financial_accounts_trusts: false,
-        first_member_with_disability_name: nil,
```
jessieay: removed on purpose?

app/steps/step_navigation.rb
adarsh: Agreed. Renamed `Preferences` to `General` in a1b092f

app/views/preferences_anything_else/edit.html.erb
adarsh: Good idea! 9dc56f6

db/migrate/20170824170659_add_anything_else_to_snap_application.rb
adarsh: My thoughts are that it's a better name :) 2f30308

spec/models/snap_application_attributes_spec.rb
adarsh: Yes, this was a duplicate and was throwing a Ruby-level warning for the duplicated key. It looks like someone else fixed this after I rebased!

PR #134, opened 2017-08-24
Validate phone length / enter individually on PDF

jessieay:
* For legibility
* When doing this, I also discovered a bug with my SSN code (fixed here)
* Using `phony_rails` to format the phone number so we don't get any
spaces or dashes. Could also use a plain regex if people thing that is
preferred.


jessieay:

jessieay:
[As MDHHS, I only want to accept 10 digit phone numbers from clients](https://trello.com/c/uRkssE1h/86-as-mdhhs-i-only-want-to-accept-10-digit-phone-numbers-from-clients)

app/models/snap_application_attributes.rb
```diff
@@ -152,6 +151,18 @@ def residential_or_homeless
     end
   end
 
+  def phone_attributes
+    return {} if snap_application.phone_number.nil?
```
adarsh: Style: I could be misremembering but I believe we were aiming to avoid explicit returns and trailing conditionals, if we were hewing to the tb style guide. Not a huge deal.

Gemfile
```diff
@@ -15,6 +15,7 @@ gem "jquery-rails"
 gem "neat", "~> 1.8" # to keep in sync with getcalfresh
 gem "pdf-forms"
 gem "pg", "~> 0.18"
+gem "phony_rails"
```
adarsh: Solid gem name ☎️

app/steps/contact_information.rb
```diff
@@ -1,12 +1,16 @@
...
+  attr_accessor :phone_number, :phone_number_as_normalized
+  phony_normalize :phone_number, default_country_code: "US"
+
   step_attributes(
     :email,
     :phone_number,
     :sms_subscribed,
   )
 
+  validates_plausible_phone :phone_number, with: /\A\+\d+/
```
adarsh: 😁

app/views/contact_information/edit.html.erb
```diff
@@ -8,7 +8,7 @@
   </header>
   <div class="form-card__content">
     <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
-      <%= f.mb_input_field :phone_number, "What is the best phone number to reach you?", options:  { placeholder: "Phone #" } %>
+      <%= f.mb_input_field :phone_number, "What is the best phone number to reach you?" %>
```
adarsh: Copy: "What is the best phone number _at which_ to reach you?" maybe?

app/models/snap_application_attributes.rb
jessieay: I dig it -- updated at b6ac8b2

app/views/contact_information/edit.html.erb
jessieay: I think that's a good suggestion but I will let this one go for now since we will do a copy audit soon anyway. I know simplicity of language is a big goal.

PR #135, opened 2017-08-24
Dockerize

mgwalker:
Reason for Change
===================
Puts the Rails app into a docker container, to simplify the process of getting up and running as a developer.  Also eliminates the need for runtime version management (e.g., `rbenv`) since everything is tucked inside the container.

Changes
=======
* Add `Dockerfile` to build an image first based on the [pdfktk image](https://hub.docker.com/r/mnuessler/pdftk/).  Then layers from the [Ruby Dockerfile](https://github.com/docker-library/ruby/blob/master/2.4/jessie/Dockerfile) with some additions necessary for this project.  I also removed the `buildDeps` cleanup step because some of those dependencies are still needed downstream from image build.
* Add `docker-compose.yml` that setups a PostgreSQL container and links it to the Rails app container.  Sets up development runtime environment variables and setups port forwarding.
* Modify database configuration to pull database host, username, and password from environment variables (see [The Twelve-Factor App](https://12factor.net/), but also this is how that information gets pushed into the Rails app container).
* Modify `Procfile.dev` to have Rails server and mailcatcher bind to all interfaces (`0.0.0.0`) instead of just whatever they default to (`localhost`?), so the port forwarding will work
* Add phantomjs and poltergeist to the Gemfile so `rake` tests run.
* Updated `README.md` to include instructions for running with Docker.

Minor
=====
* N/A

Risks
=====
* The `delayed_job` didn't run (it complained that `strftime` was not a method on a nil object; rooting around the code, this looks related to the applicant's birthday).  To be honest, I'm not familiar enough with Rails to get a better sense of what's going on, so I don't know if this is a known issue or something introduced by being in Docker.  But I figure opening this PR lets the folks at CfA who actually know the code dig into the issue a bit if they want, rather than me spinning my wheels.
* As a result of the previous bullet, I also couldn't verify that the PDF was actually created or that the email was sent.


adarsh:
Hello!

Thanks for your interest in this project, however we are not currently soliciting external contributions, as described in [CONTRIBUTING.md][1]:

> While this project is open source, we are not currently seeking external contributions. If you would like to learn more about this project, shoot us an email at alanw@codeforamerica.org.



[1]: https://github.com/codeforamerica/michigan-benefits/blob/cf7ee9f7c839c9f72b02dae111a652f9870b2279/CONTRIBUTING.md

alanjosephwilliams:
Hello @mgwalker! Great to have 18F in the building. Thanks for putting containerization on the radar. As @adarsh mentioned, while we're building in the open, we're currently in a Discovery Phase that has a narrow roadmap. We will circle back with the broader community in late Fall when that phase concludes.

PR #136, opened 2017-08-25
PDF tweaks

jessieay:
* Based on going through and doing detailed QA
* Removed money inputs for floats for now until we fix https://trello.com/c/dbpMAUh0/156-bug-money-input-not-working-for-float-fields


app/models/snap_application_attributes.rb
```diff
@@ -207,8 +207,7 @@ def members_not_buy_food_with
       members.
       where.
       not(buy_food_with: true).
-      select('first_name || " " || last_name AS full_name').
-      pluck(:first_name, :last_name).
+      pluck('concat(first_name, \' \', last_name) AS full_name').
```
jayroh: ✨ ✨ ✨ ✨ ✨ ✨

app/models/snap_application_attributes.rb
```diff
@@ -221,7 +220,7 @@ def financial_accounts(*types)
       end
     end
 
-    included.include?(true)
+    bool_to_checkbox(included.include?(true))
```
jayroh: `included.include?(true)`

Could this be pulled out to something that describes it a little better? 😸

app/views/income_additional/edit.html.erb
```diff
@@ -13,7 +13,7 @@
     <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
 
       <% @income_fields.each do |method, label| %>
-        <%= f.mb_money_field method, label, notes: ["Monthly amount"] %>
+        <%= f.mb_input_field method, label, notes: ["Monthly amount"] %>
```
jayroh: I'll take a look at how GCF does this and will see about using that approach instead of the js lib I previously included.

app/models/snap_application_attributes.rb
jessieay: done at 5fcad41 ;) -- agree that the code situation was funky before 💩

app/views/income_additional/edit.html.erb
jessieay: cool! ✔️

PR #137, opened 2017-08-25
Validate 10 digit phone number

jessieay:
* Moves away from phony rails, which was not working 100% of the time
* This is better for PDF filling because we want 10 digits for the PDF
* In future, might be good to allow spaces or dashes for readability.


jessieay:

spec/controllers/contact_information_controller_spec.rb
```diff
@@ -24,7 +24,7 @@
     context "when valid" do
       it "redirects to the next step" do
         valid_params = {
-          phone_number: "222-222-2222",
+          phone_number: "2222222222",
```
adarsh: Relevant: https://youtu.be/pxEu4uf0-b8?t=75

PR #138, opened 2017-08-25
Add default for "buy food with" attr on members

jessieay:
* Default to true


jessieay:
[Show correct names for "people I do not buy food with"](https://trello.com/c/UPsAfZKS/170-show-correct-names-for-people-i-do-not-buy-food-with)

PR #139, opened 2017-08-25
PDF tweaks based on QA

jessieay:
Section A:
- add data in checkboxes for homeless and not homeless

Section K / L:
* Make larger boxes "multiline" so text wraps

Section M:
* Fix naming of "pay interval other detail" so that the type of pay
interval is filled in


jessieay:
[PDF tweaks](https://trello.com/c/3bkRQvZ1/171-pdf-tweaks)

spec/models/employed_member_attributes_spec.rb
```diff
@@ -25,11 +25,11 @@
         second_employed_full_name: "Foo Flace",
         second_employed_employer_name: "McTacoBacon R Us",
```
adarsh: Yum!

PR #140, opened 2017-08-25
Prepend money fields with "$"

jessieay:
* Move away from mask money js, which was causing a weird bug with
floats and saving values as zero.

screen shot 2017-08-25 at 1 51 05 pm


jessieay:

jessieay:
[Bug: money input not working for float fields](https://trello.com/c/dbpMAUh0/156-bug-money-input-not-working-for-float-fields)

app/helpers/mb_form_builder.rb
```diff
@@ -34,25 +34,20 @@ def mb_input_field(
   def mb_money_field(
     method,
     label_text,
-    type: "text",
+    type: :tel,
```
adarsh: Should this be `:tel`? Does that mean Telephone?

app/helpers/mb_form_builder.rb
```diff
@@ -34,25 +34,20 @@ def mb_input_field(
...
     notes: [],
     options: {},
-    classes: [],
-    prefix: nil,
+    classes: ["form-width--med"],
+    prefix: "$",
```
adarsh: Nice. This is super clear.

app/helpers/mb_form_builder.rb
jessieay: yeah it's so that on mobile the number keyboard is pulled up!

PR #141, opened 2017-08-26
Frontend audit with Alan

jessieay:
* Update HTML to use classes that exist in our CSS
* Override / add classes where necessary
* Other small UI tweaks

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -14,3 +14,52 @@
 //= require jquery_ujs
 //= require handlebars
 //= require_tree .
+
+var radioSelector = (function() {
+  var rs = {
+    init: function() {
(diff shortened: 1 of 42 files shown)


jessieay:
@adarsh I dont want to fix these CC issues just now, related to some jquery I copied over from GCF

jessieay:
[Make "beta" banner less severe, using "teal"](https://trello.com/c/j5IvTHcj/173-make-beta-banner-less-severe-using-teal)

zspencer:
I don't appear to have the rights to dismiss the codeclimate red x thing; but the concern it has doesn't really seem to be worth blocking this pull request.

jessieay:
@zspencer thanks for taking a look!

making these changes made me realize that we should perhaps be using shared partials for this stuff, but all in due time!

PR #143, opened 2017-08-28
Send system emails from `hello@`

jessieay:
screen shot 2017-08-28 at 9 49 27 am

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
-  default from: "admin@#{ENV['EMAIL_DOMAIN']}"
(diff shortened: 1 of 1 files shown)


jessieay:
[send emails to users from hello@michiganbenefits.org](https://trello.com/c/kkW9ei5I/166-send-emails-to-users-from-hellomichiganbenefitsorg)

jessieay:
@zspencer I think that's a good suggestion! 

I believe we already have the `EMAIL_DOMAIN` env var set up because we want to make sure it maps with what Mailgun is expecting in terms of sending domain. We cannot send emails from other domains because a domain must be "verified" by mailgun in order to send emails from that domain: https://help.mailgun.com/hc/en-us/articles/202052074-How-do-I-verify-my-domain-

So I guess one benefit of this env var is that we know its value should not change, since sending from `foo.org` would not work. But, it does seem kind of weird. I will think on this....

PR #144, opened 2017-08-28
Fix formatting of money input fields

jessieay:
* Move from floats to integers
* Make inputs full width

screen shot 2017-08-28 at 9 54 17 am


jessieay:

PR #145, opened 2017-08-28
Fax particular offices based upon which zip code the applicant resides in

zspencer:
This will resolve https://trello.com/c/fDsq8kig/154-when-a-client-submits-an-application-it-goes-to-the-correct-mdhhs-office-based-on-their-zip-code

Changes
=======
- [x] Remove the FAX_NUMBER environment variable
- [x] ~~~Maybe? Remove the FAX_RECIPIENT environment variable?~~~
- [ ] Write a feature spec that demonstrates differentiating fax recipient based upon zip code
- [x] Write unit-level permutation specs that:
   - [x] Exercise determining the fax number based upon which zip codes an office serves and the application is from
   - [x] Excercise what happens when the zip code is not found in any offices

Risks
=====
- [x] Did I miss zip codes for the mapping?
- [x] Do we have the appropriate fax numbers for the offices for production?
- [ ] Did we set the APP_RELEASE_STAGE variable appropriately for each application in the pipeline?


zspencer:
[When a client submits an application, it goes to the correct MDHHS office based on their zip code](https://trello.com/c/fDsq8kig/154-when-a-client-submits-an-application-it-goes-to-the-correct-mdhhs-office-based-on-their-zip-code)

zspencer:
@jessieay - How do you feel about this? 

In particular I'm thinking about:
 - Maybe separating out the concern for the configuration 
  - Renaming FaxRecipient to Office?
  - Renaming the FaxRecipient::Office class to Finders or something?

I'm also kind of hesitant to keep each environment in the same YML file, but I don't think it's *that* bad yet.

zspencer:
(Also, thought I had during lunch, it probably wouldn't be *too hard* to migrate this to something that is database backed? Is that worth doing if it's doable by end of day?)

app/jobs/send_application_job.rb
```diff
@@ -30,22 +31,19 @@ def send_pdf
...
-  def fax_number
-    ENV.fetch("FAX_NUMBER", "")
-  end
-
-  def fax_recipient_name
-    ENV.fetch("FAX_RECIPIENT", "Michigan Bridges")
+  def fax_recipient
+    FaxRecipient.new(residential_address: residential_address)
   end
+  delegate :number, :name, to: :fax_recipient, prefix: true
```
jessieay: this is not highly important to me, but I generally prefer explicit methods rather than `delegate` because when looking for the logic behind something I usually search for the method name, eg `def fax_recipient_number` and if it is defined with`delegate` I will not find it.

app/models/fax_recipient.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def number
+    Office.find_by(zip: residential_address.zip)["fax_number"]
+  end
+
+  def name
+    ENV.fetch("FAX_RECIPIENT", "Michigan Bridges")
+  end
+
+  class Office
```
jessieay: Generally prefer one class per file -- thoughts on putting this baby class in its own file?

config/offices.yml
```diff
@@ -0,0 +1,27 @@
...
+  # Which office to send the fax to in the event the submitted
+  # zip code is not in our mapping
+  fallback_office: "union"
+
+  # Because YML inheritance doesn't work well with nested hashes,
+  # We are relying on matching the list of offices per-environment with
+  # the list of zip codes a given office serves to reduce configuration duplication.
+  # This may be unnecessary?
+  covered_zip_codes:
+    clio: [48415,48417,48420,48421,48433,48437,48457,48458,48460,48463,48464,48473,48504,48506,48531,48532,48746]
```
jessieay: I think that our rubocop config pefers spaces after commas within arrays. Or is not doing that a yml thing?

config/offices.yml
```diff
@@ -0,0 +1,27 @@
+# APP_RELEASE_STAGE
+defaults: &defaults
+  # Which office to send the fax to in the event the submitted
+  # zip code is not in our mapping
+  fallback_office: "union"
+
+  # Because YML inheritance doesn't work well with nested hashes,
+  # We are relying on matching the list of offices per-environment with
+  # the list of zip codes a given office serves to reduce configuration duplication.
+  # This may be unnecessary?
```
jessieay: Makes sense -- don't think we *need* this comment as long as we have good tests for what's going on in here :)

spec/models/fax_recipient_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
+
+RSpec.describe FaxRecipient do
+  subject(:instance) do
```
jessieay: generally prefer to not use `subject` or `let` in RSpec tests if that's OK with you. We have it in some places so they can be used in shared examples but I find these aspects of the RSpec DSL to make tests harder to read and update.

config/offices.yml
```diff
@@ -0,0 +1,27 @@
...
+  covered_zip_codes:
+    clio: [48415,48417,48420,48421,48433,48437,48457,48458,48460,48463,48464,48473,48504,48506,48531,48532,48746]
+    union: [48411,48423,48429,48430,48436,48438,48439,48442,48449,48451,48501,48502,48503,48505,48507,48509,48519,48529]
```
jessieay: might be helpful to have each element of these arrays on its own line for readability

config/offices.yml
zspencer: Oooh good point. I was doing regexs to transform this and lost the space.

PR #146, opened 2017-08-28
Append to 1171 PDF if more than 6 household members

jessieay:
**WHY**: Form only has place for 6 members in "Section C", but we allow
people to add unlimited household members in the web version of the app.


jessieay:

jessieay:
@zspencer not sure if there is a good way to open up the generated PDF and look at its contents. Do you know of one? I tested this manually to see that the info looked OK but would be good to get something in RSpec

zspencer:
There are ruby PDF parsers, but they are quite finicky. While I think it would be valuable to start adding them to our test suite, I would prefer we had a spike card to poke at it instead of making it a requirement for this feature.

PR #147, opened 2017-08-29
Configure Sentry to capture unhandled errors

jessieay:
* Per setup rec from Heroku
* See https://devcenter.heroku.com/articles/sentry
=


jessieay:
[Add monitoring for staging and prod so we know when there are errors](https://trello.com/c/9YheOula/119-add-monitoring-for-staging-and-prod-so-we-know-when-there-are-errors)

PR #150, opened 2017-08-29
Add 3 logo footer to PDF

jessieay:
* This adds the footer to each page of the PDF, no matter the length
* Using a bounding box for text so that the footer image and text do not
overlap

[example.pdf](https://github.com/codeforamerica/michigan-benefits/files/1261086/example.pdf)


jessieay:

jessieay:
@zspencer when the members expand past the bounding box, it goes to a second page. If you check out the linked example in the PR description, you will see that the footer is on both pages and that the member details go onto both pages and are not obscured. 🎉

zspencer:
Oh gotcha. I should have checked the output instead of just reading the code :D

PR #151, opened 2017-08-29
Fax signed applications after 30 minutes of inactivity

zspencer:
Reason for Change
===================
Previously, we had tied together the notion of faxing the office an application with emailing the applicant. This changes it so that signed applications are automatically faxed after 30 minutes.

Changes
=======
* Split out SendApplicationJob into EmailApplicationJob and FaxApplicationJob
* Add to `SnapApplication` `#faxed_at`, `#faxed?`, and `.faxable` methods
* Write rake task to enqueue faxes `:queue_faxes`

Minor
=====
* N/A?

Risks
=====
* If we do not add a worker with a schedule, we will not send faxes
* If the queues get super backed up it's possible we could queue multiple faxes for the same application; and if both of those execute at the same time we may double fax.

### DIFF
--- a/app/jobs/fax_application_job.rb
+++ b/app/jobs/fax_application_job.rb
@@ -1,10 +1,10 @@
-class SendApplicationJob < ApplicationJob
-  def perform(snap_application:)
-    @snap_application = snap_application
-
+class FaxApplicationJob < ApplicationJob
+  def perform(snap_application_id:)
+    @snap_application = SnapApplication.find(snap_application_id)
+    return if snap_application.faxed?
     create_pdf
-    send_pdf
     fax_pdf
+    snap_application.update(faxed_at: Time.zone.now)
   ensure
     complete_form_with_cover.try(:close)
     complete_form_with_cover.try(:unlink)
@@ -19,13 +19,6 @@ class SendApplicationJob < ApplicationJob
     @complete_form_with_cover = Dhs1171Pdf.new(
       snap_application: snap_application,
     ).completed_file
-  end
-
-  def send_pdf
-    ApplicationMailer.snap_application_notification(
-      file_name: complete_form_with_cover.path,
-      recipient_email: snap_application.email,
-    ).deliver
   end
 
   def fax_pdf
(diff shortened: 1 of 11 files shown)


zspencer:

app/jobs/fax_application_job.rb
```diff
@@ -1,10 +1,10 @@
-class SendApplicationJob < ApplicationJob
-  def perform(snap_application:)
-    @snap_application = snap_application
-
+class FaxApplicationJob < ApplicationJob
+  def perform(snap_application_id:)
```
jessieay: it seems inconsistent to pass in the `id` here and the object in the other job. Thoughts on doing one or the other for both?

spec/models/snap_application_spec.rb
```diff
@@ -62,4 +62,52 @@
       end
     end
   end
+
+  describe ".enqueue_faxes" do
+    it "schedules jobs for signed, unfaxed application updated awhile ago" do
+      signed_unfaxed_updated_awhile_ago = create(:snap_application,
```
jessieay: it might be easier to read if we put each of these vars into a private method. Just to make the spec easier to read. No biggie either way

app/jobs/fax_application_job.rb
zspencer: Yea! I was thinking about making both use IDs, but wanted to decrease surface area in the short term. 

I don't trust delayed jobs serialization/deserialization to gracefully handle if/when we enqueue multiple fax jobs for the same snap application.

spec/models/snap_application_spec.rb
zspencer: Good point! I normally use `let!` for that, since that's basically what `let!` does.

jessieay: yes for sure! Joel and I have a weird allergic reaction to `let` that might not entirely be fair but is very real. I always go back to this cheeky blog post when discussing it https://robots.thoughtbot.com/lets-not

So for now I'd love to do private methods instead of `let`, but inline is also 💯

PR #152, opened 2017-08-29
Replace hello@michiganbenefits.org

adarsh:
Reason for Change
=================
* We are launching shortly with a non-functioning customer service email address (hello@michiganbenefits.org).
* For now, sending email to Alan is fine.

Changes
=======
* Replace the default "from" email address.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
-  default from: "hello@#{ENV['EMAIL_DOMAIN']}"
+  default from: "hello@#{ENV['EMAIL_DOMAIN']}",
+          reply_to: "alanw@codeforamerica.org"
+
   layout "mailer"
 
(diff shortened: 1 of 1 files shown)


adarsh:
NOTE: Do not merge this PR. We will only merge it if necessary prior to launch.

app/mailers/application_mailer.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
-  default from: "hello@#{ENV['EMAIL_DOMAIN']}"
+  default from: "alanw@codeforamerica.org"
```
jessieay: I am not sure this will work since `codeforamerica.org` is not setup on mailgun as a verified domain: http://mailgun-documentation.readthedocs.io/en/latest/user_manual.html#verifying-your-domain

But we could set his email address as the `reply_to`, I think?

adarsh: Oh even better. Fixed in bf58da2.

adarsh:
image

I have performed acceptance testing on staging.

PR #153, opened 2017-08-29
Append pages when greater than 2 employed or self employed

jessieay:
**WHY**: there is only enough room on the form for 2 of each

Minor: this also adds checkboxes to "Section C" so that whoever is
reading the app knows when there are additional pages. We already have
these checkboxes for the employment sections.

Sample with a bunch of appended pages:

[appendages all over the place pdf.pdf](https://github.com/codeforamerica/michigan-benefits/files/1261881/appendages.all.over.the.place.pdf.pdf)


jessieay:

PR #158, opened 2017-08-29
Load the environment when using Rails models in rake tasks

zspencer:
This is what happens when you test via the command line and not the actual rake task.


zspencer:

adarsh:
doh

PR #159, opened 2017-08-30
Refactor pass for fax changes

zspencer:
@jessieay made the good point that we want our export jobs to be consistent; so I took some time to:

1. Pull up the shared behavior of PDF generation into the SnapApplication (and test it there)
2. Simplify the tests to focus on the behavior of the export
3. Simplify the Jobs themselves now that the SnapApplication owns the PDF itself.

### DIFF
--- a/app/models/snap_application.rb
+++ b/app/models/snap_application.rb
@@ -14,6 +14,12 @@ class SnapApplication < ApplicationRecord
     end
   end
 
+  def pdf
+    @pdf ||= Dhs1171Pdf.new(
+      snap_application: self,
+    ).completed_file
+  end
+
   def monthly_gross_income
     [
       monthly_additional_income,
--- a/spec/jobs/fax_application_job_spec.rb
+++ b/spec/jobs/fax_application_job_spec.rb
@@ -2,70 +2,37 @@ require "rails_helper"
 
 RSpec.describe FaxApplicationJob do
   describe "#perform" do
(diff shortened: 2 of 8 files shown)


zspencer:

app/models/snap_application.rb
```diff
@@ -14,6 +14,12 @@ def self.enqueue_faxes
     end
   end
 
+  def pdf
+    @pdf ||= Dhs1171Pdf.new(
```
jessieay: nice! would this memoization cause problems if we were to update the snap app and want the PDF updated as well?

spec/jobs/fax_application_job_spec.rb
```diff
@@ -2,70 +2,37 @@
 
 RSpec.describe FaxApplicationJob do
   describe "#perform" do
-    it "creates a PDF with the snap application data" do
+    it "sends a fax when the application hasn't been faxed already" do
       snap_application = create(:snap_application, :with_member)
-      tempfile = Tempfile.new("send_application_job_spec")
```
jessieay: nice spec cleanup!

app/models/snap_application.rb
zspencer: It could, but it is unlikely as every request/response cycle retrieves a snap application from the database. (This may not be as true with action cable tho?). We could also do a `dirty?` check on the applicant information and force regeneration when the application is dirty.

PR #160, opened 2017-08-30
Update style of success page

jessieay:
- Iconography/Illustration at the top
- "Blue Text" that says "Congratulations"
- Change the headline text
- Include subheader text that gives the client the date submitted.
- Include the specification of which program was submitted, using the icons that represent the FAP program (found on the homepage as well).

screen shot 2017-08-30 at 10 45 20 am

### DIFF
--- a/app/steps/success.rb
+++ b/app/steps/success.rb
@@ -1,14 +1,14 @@
 # frozen_string_literal: true
 
 class Success < Step
-  step_attributes :email, :faxed_at
+  step_attributes :email, :signed_at
 
   validates :email,
     presence: { message: "Make sure to provide an email address" }
 
   def submitted_date
-    if faxed_at.present?
-      faxed_at.strftime("%m-%d-%Y")
+    if signed_at.present?
+      signed_at.strftime("%m-%d-%Y")
     end
   end
 end
--- a/app/steps/success.rb
+++ b/app/steps/success.rb
@@ -1,8 +1,14 @@
 # frozen_string_literal: true
 
 class Success < Step
-  step_attributes :email
+  step_attributes :email, :faxed_at
 
   validates :email,
     presence: { message: "Make sure to provide an email address" }
+
+  def submitted_date
+    if faxed_at.present?
+      faxed_at.strftime("%m-%d-%Y")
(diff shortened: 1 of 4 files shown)


jessieay:
[Include headline, icon, and information on "Confirmation Page"](https://trello.com/c/SnMGleBb/138-include-headline-icon-and-information-on-confirmation-page)

app/steps/success.rb
```diff
@@ -1,8 +1,14 @@
...
 class Success < Step
-  step_attributes :email
+  step_attributes :email, :faxed_at
 
   validates :email,
     presence: { message: "Make sure to provide an email address" }
+
+  def submitted_date
+    if faxed_at.present?
+      faxed_at.strftime("%m-%d-%Y")
```
zspencer: I think we may want signed_at here? faxed_at will be ~30m from signature

app/views/success/edit.html.erb
```diff
@@ -3,9 +3,22 @@
...
   <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
     <header class="form-card__header">
+      <div class="illustration illustration--success"></div>
+      <div class="pre-title">Congratulations</div>
       <div class="form-card__title">
-        Your Michigan Benefits application is ready to be submitted!
+        Your application has been successfully submitted.
       </div>
+      <div class="text--help text--centered">
+        Submitted to MDHHS on <%= @step.submitted_date %>.
```
zspencer: when this is nil, it will say `Submitted to MDHHS on .` is that OK?

jessieay: In practice someone should never see this page until they've signed. I added the nil check so that us developers don't get an error when skipping steps :)

app/steps/success.rb
jessieay: Updated at 2240865

This does bring up the issue of not faxing right away. If someone is on the page, they are *done* and we can safely fax their app rather than waiting 30 mins. Thoughts on a separate PR for faxing right away when someone finishes doc upload or indicates that they do not want to do doc upload?

PR #161, opened 2017-08-30
Use numeric keyboard on mobile for number fields

jessieay:
**WHY**: easier to use!

Note that money fields already use this because we set `type: tel` in
the money input helper method (I confirmed that this worked on mobile)

Minor: extracted shared partial for addresses since I noticed that our
two address forms were already diverging in some small ways.


jessieay:
[Make sure form inputs select correct keyboard on mobile](https://trello.com/c/8dWt7QB8/153-make-sure-form-inputs-select-correct-keyboard-on-mobile)

PR #162, opened 2017-08-31
First pass at putting together the shape of a metrics dashboard

zspencer:
**Why?** Administration is going to be more important as we go from 0 snap applications a day to > 9000 (OK OK, 10)

### TODO
- [x] Actually make the resend fax button work
- [x] Make the list of fields on the index reflect the trello card 
- [x] Start recording when we execute the emailed application job
- [x] Maybe? Make a "ExportAttempt" model for recording which kind of export and meta data about the export and if it succeeded/failed so we have a better trail of what is going on.
screen shot 2017-09-07 at 3 14 33 pm

### DIFF
--- a/app/assets/stylesheets/michigan-benefits/administrate-extensions.scss
+++ b/app/assets/stylesheets/michigan-benefits/administrate-extensions.scss
@@ -1,0 +1,16 @@
+// Extends administrate CSS per https://github.com/thoughtbot/administrate/pull/436/files
+
+.main-content__funnel {
+  display: flex;
+  flex: 1 1 100%;
+  flex-wrap: wrap;
+
+  padding: 1rem 2rem;
+  h3 {
+    flex: 1 1 100%;
+  }
+}
+
+.main-content__funnel__stage {
+  flex: 1 4 25%;
+}
--- a/app/controllers/admin/snap_applications_controller.rb
+++ b/app/controllers/admin/snap_applications_controller.rb
@@ -17,5 +17,25 @@ module Admin
 
     # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
     # for more information
+    #
+    def resend_fax
+      application = SnapApplication.find(params[:id])
+      application.exports.faxed.succeeded.update_all(
+        status: :failed,
+        metadata: "Reset by admin",
+      )
+      Export.create_and_enqueue!(snap_application: application, destination:
+                                 :fax)
+      flash[:notice] = "Resent fax #{application.email}!"
+      redirect_to admin_root_path
+    end
+
+    def pdf
+      application = SnapApplication.find(params[:id])
+      send_data(application.pdf.read,
+        filename: "snap-application-#{application.email}.pdf",
+        type: "application/pdf",
+        disposition: "inline")
+    end
   end
 end
--- a/app/controllers/admin/snap_applications_controller.rb
+++ b/app/controllers/admin/snap_applications_controller.rb
@@ -29,7 +29,7 @@ module Admin
     def pdf
       application = SnapApplication.find(params[:id])
       send_data(application.pdf.read,
-        filename: "snap-application-#{application.email}.pdf",
+        filename: "snap-application-#{application.signature}.pdf",
         type: "application/pdf",
         disposition: "inline")
     end
--- a/app/controllers/admin/snap_applications_controller.rb
+++ b/app/controllers/admin/snap_applications_controller.rb
@@ -20,13 +20,9 @@ module Admin
     #
     def resend_fax
       application = SnapApplication.find(params[:id])
-      application.exports.faxed.succeeded.update_all(
-        status: :failed,
-        metadata: "Reset by admin",
-      )
-      Export.create_and_enqueue!(snap_application: application, destination:
-                                 :fax)
-      flash[:notice] = "Resent fax #{application.email}!"
+      Export.create_and_enqueue!(snap_application: application,
+                                 destination: :fax, force: true)
+      flash[:notice] = "Resent fax for #{application.signature}!"
       redirect_to admin_root_path
     end
 
--- a/app/controllers/success_controller.rb
+++ b/app/controllers/success_controller.rb
@@ -12,14 +12,14 @@ class SuccessController < StandardStepsController
   private
 
   def before_rendering_edit_hook
-    FaxApplicationJob.
-      perform_later(snap_application_id: current_snap_application.id)
+    Export.create_and_enqueue(destination: :fax,
+                              snap_application: current_snap_application)
   end
 
   def after_successful_update_hook
     flash[:notice] = flash_notice
-    EmailApplicationJob.
-      perform_later(snap_application_id: current_snap_application.id)
+    Export.create_and_enqueue(destination: :email,
+                              snap_application: current_snap_application)
   end
 
   def flash_notice
--- a/app/dashboards/snap_application_dashboard.rb
+++ b/app/dashboards/snap_application_dashboard.rb
@@ -1,68 +1,18 @@
 require "administrate/base_dashboard"
 
 class SnapApplicationDashboard < Administrate::BaseDashboard
-  ATTRIBUTE_TYPES = {
-    id: Field::Number,
-    created_at: Field::DateTime,
-    updated_at: Field::DateTime,
-    signature: Field::String,
-    signed_at: Field::DateTime,
-    email: Field::String,
-    documents: Field::String,
-    phone_number: Field::String,
-    sms_subscribed: Field::Boolean,
-    consent_to_terms: Field::Boolean,
-    mailing_address_same_as_residential_address: Field::Boolean,
-    unstable_housing: Field::Boolean,
-    everyone_a_citizen: Field::Boolean,
-    anyone_disabled: Field::Boolean,
-    anyone_new_mom: Field::Boolean,
-    anyone_in_college: Field::Boolean,
-    anyone_living_elsewhere: Field::Boolean,
-    income_change: Field::Boolean,
-    income_change_explanation: Field::Text,
-    additional_income: Field::Text,
-    income_child_support: Field::String,
-    income_foster_care: Field::String,
-    income_other: Field::String,
-    income_pension: Field::String,
-    income_social_security: Field::String,
-    income_ssi_or_disability: Field::String,
-    income_unemployment_insurance: Field::String,
-    income_workers_compensation: Field::String,
-    rent_expense: Field::Number,
-    property_tax_expense: Field::Number,
-    insurance_expense: Field::Number,
-    utility_heat: Field::Boolean,
-    sms_consented: Field::Boolean,
-    utility_cooling: Field::Boolean,
-    utility_electrity: Field::Boolean,
-    utility_water_sewer: Field::Boolean,
-    utility_trash: Field::Boolean,
-    utility_phone: Field::Boolean,
-    utility_other: Field::Boolean,
-    dependent_care: Field::Boolean,
-    medical: Field::Boolean,
-    court_ordered: Field::Boolean,
-    monthly_care_expenses: Field::Number,
-    monthly_medical_expenses: Field::Number,
-    monthly_court_ordered_expenses: Field::Number,
-    care_expenses: Field::String,
-    medical_expenses: Field::String,
-    court_ordered_expenses: Field::String,
-    money_or_accounts_income: Field::Boolean,
-    real_estate_income: Field::Boolean,
-    vehicle_income: Field::Boolean,
-    financial_accounts: Field::String,
-    total_money: Field::Number,
-  }.freeze
-
   COLLECTION_ATTRIBUTES = %i[
     id
+    signature
     sms_consented
-    email
     phone_number
+    email
+    zip
+    signed_at
+    faxed_at
+    fax_metadata
     created_at
+    emailed_at
   ].freeze
 
   SHOW_PAGE_ATTRIBUTES = %i[
@@ -171,6 +121,9 @@ class SnapApplicationDashboard < Administrate::BaseDashboard
     financial_accounts
     total_money
   ].freeze
+
+  ATTRIBUTE_TYPES = (FORM_ATTRIBUTES + SHOW_ATTRIBUTES + COLLECTION_ATTRIBUTES).
+    uniq.freeze
 
   def display_resource(snap_application)
     "Snap Application ##{snap_application.id}"
--- a/app/dashboards/snap_application_dashboard.rb
+++ b/app/dashboards/snap_application_dashboard.rb
@@ -122,8 +122,8 @@ class SnapApplicationDashboard < Administrate::BaseDashboard
     total_money
   ].freeze
 
-  ATTRIBUTE_TYPES = (FORM_ATTRIBUTES + SHOW_ATTRIBUTES + COLLECTION_ATTRIBUTES).
-    uniq.freeze
+  ATTRIBUTE_TYPES = (FORM_ATTRIBUTES + SHOW_PAGE_ATTRIBUTES +
+                     COLLECTION_ATTRIBUTES).uniq.freeze
 
   def display_resource(snap_application)
     "Snap Application ##{snap_application.id}"
--- a/app/dashboards/snap_application_dashboard.rb
+++ b/app/dashboards/snap_application_dashboard.rb
@@ -1,6 +1,65 @@
 require "administrate/base_dashboard"
 
 class SnapApplicationDashboard < Administrate::BaseDashboard
+  ATTRIBUTE_TYPES = {
+    id: Field::Number,
+    additional_income: Field::Text,
+    anyone_disabled: Field::Boolean,
+    anyone_in_college: Field::Boolean,
+    anyone_living_elsewhere: Field::Boolean,
+    anyone_new_mom: Field::Boolean,
+    care_expenses: Field::String,
+    consent_to_terms: Field::Boolean,
+    court_ordered: Field::Boolean,
+    court_ordered_expenses: Field::String,
+    created_at: Field::DateTime,
+    dependent_care: Field::Boolean,
+    documents: Field::String,
+    email: Field::String,
+    emailed_at: Field::DateTime,
+    everyone_a_citizen: Field::Boolean,
+    fax_metadata: Field::String,
+    faxed_at: Field::DateTime,
+    financial_accounts: Field::String,
+    income_change: Field::Boolean,
+    income_change_explanation: Field::Text,
+    income_child_support: Field::Text,
+    income_foster_care: Field::String,
+    income_other: Field::String,
+    income_pension: Field::String,
+    income_social_security: Field::String,
+    income_ssi_or_disability: Field::String,
+    income_unemployment_insurance: Field::String,
+    income_workers_compensation: Field::String,
+    insurance_expense: Field::String,
+    mailing_address_same_as_residential_address: Field::Boolean,
+    medical: Field::Boolean,
+    medical_expenses: Field::String,
+    money_or_accounts_income: Field::Boolean,
+    monthly_care_expenses: Field::String,
+    monthly_court_ordered_expenses: Field::String,
+    monthly_medical_expenses: Field::String,
+    phone_number: Field::String,
+    property_tax_expense: Field::Number,
+    real_estate_income: Field::Boolean,
+    rent_expense: Field::Number,
+    signature: Field::String,
+    signed_at: Field::DateTime,
+    sms_subscribed: Field::Boolean,
+    total_money: Field::Number,
+    unstable_housing: Field::Boolean,
+    updated_at: Field::DateTime,
+    utility_cooling: Field::Boolean,
+    utility_electrity: Field::Boolean,
+    utility_heat: Field::Boolean,
+    utility_other: Field::Boolean,
+    utility_phone: Field::Boolean,
+    utility_trash: Field::Boolean,
+    utility_water_sewer: Field::Boolean,
+    vehicle_income: Field::Boolean,
+    zip: Field::String,
+  }.freeze
+
   COLLECTION_ATTRIBUTES = %i[
     id
     signature
@@ -121,9 +180,6 @@ class SnapApplicationDashboard < Administrate::BaseDashboard
     financial_accounts
     total_money
   ].freeze
-
-  ATTRIBUTE_TYPES = (FORM_ATTRIBUTES + SHOW_PAGE_ATTRIBUTES +
-                     COLLECTION_ATTRIBUTES).uniq.freeze
 
   def display_resource(snap_application)
     "Snap Application ##{snap_application.id}"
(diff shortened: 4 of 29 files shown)


zspencer:
[Epic: MVP Admin Dashboard](https://trello.com/c/Pjvb1NGG/188-epic-mvp-admin-dashboard)

zspencer:
@alanjosephwilliams or @adarsh - Here's where the UI is at right now, most of it doesn't work, but I'm pretty happy with my grasp on how to extend administrate UI successfully, so that's nice.

screen shot 2017-08-30 at 5 15 50 pm

app/assets/stylesheets/michigan-benefits/administrate-extensions.scss
```diff
@@ -0,0 +1,14 @@
+.main-content__funnel {
```
jessieay: I remember you saying that you struggled to figure out how to do this w/administrate  -- thoughts on linking to relevant docs as a comment at the top of this file?

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,5 +17,23 @@ class SnapApplicationsController < Admin::ApplicationController
 
     # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
     # for more information
+    #
+    def resend_fax
+      application = SnapApplication.find(params[:id])
+      application.exports.faxed.succeeded.update_all(status: :failed, metadata:
```
jessieay: super minor style thing: I think this would be more readable as

```
application.exports.faxed.succeeded.update_all(
   status: :failed, 
   metadata:  "Reset by admin",
)
```

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,5 +17,23 @@ class SnapApplicationsController < Admin::ApplicationController
...
+                                                     "Reset by admin")
+      Export.create_and_enqueue!(snap_application: application, destination:
+                                 :fax)
+      flash[:notice] = "Resent fax #{application.email}!"
```
jessieay: a bit confused about why we'd put the email address in the flash if we are sending a fax?

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,5 +17,23 @@ class SnapApplicationsController < Admin::ApplicationController
...
+      redirect_to admin_root_path
+    end
+
+    def pdf
+      application = SnapApplication.find(params[:id])
+      send_data application.pdf.read,
```
jessieay: parens around args here for easier reading?

app/dashboards/snap_application_dashboard.rb
```diff
@@ -3,25 +3,29 @@
...
-    signed_at: Field::DateTime,
-    email: Field::String,
-    documents: Field::String,
-    phone_number: Field::String,
-    sms_subscribed: Field::Boolean,
-    consent_to_terms: Field::Boolean,
-    mailing_address_same_as_residential_address: Field::Boolean,
-    unstable_housing: Field::Boolean,
-    everyone_a_citizen: Field::Boolean,
+    additional_income: Field::Text,
```
jessieay: are we showing every field right now? If not, I'd probably opt to not include these array fields, since we mostly use them to determine which fields we need values for.Eg: if there is an additional income from pension, we get a value for `additional_income_pension` -- so showing both is kind of duplicative.

understand if we want to start with all fields are pare down later, too

app/jobs/email_application_job.rb
```diff
@@ -1,16 +1,14 @@
...
 
-  private
+    export.execute do |snap_application|
+      ApplicationMailer.snap_application_notification(
+        file_name: snap_application.pdf.path,
+        recipient_email: snap_application.email,
+      ).deliver
 
-  attr_reader :snap_application
+      "Emailed to #{snap_application.email}"
```
jessieay: is this a logging statement? if so, do we want to explicitly pass it to `Rails.logger` ?

app/jobs/fax_application_job.rb
```diff
@@ -1,25 +1,18 @@
...
+        number: fax_recipient.number,
+        file: snap_application.pdf.path,
+        recipient: fax_recipient.name,
+      )
 
-  def fax_recipient
-    @fax_recipient ||=
-      FaxRecipient.new(residential_address:
-                         snap_application.residential_address)
+      "Faxed to #{fax_recipient.name} (#{fax_recipient.number})"
```
jessieay: same comment re: logging

app/models/export.rb
```diff
@@ -0,0 +1,88 @@
...
+class Export < ApplicationRecord
+  class UnknownExportTypeError < StandardError; end
+
+  include Informed
+
+  belongs_to :snap_application
+  validates :status, inclusion: { in: %i(new queued in_process succeeded failed
+                                         unnecessary) }
+  validates :destination, inclusion: { in: %i(fax email) }
+  before_validation { self.destination = destination.to_sym }
```
jessieay: always happy to avoid callbacks -- any way we can move these outside of `before_validation` ?

app/models/export.rb
```diff
@@ -0,0 +1,88 @@
...
+  default_scope { order(updated_at: :desc) }
+  scope :faxed, -> { where(destination: :fax) }
+  scope :emailed, -> { where(destination: :email) }
+  scope :succeeded, -> { where(status: :succeeded) }
+  scope :completed, -> { where.not(completed_at: nil) }
+  scope :application_ids, -> { pluck(:snap_application_id) }
+  scope :latest, -> { first }
+
+  def self.enqueue_faxes
+    SnapApplication.faxable.find_each do |snap_application|
```
jessieay: woooo `find_each`

app/models/export.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  def self.create_and_enqueue!(params)
+    create!(params).enqueue
+  end
+
+  # Queues up the export job for the given destination
+  def enqueue
+    transaction do
+      save!
```
jessieay: feels a bit weird to me to call `save!` in here since I'd expect it just to enqueue (feels like it is doing two things instead of one) -- thoughts?

app/models/export.rb
```diff
@@ -0,0 +1,88 @@
...
+        EmailApplicationJob.perform_later(export_id: id)
+      else
+        raise UnknownExportTypeError, destination
+      end
+      transition_to(new_status: :queued)
+    end
+  end
+  inform_on :enqueue, level: :info, also_log: { values: %i[destination status] }
+
+  def execute
```
jessieay: this method feels pretty hairy -- any way we could simplify?

app/models/export.rb
```diff
@@ -0,0 +1,88 @@
...
+    snap_application.pdf.try(:unlink)
+  end
+  # inform_on :execute, level: :info, also_log: { values: [:status] }
+
+  def transition_to(new_status:)
+    update!(status: new_status)
+  end
+
+  def queued?
+    status == :succeeded
```
jessieay: is this logic correct?

app/views/admin/snap_applications/_collection.html.erb
```diff
@@ -0,0 +1,97 @@
...
+          <td class="cell-data cell-data--<%= attribute.html_class %>">
+            <a href="<%= polymorphic_path([namespace, resource]) -%>"
+               class="action-show"
+               >
+              <%= render_field attribute %>
+            </a>
+          </td>
+        <% end %>
+
+        <td><%= button_to("Refax", resend_fax_admin_snap_application_path(resource), class: "button", method: :post) %></td>
```
jessieay: faxing something to an office is kind of a big deal, should we add a quick js alert here so it's harder to accidentally do it?

app/views/admin/snap_applications/_collection.html.erb
```diff
@@ -0,0 +1,97 @@
...
+        <td><%= link_to("Download", pdf_admin_snap_application_path(resource)) %></td>
```
jessieay: ❤️

app/views/admin/snap_applications/_funnel_stage.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="main-content__funnel__stage">
+  <h4><%= current_stage.count %> <%= name %>
+      <%- if previous_stage.present? %>
+        (<%= number_to_percentage((current_stage.count.to_f / previous_stage.count.to_f) * 100, precision: 1) %>)
```
jessieay: possible to throw this line in a method of some sort? feels like it could use a test

config/initializers/informed.rb
```diff
@@ -0,0 +1 @@
+Informed.logger = Rails.logger
```
jessieay: thoughts on putting this logger stuff in a separate PR? feels separate from this admin feature to me

spec/models/export_spec.rb
```diff
@@ -0,0 +1,89 @@
+require "rails_helper"
+
+RSpec.describe Export do
+  describe ".enqueue_faxes" do
+    it "schedules jobs for signed, unfaxed application updated awhile ago" do
+      signed_unfaxed_updated_awhile_ago = create(:snap_application,
+                                                 signed_at: 31.minutes.ago,
```
jessieay: `31` is referenced a lot here and feels like a magic number. What about referencing a constant instead?

spec/models/export_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  describe "#enqueue" do
+    before do
+      allow(FaxApplicationJob).to receive(:perform_later)
+      allow(EmailApplicationJob).to receive(:perform_later)
+    end
+
+    it "sends fax" do
+      export = FactoryGirl.build(:export, destination: "fax")
```
jessieay: no need for the `FactoryGirl` -- we're including syntax methods here 🌈

config/initializers/informed.rb
jessieay: We have a card "up next" for logging -- I could tackle that and use the work you have here w/r/t logging https://trello.com/c/WqpFBnYO/179-develop-strategy-for-logging

zspencer: @jessieay Go for it! I was using it mostly for debugging and am happy to rip it out.

app/views/admin/snap_applications/_collection.html.erb
zspencer: Sure thing!

app/jobs/fax_application_job.rb
zspencer: These are what is stored as metadata after the export happens so that the admin dashboard says what actually happened. I didn't want to make the data structured just yet sicne Iw asn't sure which data would be needed.

app/dashboards/snap_application_dashboard.rb
zspencer: @jessieay - I *think* this was auto-generated, we just re-sorted them alphabetically. I'm OK to pare it down tho!

app/controllers/admin/snap_applications_controller.rb
zspencer: I was trying to figure out what data would be most useful, and totally blanked on using signature + fax recipient number.

app/controllers/success_controller.rb
```diff
@@ -12,14 +12,14 @@ def next_path
   private
 
   def before_rendering_edit_hook
-    FaxApplicationJob.
-      perform_later(snap_application_id: current_snap_application.id)
+    Export.create_and_enqueue(destination: :fax,
```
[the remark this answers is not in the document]
zspencer: yeaaaaaaaaaaaaa, I started with ExportAttempt, then went down to Export, and I'm still not super pleased. I was thinking that Faxing, Emailing, and uhh WebDrivingViaBridges are all the same kind of thing; and we will likely want to include a record of every time we try and execute one of those things, including if it failed and why.

app/models/export.rb
zspencer: Nope! This method may be dead.

app/models/export.rb
zspencer: This protects us from starting a job with a nil export. i.e. `Export.new(...).enqueue` which would break the linkage between the snap application and the particular job to be done.

app/models/export.rb
zspencer: Yeah... I think Rails 5 has types for attributes? can we make Status a symbol?

app/models/export.rb
```diff
@@ -0,0 +1,68 @@
+# Adapter to fire off the correct job, track transitions and responses.
```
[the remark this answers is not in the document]
zspencer: It was more a signal because I wasn't happy with the name. Maybe `SnapApplicationJobAdapter`?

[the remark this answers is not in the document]
jessieay: sure! other ideas:

`ExportJobAdapter`
`Send1171PdfAdapter`

app/models/export.rb
jessieay: are you talking about using AR:Enum? http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

app/assets/stylesheets/michigan-benefits/administrate-extensions.scss
zspencer: Sadly it wasn't actually documented, I had to dig into the source code to find it :(

jessieay: even better! if you can [permalink](https://help.github.com/articles/getting-permanent-links-to-files/) to the source code that you found, that would be awesome

lib/tasks/data_integrity.rake
```diff
@@ -0,0 +1,18 @@
+namespace :data_integrity do
```
zspencer: We should move this to `one_offs` so it can be garbage collected later.

config/initializers/administrate.rb
```diff
@@ -0,0 +1 @@
+Administrate::Engine.add_stylesheet "michigan-benefits/administrate-extensions"
```
zspencer: Link to https://github.com/thoughtbot/administrate/blob/736266dd81f0a1b5ded6a268b93bd7de5c6c1a6f/lib/administrate/engine.rb#L31

jessieay: that is highlighting a line of code that just has an `end` -- typo?

zspencer: I totally forgot you can highlight an entire range:
https://github.com/thoughtbot/administrate/blob/736266dd81f0a1b5ded6a268b93bd7de5c6c1a6f/lib/administrate/engine.rb#L25-L31

I';m going to add them as comments; but like I said there isn't actual docs for this :(

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,5 +17,25 @@ class SnapApplicationsController < Admin::ApplicationController
 
     # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
     # for more information
+    #
+    def resend_fax
+      application = SnapApplication.find(params[:id])
+      application.exports.faxed.succeeded.update_all(
+        status: :failed,
+        metadata: "Reset by admin",
```
jessieay: "Resent" ? (spelling)

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,5 +17,25 @@ class SnapApplicationsController < Admin::ApplicationController
 
     # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
     # for more information
+    #
+    def resend_fax
+      application = SnapApplication.find(params[:id])
+      application.exports.faxed.succeeded.update_all(
```
jessieay: so we want to update successful faxes to show as failed? I am kind of confused.

I think it being set to failed so that it can re-run. Is there a way to do something differently to make it less confusing? Maybe use a different export method like `admin_fax` that works even if something has already been faxed?

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,5 +17,25 @@ class SnapApplicationsController < Admin::ApplicationController
...
+        status: :failed,
+        metadata: "Reset by admin",
+      )
+      Export.create_and_enqueue!(snap_application: application, destination:
+                                 :fax)
+      flash[:notice] = "Resent fax #{application.email}!"
```
jessieay: this flash message is kind of confusing -- we faxed an email? Also, email could be nil so this flash message might be unhelpful in some cases

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,5 +17,25 @@ class SnapApplicationsController < Admin::ApplicationController
...
+      redirect_to admin_root_path
+    end
+
+    def pdf
```
jessieay: this may be too much, but would it be possible to make these routes RESTful?

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,5 +17,25 @@ class SnapApplicationsController < Admin::ApplicationController
...
+      application = SnapApplication.find(params[:id])
+      send_data(application.pdf.read,
+        filename: "snap-application-#{application.email}.pdf",
```
jessieay: email is not a required attribute -- maybe use DOB or name instead?

app/controllers/success_controller.rb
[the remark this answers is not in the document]
jessieay: We can always rename later!

app/jobs/fax_application_job.rb
```diff
@@ -1,25 +1,18 @@
...
-  ensure
-    snap_application.pdf.try(:close)
-    snap_application.pdf.try(:unlink)
-  end
-
-  private
+  def perform(export_id:)
+    export = Export.find(export_id)
+    export.execute do |snap_application|
+      return if snap_application.faxed?
```
jessieay: so won't this fail if an admin is trying to re-fax?

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,5 +17,25 @@ class SnapApplicationsController < Admin::ApplicationController
...
     # for more information
+    #
+    def resend_fax
+      application = SnapApplication.find(params[:id])
+      application.exports.faxed.succeeded.update_all(
+        status: :failed,
+        metadata: "Reset by admin",
+      )
+      Export.create_and_enqueue!(snap_application: application, destination:
+                                 :fax)
```
jessieay: did you see my previous comment about style? It's small but I do think this format is harder to read than
```
Export.create_and_enqueue!(
  snap_application: application,
  destination: :fax,
)
```

db/migrate/20170831164445_create_exports.rb
```diff
@@ -0,0 +1,12 @@
+class CreateExports < ActiveRecord::Migration[5.1]
+  def change
+    create_table :exports do |t|
+      t.references :snap_application
+      t.string :destination
```
jessieay: did you think about using active record enum for this like I suggested previously? Would make the code work the same as it already does but this would be an integer field

app/jobs/email_application_job.rb
```diff
@@ -1,16 +1,14 @@
...
-    ).deliver
-  ensure
-    snap_application.pdf.try(:close)
-    snap_application.pdf.try(:unlink)
-  end
+  def perform(export_id:)
+    export = Export.find(export_id)
 
-  private
+    export.execute do |snap_application|
```
jessieay: similar to how we check if an app was faxed before faxing, should we check that it was emailed?

app/jobs/fax_application_job.rb
jessieay: gotcha - any way to make the fact that it's metadata more obvious? To me these read as random strings! 🍓

spec/models/export_spec.rb
```diff
@@ -0,0 +1,92 @@
+require "rails_helper"
+
+RSpec.describe Export do
+  describe ".enqueue_faxes" do
+    it "schedules jobs for signed, unfaxed application updated awhile ago" do
+      A_WHILE_AGO = 31.minutes.ago
```
jessieay: sorry, when I commented before I was not clear on my thinking about constants in this context.

What I meant to say that the fact that one is more than 30 mins (a while ago, in this app) and one is less than 30 mins (less than a while ago, in this app) is where the magic number comes in. 

That magic number is in `app/models/snap_application.rb`. Perhaps create a cleanup ticket to change that to a constant in a separate PR and then here, when we add that constant, we can say `CONSTANT.minutes + 1.minute` or  `CONSTANT.minutes - 5.minutes` -- does that make sense?

spec/models/funnel_stage_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    context "when not given a previous cohort" do
+      it "starts from 100% conversion" do
+        funnel_stage = FunnelStage.new(
+          cohort: Array.new(3),
+          previous_cohort: [],
+          name: "Signup",
+        )
+        expect(funnel_stage.conversion_rate).to eq(1)
```
jessieay: very minor: it's a tad confusing that the spec description is in percent but the values returned are out of 1.

app/controllers/admin/snap_applications_controller.rb
zspencer: Good call. Using signature.

app/dashboards/snap_application_dashboard.rb
zspencer: I've pruned this down to only the fields we expose in the index, show and edit pages,

app/jobs/email_application_job.rb
zspencer: I'm not sure, part of me thinks that "multiple emails probably isn't that big a deal" but I am totally down for making those consistent.

app/jobs/fax_application_job.rb
zspencer: I've been thinking about this, and it may be an indication we're mixing command/query separation. Perhaps the jobs should have a "success_message" in their interface? And then assume they succeeded in the event that no exceptions were thrown?

config/routes.rb
```diff
@@ -2,7 +2,10 @@
 
 Rails.application.routes.draw do
   namespace :admin do
-    resources :snap_applications
+    resources :snap_applications do
+      post "resend_fax", on: :member
+      get "pdf", on: :member
```
zspencer: If I get through the other feedback by lunch I'll make these RESTful

db/migrate/20170831164445_create_exports.rb
zspencer: In the past I've been bitten in the ass by changing the available options on an active_record enum fields. (i.e. given `:a, :b, :d` as options, when you add `:c` things that were reading as `:d` are now reading as `:c`) 

This is *probably* fixed by now, but I'm still ActiveRecord::Enum shy.

app/jobs/fax_application_job.rb
zspencer: And I just realized that since this is using a block as an interface, we don't have a reference to the actual job... which... uhh... I feel conflicted about.

PR #163, opened 2017-08-31
Fallback office is clio, fix unions number

zspencer:
#### Why?
Because [Lena said that union doesn't frequently check their fax and hasn't had training](https://cfa.slack.com/archives/C4BRQQEUU/p1504196503000049), while Clio has.

Also, because Unions fax number was wrong :X :X :X


zspencer:
[Final fax tweaks](https://trello.com/c/tiljTPIz/191-final-fax-tweaks)

PR #165, opened 2017-08-31
Filter out ssns from logging statements in params

zspencer:
### Why?
Logging social security numbers is A Bad Idea


zspencer:
[Add in params filtering to exclude SSN](https://trello.com/c/JP5xe7kf/195-add-in-params-filtering-to-exclude-ssn)

PR #166, opened 2017-08-31
Fax office immediately after applicant indicates they are done

zspencer:
### Why? 
Waiting for a fax is not a great user experience when demoing.

### DIFF
--- a/app/controllers/standard_steps_controller.rb
+++ b/app/controllers/standard_steps_controller.rb
@@ -4,7 +4,7 @@ class StandardStepsController < StepsController
   def edit
     @step = step_class.new(existing_attributes.slice(*step_attrs))
 
-    before_rendering_edit
+    before_rendering_edit_hook
   end
 
   def update
@@ -12,7 +12,7 @@ class StandardStepsController < StepsController
 
     if @step.valid?
       current_snap_application.update!(step_params)
-      after_successful_update
+      after_successful_update_hook
       redirect_to(next_path)
     else
       render :edit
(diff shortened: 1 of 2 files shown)


app/controllers/standard_steps_controller.rb
```diff
@@ -18,6 +21,14 @@ def update
 
   private
 
+  def after_successful_update
+
```
adarsh: Can you put a `#noop` comment in here to just clarify it's emptiness? 🕳

app/controllers/success_controller.rb
```diff
@@ -1,26 +1,22 @@
...
-
   def previous_path(*_args)
     nil
   end
 
+  def next_path
+    root_path(anchor: "fold")
+  end
+
   private
```
adarsh: Style: newline after `private`

app/controllers/standard_steps_controller.rb
```diff
@@ -3,13 +3,16 @@
 class StandardStepsController < StepsController
   def edit
     @step = step_class.new(existing_attributes.slice(*step_attrs))
+
+    before_rendering_edit
```
adarsh: Naming: `before_rendering_edit` > `before_rendering_edit_action`? or `before_edit_action`?

app/controllers/standard_steps_controller.rb
zspencer: Yep!

app/controllers/standard_steps_controller.rb
```diff
@@ -3,13 +3,16 @@
...
   end
 
   def update
     @step = step_class.new(step_params)
 
     if @step.valid?
       current_snap_application.update!(step_params)
+      after_successful_update
```
adarsh: Naming: `after_successful_update` > `after_successful_update_action`

app/controllers/standard_steps_controller.rb
```diff
@@ -21,8 +21,13 @@ def update
 
   private
 
+  # Hook for step controllers to add behavior after successful update. This is
+  # intentionally a no-op in the StandardStepsController.
```
adarsh: I'd prefer the method name be clearer than we add comments. 

What about `after_successful_update_hook`?

zspencer: Done.

app/controllers/success_controller.rb
zspencer: As in there shouldn't be one?

PR #169, opened 2017-09-05
Install foreman as part of bin/setup

adarsh:
Reason for Change
=================
* It is required for local development but NOT in the Gemfile.
* This could be confusing.
* Why not just set it up in `bin/setup`?

Changes
=======
* Add a step to install the `foreman` gem.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -8,8 +8,12 @@ chdir APP_ROOT do
   # Add necessary setup steps to this file.
 
   step "Installing dependencies" do
-    system! "gem install bundler --conservative"
+    system!("gem install bundler --conservative")
     system("bundle check") || system!("bundle install")
+  end
+
+  step "Installing foreman for controlling local servers" do
+    system!("gem install foreman")
   end
 
   step "Installing pdftk, a PDF library (requires Homebrew)" do


adarsh:
[README and Contributing Guidelines updates based CZI](https://trello.com/c/4lHSn8NQ/194-readme-and-contributing-guidelines-updates-based-czi)

bin/setup
```diff
@@ -8,10 +8,14 @@ chdir APP_ROOT do
   # Add necessary setup steps to this file.
 
   step "Installing dependencies" do
-    system! "gem install bundler --conservative"
+    system!("gem install bundler --conservative")
     system("bundle check") || system!("bundle install")
   end
 
+  step "Installing foreman for controlling local servers" do
+    system!("gem install foreman")
```
jessieay: should we check if foreman is already installed before doing this?

jayroh: What's the down-side to placing it in the Gemfile explicitly?

jessieay: https://github.com/thoughtbot/suspenders/pull/330 :)

zspencer: I tend to lean in the direction of "Things that live completely outside the application should not be in the Gemfile"  (Say, `rerun` or `mailcatcher`) but have no strong preferences here.

jayroh:
Touché!

bin/setup
adarsh: > should we check if foreman is already installed before doing this?
 
I thought of this, but if we call `gem install` and it **is** installed, bundler will just upgrade it. So I'm inclined to leave it, unless you feel like that is an unpleasant surprise.

PR #173, opened 2017-09-06
Add sample application with built in db:seed task

jayroh:
For things like testing the web driving, it is required that we have a
"finished" application that can be used for that work.

This PR adds sample data that came from one end-to-end
submission with the bare minimum. With regards to the data - nils, empty
arrays, or anything that was "barren" was removed.

Also included - a snap application with a more complete profile.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,1 +1,178 @@
 # frozen_string_literal: true
+
+puts "Creating minimal snap application..."
+
+minimal_application = SnapApplication.
+  where(email: "minimal.application@example.com").
+  first_or_create(
+    signature: "Minimal Application",
+    signed_at: Time.current,
+    phone_number: "5551230000",
+    sms_subscribed: false,
+    consent_to_terms: true,
+    everyone_a_citizen: true,
+    anyone_disabled: false,
+    anyone_new_mom: false,
+    anyone_in_college: false,
+    anyone_living_elsewhere: false,
+    income_change: false,
+    rent_expense: 500,
+    property_tax_expense: 0,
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1 +1,61 @@
 # frozen_string_literal: true
+
+puts "Creating test snap application..."
+
```
zspencer: _not blocking_ - Does this make sense to make a factory?

jayroh: You *could* do that, sure. I don't have any strong feelings towards making this use factory girl. However, that would require pulling in FG to the development group, and maintaining it over in that context is a jump away.

In truth, I have no argument for or against it. What about you, @zspencer?

jayroh:
> Thoughts on potentially adding the other half (a "full" application, which would have multiple members, different employment statuses etc)

Yup. I can do that.

jayroh:
[Easy way to create a minimal or full 1171 application in order to do QA](https://trello.com/c/DWz25eGR/185-easy-way-to-create-a-minimal-or-full-1171-application-in-order-to-do-qa)

jayroh:
@jessieay I'm not sure *how* complete we want it but my latest commit adds another snap application that has multiple members, multiple addresses, more answers.

db/seeds.rb
zspencer: Yeah I'm fine leaving it as a normal model. One of the things I was thinking about for testing was to add to the admin dashboard  SNAPApplication generator which could use factory girl traits and what not to build an app; but I don't think that's necessary for now.

PR #174, opened 2017-09-06
Add verification documents to 1171 PDF

jessieay:
**WHY**: so we can verify identity.

Also adds a header to docs with name and birthdate of primary applicant.

NOTE: Uploading verification docs that are PDFs does not work yet. Going to tackle that next.


jessieay:
[Attach Uploaded PDFs and Images (NOT DOC/DOCX) to 1171 PDF](https://trello.com/c/Lm4NpZ4T/197-attach-uploaded-pdfs-and-images-not-doc-docx-to-1171-pdf)

PR #175, opened 2017-09-06
Click through benefits and fill in personal info

jayroh:
There are two benefit overview pages that the driver can click through
before coming to a more involved form page where they enter personal
information:

* Name
* Gender
* Birth date
* Address/Location
* Phone number

If the client has an address that MIBridges considers an invalid physical
address it will re-display that personal information page and confirm that
the client *still* wants to submit with that street address. If so, this
will confirm that this is, indeed, the address we would like to continue
with.

Also...
-------

* Use more specific selector for agreement checkbox
* Remove demo preable w/google search page. For demo purposes we had the driver
  go to Google to search for MI Bridges before starting an application. For
  purposes of actually driving the application, however, we do not need this. So,
  let's remove it.

### DIFF
--- a/lib/mi_bridges/driver.rb
+++ b/lib/mi_bridges/driver.rb
@@ -6,10 +6,6 @@ require "capybara/drivers/headless_chrome"
 
 module MiBridges
   class Driver
-    DEMO_SETUP = [
-      GoogleSearchPage,
-    ].freeze
-
     SIGN_UP_FLOW = [
       HomePage,
       CreateAccountPage,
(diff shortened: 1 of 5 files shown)


jessieay:
Would it be possible to start writing some tests for this, or will that come later?

jayroh:
> Would it be possible to start writing some tests for this, or will that come later?

I think this is too volatile at this point to spend too much time writing tests. I'd still consider this all a spike just to see where the problems could be, or are, which would then inform us of what and how we wanted to test.

The testing question is one of those things I would love to discuss as a group because this portion of the app is so *ultra* out of the ordinary.

PR #176, opened 2017-09-06
Allow PDFs verification docs to be appended

jessieay:
* This solution is basically to not add the header to docs if they are
PDFs. This is what GetCalFresh does
* Also adds webmock here so that we are not stubbing `open` and instead
return a real ficture of a web response (made with curl) so that we are
testing remote PDF downloading properly.


jessieay:
[Attach Uploaded PDFs and Images (NOT DOC/DOCX) to 1171 PDF](https://trello.com/c/Lm4NpZ4T/197-attach-uploaded-pdfs-and-images-not-doc-docx-to-1171-pdf)

app/models/verification_document.rb
```diff
@@ -7,33 +7,31 @@ def initialize(url:, snap_application:)
...
-      magick_object.format("png")
-    end
-
-    magick_object
+    MiniMagick::Image.open(local_file.path)
   end
 
   def local_file
-    open(url)
+    @_local_fle ||= MiniMagick::Image.open(url)
```
zspencer: Not blocking - Spelling :D

jessieay: eagle 👁  ! 

fixed at b2d534b

jessieay:
@zspencer thanks for the suggestion! these tests are definitely much stronger now

jessieay:
[add webmock so that we are not hitting the live internet in tests](https://trello.com/c/VcF8jS7c/222-add-webmock-so-that-we-are-not-hitting-the-live-internet-in-tests)

PR #180, opened 2017-09-07
Fix the document uploading UX

jayroh:
## Problem

When adding documents to an application, there is a duplication of
elements - the first being the "preview" and the second being the final
uploaded image/file.

## Solution

When an upload is sucessful or results in an error, the preview element
is removed from the DOM.

## Problem

As files are being uploaded, the "done uploading" button is able to be
clicked before things are completely finished uploading. This will
occasionally result in the form being submitted before a/some file(s)
are completely uploaded.

## Solution
The "DONE" button is disabled until uploading is finished.

# Demo

### DIFF
--- a/app/assets/javascripts/upload_setup.js
+++ b/app/assets/javascripts/upload_setup.js
@@ -7,12 +7,11 @@ var uploadSetup = function (previewTemplate) {
     previewTemplate: previewTemplate,
     acceptedFiles: '.pdf,.jpg,.jpeg,.png,.gif',
     maxFilesize: 3, // MB
-    error: function (_file, msg) {
+    error: function (file, msg) {
       window.alert(msg)
+      $(file.previewElement).remove()
     },
-    success: function (f) {
-      $('.button--next').removeAttr('disabled')
-
+    success: function (file) {
       var button = $('#click-to-upload')
       var buttonHtml = button.html()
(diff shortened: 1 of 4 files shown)


jayroh:
[Issues uploading images](https://trello.com/c/N8LHOkhZ/200-issues-uploading-images)

jayroh:
Feature spec would be tough in this context.

A JS test harness (jasmine, konacha, mocha, chai, et al) being set up would make this easier, for sure. Those callbacks could be extracted and tested properly. Doable, yes, but it'd take some time/effort that's not within the scope here.

zspencer:
I am really in favor of us taking a few hours (in a different PR) to get a UI test harness in place. I am pretty comfortable with Jasmine, but I know that it's not super great for Async, and am happy to defer to all y'alls judgement there.

jessieay:
I don't think js tests should block this  because it contains important UX improvements but I will note that we have a card in up next for adding a js test harness: https://trello.com/c/pfOLyu4l/104-add-a-javascript-test-harness-so-that-all-new-js-is-and-can-be-tested-going-forward

YAY

jessieay:
[Bug: can submit verification documents before uploaded successfully](https://trello.com/c/I1ahwDrT/223-bug-can-submit-verification-documents-before-uploaded-successfully)

jessieay:

PR #183, opened 2017-09-07
Show signature in EST on PDF

jessieay:
**WHY**: Math is hard. And this reads better, anyway!


jessieay:
[Change UTC on Signature to EST](https://trello.com/c/gR86dMYm/237-change-utc-on-signature-to-est)

PR #185, opened 2017-09-07
Brief guide for deploying

zspencer:
**Why?** - Deploying is scary! Lists make it less scary. Automation makes it even less scarier; but I didn't want to dive too deep too fast.

### DIFF
--- a/DEPLOYING.md
+++ b/DEPLOYING.md
@@ -1,19 +1,15 @@
 ### Timing for Deploys
 
-* Staging: Deployed on successful builds from Master.
+* Staging: Deployed automatically on successful builds from Master.
 * Production: Deployed after acceptance at standup.
 
-### Pre-Production Checklist
-- [ ] Review the buildpacks on staging and production. Are they the same? If
-  not, make them the same.
-- [ ] Skim the diff between the current production sha and master
-  - `git fetch --all && git diff production/master staging/master`
-  - [ ] Any new ENV variables? If so, set them.
-  - [ ] What about rake tasks? If so, make note of them to run after deploy.
-  - [ ] Any scary looking migrations that could result in data loss or prevent
-    roll back? If so, perhaps we should hold off on the deploy until we fix the
-    scary migrations
+### Deploying to Production
+Before production, we want to make sure that:
+* All tickets that are awaiting acceptance have been accepted
+* We've done a quick code review just to make sure things seem legit
+* All configuration variables are set on prod that are necessary
+* All buildpacks are the same between staging and production
 
-### Actually Deploying Production
-- [ ] `heroku pipelines:promote -a michigan-benefits-staging`
-- [ ] Run post-deploy rake tasks
+To do this, run `bin/release` and it will walk through each of those steps,
+prompting you to say it's all good.
+
--- a/DEPLOYING.md
+++ b/DEPLOYING.md
@@ -1,0 +1,19 @@
+### Timing for Deploys
+
+* Staging: Deployed on successful builds from Master.
+* Production: Deployed after acceptance at standup.
+
+### Pre-Production Checklist
+- [ ] Review the buildpacks on staging and production. Are they the same? If
+  not, make them the same.
+- [ ] Skim the diff between the current production sha and master
+  - `git fetch --all && git diff production/master staging/master`
+  - [ ] Any new ENV variables? If so, set them.
+  - [ ] What about rake tasks? If so, make note of them to run after deploy.
+  - [ ] Any scary looking migrations that could result in data loss or prevent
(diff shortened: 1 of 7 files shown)


DEPLOYING.md
```diff
@@ -0,0 +1,19 @@
+### Timing for Deploys
+
+* Staging: Deployed on successful builds from Master.
+* Production: Deployed after acceptance at standup.
+
+### Pre-Production Checklist
+- [ ] Review the buildpacks on staging and production. Are they the same? If
+  not, make them the same.
+- [ ] Skim the diff between the current production sha and master
+  - `git fetch --all && git diff production/master staging/master`
```
jessieay: maybe provide github URL as well for those who are more comfortable viewing diffs there?

DEPLOYING.md
```diff
@@ -0,0 +1,19 @@
...
+  - [ ] Any new ENV variables? If so, set them.
+  - [ ] What about rake tasks? If so, make note of them to run after deploy.
```
jessieay: this would be for a one-off, right?

DEPLOYING.md
```diff
@@ -0,0 +1,19 @@
...
+  - [ ] Any scary looking migrations that could result in data loss or prevent
```
jessieay: seems like this should be a part of code review

DEPLOYING.md
```diff
@@ -0,0 +1,19 @@
...
+    roll back? If so, perhaps we should hold off on the deploy until we fix the
+    scary migrations
+
+### Actually Deploying Production
+- [ ] `heroku pipelines:promote -a michigan-benefits-staging`
+- [ ] Run post-deploy rake tasks
```
jessieay: maybe add a step for clicking around on prod to ensure it's not broken ?

bin/setup
```diff
@@ -7,6 +7,11 @@ chdir APP_ROOT do
   # This script is a starting point to setup your application.
   # Add necessary setup steps to this file.
 
+  step "Adding heroku remotes for `-r` happinesss" do
+    system!("heroku git:remote -a michigan-benefits-staging -r staging")
+    system!("heroku git:remote -a michigan-benefits-production -r production")
```
jessieay: yay!

DEPLOYING.md
zspencer: Yep!

DEPLOYING.md
zspencer: like, as a template for them to fill out or as a script to run that actually figures out what the URL should be? (I think both are valid)

DEPLOYING.md
zspencer: Yeah, I would hope that to be the case.

DEPLOYING.md
zspencer: Yah, I am also comfortable putting in a faux-office that faxes to our staging environment as part of this PR. Thoughts?

DEPLOYING.md
jessieay: either works for me!

DEPLOYING.md
jessieay: sure!

DEPLOYING.md
```diff
@@ -0,0 +1,19 @@
+### Timing for Deploys
+
+* Staging: Deployed on successful builds from Master.
```
adarsh: Maybe "Deployed _automatically_ on successful builds from Master"?

adarsh:
What do you all think about a `bin/deploy` script which encapsulates whatever details we have? It wouldn't cover things like "Hey, did you add some `ENV` references?" but it's something.

I found this one from a previous project:

```
#!/bin/sh

# Run this script to deploy the app to Heroku.
#
# ./bin/deploy (pushes to staging by default)
# ./bin/deploy staging
# ./bin/deploy production

set -e

branch="$(git symbolic-ref HEAD --short)"
target="${1:-staging}"

git push "$target" "$branch:master"
heroku run rake db:migrate --exit-code --remote "$target"
heroku restart --remote "$target"
```

jayroh:
> What do you all think about a bin/deploy script which encapsulates whatever details we have? 

Big fan of this. Deploys end up incrementally growing into a non-trivial process so this ends up being extremely helpful.

bin/release-diff
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+git fetch --all
+
+staging_sha=$(git log --oneline staging/master | head -n 1 | cut -d " " -f1)
+prod_sha=$(git log --oneline production/master | head -n 1 | cut -d " " -f1)
+
+open "https://github.com/codeforamerica/michigan-benefits/compare/$prod_sha...$staging_sha"
```
jessieay: beauty!

PR #186, opened 2017-09-08
Refactor js that warns when exiting application

jayroh:
Since we have several new links that need to be accounted for -- namely
privacy and terms -- this JS can/should be changed to be more amenable
to changes.

This update creates a whitelist of strings that the js will look for
and, if the link contains any of those strings, it will not consider it
an external link.

TLDR, if these are substrings of something you click, the warning
message won't pop up.

### DIFF
--- a/app/assets/javascripts/form_exit.js.erb
+++ b/app/assets/javascripts/form_exit.js.erb
@@ -5,10 +5,20 @@
     if(outsideOfSteps) return;
 
     for(var i=0; i<links.length; i++){
-      var stepsLink = links[i].href.indexOf('steps') >= 0
-      var skipLink = links[i].href.indexOf('skip_') >= 0
+      var isAnExternalLink = true
+      var whiteList = [
+        '/privacy',
+        '/steps',
+        '/terms',
+        'skip_'
+      ]
+      for(var j=0; j<whiteList.length; j++){
+        if(links[i].href.indexOf(whiteList[j]) >= 0){
+          isAnExternalLink = false
+        }
+      }
 
-      if(!stepsLink && !skipLink){
+      if(isAnExternalLink){
         links[i].addEventListener('click', function(event){
           if (!w.confirm('<%= I18n.t("javascript.exit_app_confirmation")%>')) {
             event.preventDefault()


jayroh:
[As a mobile user, I want to be able to agree to the Terms and Conditions without having to navigate through an onerous amount of text.](https://trello.com/c/TIstlODK/164-as-a-mobile-user-i-want-to-be-able-to-agree-to-the-terms-and-conditions-without-having-to-navigate-through-an-onerous-amount-of)

jayroh:
[As a mobile user, I want to be able to agree to the Terms and Conditions without having to navigate through an onerous amount of text.](https://trello.com/c/TIstlODK/164-as-a-mobile-user-i-want-to-be-able-to-agree-to-the-terms-and-conditions-without-having-to-navigate-through-an-onerous-amount-of)

jayroh:
[As a mobile user, I want to be able to agree to the Terms and Conditions without having to navigate through an onerous amount of text.](https://trello.com/c/TIstlODK/164-as-a-mobile-user-i-want-to-be-able-to-agree-to-the-terms-and-conditions-without-having-to-navigate-through-an-onerous-amount-of)

app/assets/javascripts/form_exit.js.erb
```diff
@@ -5,10 +5,20 @@
...
-      var stepsLink = links[i].href.indexOf('steps') >= 0
-      var skipLink = links[i].href.indexOf('skip_') >= 0
+      var isAnExternalLink = true
+      var whiteList = [
+        '/privacy',
+        '/steps',
+        '/terms',
+        'skip_'
+      ]
+      for(var j=0; j<whiteList.length; j++){
```
jessieay: style question: in JS do you like to keep no spaces or would it be better to have `j = 0; j < whiteList.length`? Can't tell if ruby style is infecting my brain here

app/assets/javascripts/form_exit.js.erb
```diff
@@ -5,10 +5,20 @@
...
+        if(links[i].href.indexOf(whiteList[j]) >= 0){
+          isAnExternalLink = false
+        }
+      }
 
-      if(!stepsLink && !skipLink){
+      if(isAnExternalLink){
         links[i].addEventListener('click', function(event){
```
jessieay: so for external links we want to warn them that the session is cleared? It doesn't look like that is what is happening in the linked gif (thanks for that, BTW!) but the way I am reading this, if an "external link" is clicked, we give a js alert. 

My thought was that we only wanted to show this warning if the link they click takes them to the root path?

app/assets/javascripts/form_exit.js.erb
zspencer: It would be nice if we had a rubocop for js set up to make these decisions for us.

PR #191, opened 2017-09-08
Add "contact preference" step

jessieay:
* SMS or Email
* Removes "interview preference" step for now since we are not sending
that data anywhere.
* Follow up to this will be a page that confirms someone's phone number
and email address based on what the opt into
* Follow up will also be to actually send the email / SMS confirms

screen shot 2017-09-08 at 11 01 21 am


jessieay:

PR #192, opened 2017-09-08
Enforce SSL on production-like environments

zspencer:
**Why?** Because we're dealing with hella sensitive data here; so HTTPS should be the de-rigor.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -3,6 +3,8 @@
 require Rails.root.join("config/smtp")
 
 Rails.application.configure do
+  config.force_ssl = true
+
   config.cache_classes = true
   config.eager_load = true
   config.consider_all_requests_local       = false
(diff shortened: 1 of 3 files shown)


zspencer:
[Fix: HTTPS is broken on prod with a naked URL](https://trello.com/c/lwUirm0R/201-fix-https-is-broken-on-prod-with-a-naked-url)

config/environments/production.rb
```diff
@@ -3,6 +3,8 @@
 require Rails.root.join("config/smtp")
 
 Rails.application.configure do
+  config.force_ssl = true
```
jessieay: should we remove the ssl-related code from `app/controllers/application_controller.rb` ?

config/environments/production.rb
```diff
@@ -34,14 +36,4 @@
   end
 
   config.active_record.dump_schema_after_migration = false
-
-  config.paperclip_defaults = {
```
jessieay: this seems unrelated?

zspencer: It was dead code so I wanted to purge it.

PR #193, opened 2017-09-08
Add error message for small photo sizes

rachelcope:
Reason for Change
===================
* Notify users when they upload a file smaller than 2MB
* https://trello.com/b/aBqTrqaJ/the-digital-assister-michiganbenefitsorg

Changes
=======
* Add error message

Note
=====
* This is only the copy and styles for the error message. These error messages still need to be implemented

cc: @jessieay 

image

### DIFF
--- a/app/assets/javascripts/upload_setup.js
+++ b/app/assets/javascripts/upload_setup.js
@@ -6,7 +6,7 @@ var uploadSetup = function (previewTemplate) {
     clickable: '#click-to-upload',
     previewTemplate: previewTemplate,
     acceptedFiles: '.pdf,.jpg,.jpeg,.png,.gif',
-    maxFilesize: 3, // MB
+    maxFilesize: 8, // MB
     error: function (file, msg) {
       window.alert(msg)
       $(file.previewElement).remove()
--- a/app/views/documents/_document.html.erb
+++ b/app/views/documents/_document.html.erb
@@ -16,4 +16,8 @@ end
   </div>
   <div class="doc-preview__thumb" style="background-image: url('<%= doc_url %>')">
   </div>
+  <p class="text--error">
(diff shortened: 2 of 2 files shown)


zspencer:
So I stumbled on `app/assets/javascript/upload_setup.js` which seems to set the max upload size to 3MB; this is a pretty darn small file size window?

jessieay:
@rachelcope trello link in the PR is the the board. Do you know where the specific ticket for this lives? can no longer find it

zspencer:
[As a user, I need a heads up that the photos I share to verify my identity aren't good enough to be used by MDHHS](https://trello.com/c/9dCWnUYh/239-as-a-user-i-need-a-heads-up-that-the-photos-i-share-to-verify-my-identity-arent-good-enough-to-be-used-by-mdhhs)

jayroh:
> So I stumbled on app/assets/javascript/upload_setup.js which seems to set the max upload size to 3MB; this is a pretty darn small file size window?

Agreed.

In the meantime, to get this merged would it make sense for someone like me to pick this up and get it hooked together with the error event?

app/assets/javascripts/upload_setup.js
```diff
@@ -6,7 +6,7 @@ var uploadSetup = function (previewTemplate) {
     clickable: '#click-to-upload',
     previewTemplate: previewTemplate,
     acceptedFiles: '.pdf,.jpg,.jpeg,.png,.gif',
-    maxFilesize: 3, // MB
+    maxFilesize: 8, // MB
```
jayroh: :+1:

zspencer:
@jayroh I'm closing this until the ticket it's part of gets to the top of up-next; since I think it still needs scoping?

PR #196, opened 2017-09-08
WIP - Prevent public access to documents uploaded to S3

zspencer:
**Why** - Because public docs is scary!


jessieay:

PR #197, opened 2017-09-08
Redirect to first step if there is no current app

jessieay:
**WHAT**: this changes redirect from root path to first step. It also
ensures that someone cannot visit '/steps' without a current snap app.

**WHY**:  Experience from new teammate that was confusing:
"If I tried to click on any of the links from /steps it kicks
back to the homepage (but it doesn’t actually throw an error)"

### DIFF
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -4,7 +4,7 @@ class StepsController < ApplicationController
   layout "step"
 
   before_action :maybe_skip, only: :edit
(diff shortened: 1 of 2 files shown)


jessieay:
[If someone has not created an app, do now throw an error when they try to visit a page](https://trello.com/c/nT7uzauM/213-if-someone-has-not-created-an-app-do-now-throw-an-error-when-they-try-to-visit-a-page)

PR #198, opened 2017-09-08
Update housing expense copy

rachelcope:
Reason for Change
===================
* Update copy to be more clear for users
* https://trello.com/c/TFFAYEvx/161-update-housing-expenses-page

image

### DIFF
--- a/app/views/expenses_housing/edit.html.erb
+++ b/app/views/expenses_housing/edit.html.erb
@@ -23,10 +23,13 @@
           autofocus: true %>
 
         <%= f.mb_money_field :property_tax_expense,
-        "How much do you pay in property tax each month?" %>
+          "How much do you pay in property tax each month?",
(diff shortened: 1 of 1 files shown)


app/views/document_guide/edit.html.erb
```diff
@@ -12,11 +12,6 @@
       <p>
         You can do this later but this will help speed up the process.
       </p>
-      <p>
-        The following page will allow you to both upload photos of your documents
-        with your phone, as well as submit files you already have on your
-        computer or mobile phone/tablet.
-      </p>
```
adarsh: Was this a related change? I didn't see it mentioned in the PR. Just curious!

rachelcope: @adarsh oh that is actually for a different card. I will remove it from this branch add to the other

PR #199, opened 2017-09-08
Remove migrations so that spinning up new apps is faster

zspencer:
**Why?** Welp, turns out the documents migration relied on paperclip and was causing new review apps from spinning up. So I wound up just squashing all the migrations down.


zspencer:
[Enable Heroku review apps for pre-merge acceptance testing.](https://trello.com/c/xqWZGdNh/246-enable-heroku-review-apps-for-pre-merge-acceptance-testing)

PR #202, opened 2017-09-11
Emphasize that sms is about consent not subscription

zspencer:
**Why** - In the future, we will want to use subscription to mean signing up for a drip campaign; plus we want to make the field name to reflect that it's a "you can SMS me about anything!"  kind of thing.


zspencer:

PR #204, opened 2017-09-11
Update navigation text across pages for consistency

jessieay:
* Took this opp to reorganize our feature specs as well.


jessieay:
[Across-App Navigation Update](https://trello.com/c/i3QxqfH0/243-across-app-navigation-update)

spec/features/snap_application_with_maximum_info_spec.rb
```diff
@@ -0,0 +1,250 @@
...
+    on_page "Documents" do
+      upload_documents
+      click_on "Done uploading documents"
+    end
+
+    on_page "Application Submitted" do
+      fill_in "Your email address", with: "test@example.com"
+      click_on "Submit"
+    end
+
```
zspencer: __not blocking__ - Thoughts on checking that the snap_application.pdf_attributes has all the data you expect here?

spec/support/snap_application_form_helper.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def select_address_same_as_home_address
+    select_radio(
+      question: "Is this address the same as your home address?",
+      answer: "Yes",
+    )
+  end
+
+  def answer_household_more_info_questions(js: true)
```
zspencer: __not blocking__ - This is starting to smell like we may want a page object?

spec/features/snap_application_with_maximum_info_spec.rb
jessieay: Part of me would like to, but I am hesitant to check for data in feature specs (prefer to keep them at the level of what a person would see). We do have pretty decent controller specs coverage for things like data. What do you think, how do you usually handle things like this?

spec/support/snap_application_form_helper.rb
jessieay: good idea! I think this PR is a good first step in refactoring our feature specs because they were getting a bit difficult to read in that one big file, but I can definitely see us doing that in the near future! adding a trello card

spec/features/snap_application_with_maximum_info_spec.rb
zspencer: Yeah, I think that's valid. I think it's gonna be really hard to test what someone sees when it comes to the PDF; which is probably OK.

PR #207, opened 2017-09-11
Add office-specific landing pages for Union and Clio

jessieay:
* Starting an app from either of these pages sets the "office location"
attribute, which receives highest priority when deciding who to fax.

screen shot 2017-09-11 at 2 23 18 pm

### DIFF
--- a/app/controllers/introduce_yourself_controller.rb
+++ b/app/controllers/introduce_yourself_controller.rb
@@ -6,7 +6,7 @@ class IntroduceYourselfController < StandardStepsController
 
     if @step.valid?
       app = current_or_new_snap_application
-      app.save!
+      app.update!(office_location: step_params[:office_location])
       set_current_snap_application(app)
       member.update!(member_update_params)
       redirect_to(next_path)
@@ -22,11 +22,12 @@ class IntroduceYourselfController < StandardStepsController
   end
 
   def existing_attributes
-    HashWithIndifferentAccess.new(member.attributes)
+    HashWithIndifferentAccess.new(member.attributes).
+      merge(office_location_params)
   end
 
   def member_update_params
-    step_params
+    step_params.except(:office_location)
   end
 
   def member
@@ -36,5 +37,9 @@ class IntroduceYourselfController < StandardStepsController
 
   def current_or_new_snap_application
     current_snap_application || SnapApplication.new
+  end
+
+  def office_location_params
+    { office_location: params[:office_location] }
   end
 end
--- a/app/models/fax_recipient.rb
+++ b/app/models/fax_recipient.rb
@@ -1,12 +1,17 @@
 class FaxRecipient
   attr_accessor :residential_address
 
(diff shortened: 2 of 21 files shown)


jessieay:
[As a client, I want to visit an "Office Homepage" so that I can be seen immediately](https://trello.com/c/8KhTAItA/248-as-a-client-i-want-to-visit-an-office-homepage-so-that-i-can-be-seen-immediately)

jessieay:
["Homepages" for Clio & Union Street](https://trello.com/c/qg3JUhyR/258-homepages-for-clio-union-street)

app/models/fax_recipient.rb
```diff
@@ -1,12 +1,17 @@
...
   attr_accessor :residential_address
 
-  def initialize(residential_address:)
+  def initialize(residential_address:, office_location: "")
     @residential_address = residential_address
+    @office_location = office_location
   end
 
   def number
-    self.class.find_by(zip: residential_address.zip)["fax_number"]
```
zspencer: __not blocking__ - I feel like we're *possibly* reaching the inflection point where making this take the snap application itself isn't unreasonable; as well as composing it into the snap application as `snap_appliction.fax_recipient`

app/controllers/introduce_yourself_controller.rb
```diff
@@ -6,7 +6,7 @@ def update
 
     if @step.valid?
       app = current_or_new_snap_application
-      app.save!
+      app.update!(office_location: step_params[:office_location])
```
jessieay: I guess this could be `app.update!(office_location_params)`

app/views/pages/union.html.erb
```diff
@@ -0,0 +1,73 @@
...
+    </div>
+
+    <div class="grid__item width-one-fourth shift-one-sixth trust-badge">
+      <a href="http://www.michigan.gov/mdhhs/" class="illustration illustration--mi_mdhhs">MDHHS</a>
+      <p class="text--small">A registered Community Partner and digital application assister.</p>
+    </div>
+  </div>
+</section>
+
+<%= render 'shared/footer' %>
```
zspencer: __maybe blocking__ - Since this is the same as index, with only super minor differences, does it make sense to merge them somehow?

app/models/fax_recipient.rb
jessieay: good idea! updated at 831cbb5

app/views/pages/union.html.erb
jessieay: I am hesitant to remove all duplication because I think that makes these view templates harder to read and update, but I extracted the code that is exactly the same into a shared partial here: d2aa436

jessieay:
@zspencer thanks for the feedback!

I removed some of the duplication. This landing page stuff is specific to faxing, which will hopefully be our primary means of submitting apps for not that much longer. If it is, we can move to *completely* eliminate the view template duplication

PR #209, opened 2017-09-11
Fix rake tasks that got overlooked in refactor passes

zspencer:
*Why* Because the 30m enqueue job is currently not working.

See https://trello.com/c/0uVGs3b3/275-the-rake-task-that-queues-up-unsent-faxes-should-not-fail

Sentry: https://sentry.io/app73768247herokucom-6k/app73768247herokucom/issues/347534988/?referrer=slack

### DIFF
--- a/lib/tasks/one_offs.rake
+++ b/lib/tasks/one_offs.rake
@@ -8,8 +8,7 @@ namespace :one_offs do
   def migrate_faxed_at_to_exports
     applications = SnapApplication.where.not(faxed_at: nil)
     applications.find_each do |application|
-      recipient = FaxRecipient.new(residential_address:
-                                   application.residential_address)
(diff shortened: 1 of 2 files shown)


zspencer:
[The rake task that queues up unsent faxes should not fail.](https://trello.com/c/0uVGs3b3/275-the-rake-task-that-queues-up-unsent-faxes-should-not-fail)

jessieay:
looks like we need some tests for these!

zspencer:
@jessieay - The one off is going to be deleted tomorrow; and I'm not really sure how to test single line rake tasks?

jessieay:
I was just thinking something that exercises it so that a test run would catch this rather than a deploy failure -- but if we are deleting tomorrow, no need to do it now! ;0

PR #211, opened 2017-09-12
Fill out more driving pages and abstract some common patterns

jayroh:
Quick-start to trying this
-------------------------

```
$ bundle exec rake db:seed
Creating minimal snap application...
Minimal application created (or found) with id: 3
Creating a more complete snap application...
More complete application created (or found) with id: 4
```

Take note of the "minimal application" id (3) as it will be used when you test the driving.

```
$ bundle exec rails runner "DriverApplication.destroy_all"
$ WEB_DRIVER=chrome DEBUG_DRIVE=true ./bin/rails runner "MiBridges::Driver.new(snap_application: SnapApplication.find(3)).run"
```

Destroying all the `driver_applications` records ensures you are creating a brand new MIBridges application instead of signing in with an existing one.

Abstract the conditional checking of boxes
==========================================

The questions on these pages are constructed in a slightly confusing manner:

The best way to scope each question is to hook into an aria attribute and check
the boxes by executing javacript against a selector that works thanks to
jQuery. That is how `check_in_section` works. For a better example -- and
please note that `aria-labelledby-section-name` is the value for the html
property `aria-labelledby`:

```ruby
check_in_section("aria-labelledby-section-name") # checks "No one"
check_in_section("aria-labelledby-section-name", if: predicate?, name: "John") # checks the box next to "John" if the predicate method returns
true
```

***

Add common "Click Next Page" super class
========================================

There are more than a fair share of "review" or "summary" pages on MI Bridges
that need only to have a "Next" button clicked. To represent the behavioral
needs of these pages a super class in the form of a "ClickNextPage" class is
made in this commit for other classes to inherit from.

This allows us to

1. Change the behavior in only one place if need be
2. Give the inheriting classes their own descriptive class name so viewing the
   queue of pages in `driver.rb` self-documents.

***

Change "Never married" option to "Never Married"
================================================

This small capitalization is necessary to match the options provided in the
Michigan Bridges application.

### DIFF
--- a/app/views/personal_detail/edit.html.erb
+++ b/app/views/personal_detail/edit.html.erb
@@ -16,7 +16,7 @@
           layout: "inline" %>
         <%= f.mb_select :marital_status,
         'What is your marital status?',
-        ['Married', 'Never Married', 'Divorced', 'Widowed', 'Separated'],
+        ['Married', 'Never married', 'Divorced', 'Widowed', 'Separated'],
         include_blank: "Choose one" %>
       <%= f.mb_input_field :ssn,
       'What is your social security number?',
--- a/app/views/personal_detail/edit.html.erb
+++ b/app/views/personal_detail/edit.html.erb
@@ -16,7 +16,7 @@
           layout: "inline" %>
         <%= f.mb_select :marital_status,
         'What is your marital status?',
-        ['Married', 'Never married', 'Divorced', 'Widowed', 'Separated'],
+        ['Married', 'Never Married', 'Divorced', 'Widowed', 'Separated'],
         include_blank: "Choose one" %>
       <%= f.mb_input_field :ssn,
       'What is your social security number?',
--- a/lib/mi_bridges/driver/base_page.rb
+++ b/lib/mi_bridges/driver/base_page.rb
@@ -33,7 +33,25 @@ module MiBridges
 
       def click_id(id)
         id = "##{id}" unless id.include?("#")
-        page.execute_script("$('#{id}').click()")
+        js_click_selector(id)
+      end
+
+      def js_click_selector(selector)
+        page.execute_script %($("#{selector}").click())
+      end
+
+      def check_in_section(section, options = {})
+        predicate = options.fetch(:if, false)
+        name = options.fetch(:for, "")
+
+        widget = "[aria-labelledby='#{section}']"
+        selector = if predicate
+                     "#{widget} label:contains('#{name}') input"
+                   else
+                     "#{widget} label:contains('No one') input"
+                   end
+
+        js_click_selector(selector)
       end
 
       def next
--- a/lib/mi_bridges/driver/base_page.rb
+++ b/lib/mi_bridges/driver/base_page.rb
@@ -44,14 +44,18 @@ module MiBridges
         predicate = options.fetch(:if, false)
         name = options.fetch(:for, "")
 
-        widget = "[aria-labelledby='#{section}']"
         selector = if predicate
-                     "#{widget} label:contains('#{name}') input"
+                     selector_for_radio(name)
                    else
-                     "#{widget} label:contains('No one') input"
+                     selector_for_radio("No one")
                    end
 
-        js_click_selector(selector)
+        widget = "[aria-labelledby='#{section}']"
+        js_click_selector("#{widget} #{selector}")
+      end
+
+      def selector_for_radio(name)
+        "label:contains('#{name}') input"
       end
 
       def next
--- a/lib/mi_bridges/driver/household_member_questions_page.rb
+++ b/lib/mi_bridges/driver/household_member_questions_page.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module MiBridges
+  class Driver
+    class HouseholdMemberQuestionsPage < BasePage
+      delegate(
+        :anyone_disabled?,
+        :primary_member,
+        to: :snap_application,
+      )
+
+      delegate :first_name, to: :primary_member
+
+      def setup; end
+
+      def fill_in_required_fields
+        check_blindness_or_disability
+        check_drug_felonies
+        check_probation_or_parole
+        check_other_food_assistence_benefits
+        check_ssi_benefit
+        check_cash_benefit_from_mdhhs
+        check_refugee_asylee_or_victim
+      end
+
+      def continue
+        click_on "Next"
+      end
+
+      def check_blindness_or_disability
+        check_in_section(
+          "starBlindnessorDisability",
+          if: anyone_disabled?,
+          for: first_name,
+        )
+      end
+
+      def check_drug_felonies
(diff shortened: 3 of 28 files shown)


lib/mi_bridges/driver/base_page.rb
```diff
@@ -33,7 +33,25 @@ def close
...
+        js_click_selector(id)
+      end
+
+      def js_click_selector(selector)
+        page.execute_script %($("#{selector}").click())
+      end
+
+      def check_in_section(section, options = {})
+        predicate = options.fetch(:if, false)
+        name = options.fetch(:for, "")
```
jayroh: Why use an options hash instead of named parameters? Because `if` is a reserved keyword, but `if` as a hash key is 👍, 👌, no problemo.

lib/mi_bridges/driver/base_page.rb
```diff
@@ -33,7 +33,25 @@ def close
...
+
+        widget = "[aria-labelledby='#{section}']"
+        selector = if predicate
+                     "#{widget} label:contains('#{name}') input"
+                   else
+                     "#{widget} label:contains('No one') input"
+                   end
```
jayroh: on the fence whether the name for `widget` should be `widget`.

* `section`?
* `area`?
* `aria_section`?

lib/mi_bridges/driver/base_page.rb
jessieay: woah! good context

lib/mi_bridges/driver/household_member_questions_page.rb
```diff
@@ -0,0 +1,67 @@
...
+        check_refugee_asylee_or_victim
+      end
+
+      def continue
+        click_on "Next"
+      end
+
+      def check_blindness_or_disability
+        check_in_section(
+          "starBlindnessorDisability",
```
jessieay: interesting how many (or all?) of these names start with `star` -- did you ever consider moving that part of the string into the `check_in_section` method itself?

lib/mi_bridges/driver/household_member_questions_page.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      def check_other_food_assistence_benefits
+        check_in_section "starOtherFoodAssistanceBenefits"
+      end
+
+      def check_ssi_benefit
+        check_in_section(
+          "starSSIBenefit",
+          if: anyone_disabled?,
+          for: first_name,
```
jessieay: the way I am reading this is: "if anyone is disabled, check the box and enter the primary member's first name"

is that correct?

lib/mi_bridges/driver/household_member_questions_page.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      def check_other_food_assistence_benefits
+        check_in_section "starOtherFoodAssistanceBenefits"
+      end
+
+      def check_ssi_benefit
+        check_in_section(
+          "starSSIBenefit",
+          if: anyone_disabled?,
```
jessieay: I like this pattern of passing in the `if`. It's also a tad mysterious. Did you think about just wrapping this check_in_section` method call with an `if` statement  it isn't called if nobody is disabled?

lib/mi_bridges/driver/job_income_information_page.rb
```diff
@@ -0,0 +1,50 @@
...
+        check_fap_program_benefits
+      end
+
+      def continue
+        click_on "Next"
+      end
+
+      private
+
+      def check_fulltime_employment_status
```
jessieay: this method name brings up an interesting point: we ask if someone is employed, not specifically if employed FT. (not a problem for this PR, just something to think about for the app)

lib/mi_bridges/driver/job_income_information_page.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module MiBridges
+  class Driver
+    class JobIncomeInformationPage < BasePage
+      delegate :primary_member, to: :snap_application
```
jessieay: seems like we are doing this delegation and the `first_name` delegation in many steps. Thoughts on putting them in a shared place somewhere?

jayroh:
> The only thing I feel hesitant about is that lack of tests. Do you think it's too early to develop a testing strategy?

I don't think it's too, too, early to figure out the proper strategy to attack this in terms of testing. It's part of the conversation we should all have today.

However, I think the jury's still out on a lot of what we haven't experienced yet. We don't know what we don't know as far as how all of the final pieces are structured.

In a phrase - "thar be dragons", we just don't know where they are.

lib/mi_bridges/driver/job_income_information_page.rb
jayroh: Yup! For sure. It's common enough that it should be there for all the steps. Maybe prefixed as well - `primary_member_first_name`

lib/mi_bridges/driver/household_member_questions_page.rb
jayroh: "if anyone is disabled, check the box ***next to*** the primary member's first name"

app/views/personal_detail/edit.html.erb
```diff
@@ -16,7 +16,7 @@
           layout: "inline" %>
         <%= f.mb_select :marital_status,
         'What is your marital status?',
-        ['Married', 'Never married', 'Divorced', 'Widowed', 'Separated'],
+        ['Married', 'Never Married', 'Divorced', 'Widowed', 'Separated'],
```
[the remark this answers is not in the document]
jayroh: It's in this commit: https://github.com/codeforamerica/michigan-benefits/pull/211/commits/12d53206e1bb6462b49889c775f704717a2ab8d3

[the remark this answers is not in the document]
jessieay: cool, this still breaks the PDF stuff though so let's make sure we fix that:

see `app/models/snap_application_member_attributes.rb`  L20

zspencer:
[Merge intermediary MIBridges Driving work](https://trello.com/c/JfDlPxhe/285-merge-intermediary-mibridges-driving-work)

app/views/personal_detail/edit.html.erb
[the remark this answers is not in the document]
zspencer: I used `titleize` in the translation to mi bridges so that we aren't mixing the inputs we expect with the inputs michigan bridges expects

zspencer:
@jayroh - I tried dismissing these binding.pry exceptions on codeclimate, but I don't think the syncronization of my github user with codeclimate permissions allow the dismissal.

I *don't think* that we should remove these binding.pry as they are only used when `DEBUG_DRIVER=true` and they are super useful for prototyping and interacting with the page as we build out a solution.

jayroh:
@zspencer rebased and merged - good stuff! Love the debugger page

PR #213, opened 2017-09-12
Fix seeds

jessieay:
* They were referencing a column that no longer exists


jessieay:
should we make these idempotent?

adarsh:
LGTM! For this reason, I've seen FactoryGirl usage in seeds be a little less brittle but that's more work than you've signed up for here. 🥇

jessieay:
@jayroh nice!

PR #217, opened 2017-09-12
Admin stats cover all applications

zspencer:
Reason for Change
===================
* Previously they were only the current page
 * https://trello.com/c/Z2pv4On3/273-admin-signed-stats-are-incorrect

Changes
=======
* Use a percentage of total instead of percentage of previous  cohort

### DIFF
--- a/app/views/admin/snap_applications/index.html.erb
+++ b/app/views/admin/snap_applications/index.html.erb
@@ -50,10 +50,11 @@ It renders the `_table` partial to display details about the resources.
 </header>
 
 <section class="main-content__funnel">
-  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: [], cohort: resources, name: "Created") %>
-  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: resources, cohort: resources.signed, name: "Signed") %>
-  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: resources.signed, cohort: resources.faxed, name: "Faxed") %>
-  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: resources.faxed, cohort: resources.emailed, name: "Emailed") %>
+  <%- all_resources = resources.unscoped %>
+  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: [], cohort: all_resources, name: "Created") %>
+  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: all_resources, cohort: all_resources.signed, name: "Signed") %>
+  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: all_resources, cohort: all_resources.faxed, name: "Faxed") %>
+  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: all_resources, cohort: all_resources.emailed, name: "Emailed") %>
 </section>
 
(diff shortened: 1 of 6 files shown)


zspencer:
[Admin: Signed stats are incorrect](https://trello.com/c/Z2pv4On3/273-admin-signed-stats-are-incorrect)

app/views/admin/snap_applications/index.html.erb
```diff
@@ -50,10 +50,11 @@ It renders the `_table` partial to display details about the resources.
 </header>
 
 <section class="main-content__funnel">
-  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: [], cohort: resources, name: "Created") %>
-  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: resources, cohort: resources.signed, name: "Signed") %>
-  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: resources.signed, cohort: resources.faxed, name: "Faxed") %>
-  <%= render "funnel_stage", stage: FunnelStage.new(previous_cohort: resources.faxed, cohort: resources.emailed, name: "Emailed") %>
+  <%- all_resources = resources.unscoped %>
```
zspencer: administrate provides the `resources` value as a local, not as a helpper method, so we can't write helper methods to intercept it :(.

spec/features/admin_dashboard_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+RSpec.feature "Submit application with minimal information" do
+  scenario "The stats are accurate", javascript: true do
+    page.driver.browser.current_session.basic_authorize("admin", "password")
+    5.times { create(:snap_application, signed_at: nil) }
```
jessieay: can also do this as `create_list(:snap_application, 5, signed_at: nil)`

spec/features/admin_dashboard_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    5.times { create(:snap_application, :emailed, signed_at: nil) }
+    10.times { create(:snap_application, signed_at: 5.days.ago) }
+    5.times { create(:snap_application, :faxed, signed_at: 5.days.ago) }
+    visit admin_root_path
+    expect(page).to have_content "25 Created"
+    expect(page).to have_content "15 Signed"
+    expect(page).to have_content "5 Emailed"
+    expect(page).to have_content "5 Faxed"
```
jessieay: since creating records slows down the specs, what do you think is the minimal number of records we could create here to confirm that the numbers are accurate?

zspencer: Yea, I wanted to get over the 20 boundary that would cause pagination to kick in, but 21 should be fine instead of 25.

jessieay: in that case, what I usually do is stub the pagination numbers for the tests so I don't need quite as many records

zspencer: ooooh that's a great idea.... I should have thought of that before merge. This test takes ~1s on my machine, at some point we could do a refactor pass to optimize the test suite in a later ticket?

jessieay: yes of course! wanted to get this line of thinking on our radar bc minimal record creation is an easy win for a fast test suite, but no need to change anything now

PR #221, opened 2017-09-12
Send SMS after application is submitted

jessieay:
* Open question: do we record that an SMS was sent?


jessieay:

app/controllers/success_controller.rb
```diff
@@ -12,14 +12,22 @@ def next_path
...
-    Export.create_and_enqueue(destination: :fax,
-                              snap_application: current_snap_application)
+    Export.create_and_enqueue(
+      destination: :fax,
+      snap_application: current_snap_application,
+    )
+
+    ApplicationSubmittedSmsJob.perform_later(
+      snap_application: current_snap_application,
+    )
```
zspencer: If we add an `:sms` destination in the `Export` model, it will automatically track when and how we send SMSs.

jessieay: yeah I considered doing that, but felt like `export` was specific to exporting a PDF, conceptually, rather than "sending out a thing" -- what do you think?

zspencer: Maybe it means we need a rename to `Communique`? It's more about tracking when and where we communicate outwards (either via MiBridges, Fax, Email, SMS, etc) so that an admin can drill into the time we send stuff?

jessieay:
@zspencer updated at https://github.com/codeforamerica/michigan-benefits/commit/13e46cf9de232da6178e8eba279228cadc394862 with your feedback. please take a look!

PR #223, opened 2017-09-12
Remove default checkboxes from PDF

jessieay:
**WHY**: these checkboxes were checked by default before since, if there
were a value, we'd always know that it was for a specific interval. But,
this default behavior was confusing MDHHS workers because it would show
up even when the value was nil. Eg: no court ordered expenses, but still
a checkbox next to "monthly" in the court ordered expenses category.


jessieay:
[As an MDHHS worker, I want there to be _no marking_ on the 1171 when a client answers "no" to a question](https://trello.com/c/dBkLSa91/250-as-an-mdhhs-worker-i-want-there-to-be-no-marking-on-the-1171-when-a-client-answers-no-to-a-question)

PR #224, opened 2017-09-13
Add ghostscript to setup script

jayroh:
`gs` is a dependency of our test suite. After running into a
**command `gs` not found** error I reckon this should be added to the
setup script for future developers.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -29,6 +29,14 @@ chdir APP_ROOT do
     end
   end
 
+  step "Installing ghostscript, an interpreter for the PostScript language and for PDF" do
+    if cli_installed?("gs")
+      puts "ghostscript already installed."
+    else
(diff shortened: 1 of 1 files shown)


bin/setup
```diff
@@ -29,6 +29,14 @@ chdir APP_ROOT do
     end
   end
 
+  step "Installing ghostscript, an interpreter for the PostScript language and for PDF" do
+    if cli_installed?("gs")
```
jessieay: did you test this with gs already installed? I have it on my machine and when I run `gs` it opens a ghostscript prompt, which I imagine might make this hang?

jayroh: I did not. But now that you mention it, that's something I saw as well. I've updated with this commit: f3b351447caec0c9c4d87c65e13d7e5d491f61ed

PR #226, opened 2017-09-13
Rename args for "check_in_section" method

jessieay:
* Now that we are no longer using if, we can use named args

### DIFF
--- a/lib/mi_bridges/driver/housing_utility_bills_page.rb
+++ b/lib/mi_bridges/driver/housing_utility_bills_page.rb
@@ -26,11 +26,19 @@ module MiBridges
       private
 
       def check_housing_bills
-        check_in_section "starHousingBills", if: true, for: first_name
+        check_in_section(
+          "starHousingBills",
+          condition: true,
+          for_name: first_name,
+        )
       end
 
       def check_utility_bills
-        check_in_section "starUtilityBills", if: true, for: first_name
+        check_in_section(
+          "starUtilityBills",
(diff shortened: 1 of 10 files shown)


jessieay:
[Rename "if" to "condition" in the mi bridges exporter](https://trello.com/c/z7Yu3qWP/277-rename-if-to-condition-in-the-mi-bridges-exporter)

lib/mi_bridges/driver/housing_utility_bills_page.rb
```diff
@@ -26,20 +26,28 @@ def continue
...
 
       def check_room_and_meals
         check_in_section "starRoomandMeals"
       end
 
       def select_home_heating_payment
-        check_in_section "heatingResponseGroup", if: true, for: "No"
-        check_in_section "meapReceivedGroup", if: true, for: "No"
+        check_in_section "heatingResponseGroup", condition: true, for_name: "No"
+        check_in_section "meapReceivedGroup", condition: true, for_name: "No"
```
jayroh: *Most* of the time the check box is for a name, but sometimes it's not. Sometime's it's something like "No" or "None".  Is there something else that might be better? Maybe `for_label`?

jessieay: Great point! Renamed at https://github.com/codeforamerica/michigan-benefits/commit/d504e3716a1dc1dc09a5a669ecd99b753ec6db05

PR #228, opened 2017-09-13
Do not display newly uploaded file post-upload

jayroh:
As a step towards locking down access to files uploaded to S3 we must no
longer display the successful image URL after the file is uploaded.
Instead, let's just keep the preview element there and let the JS show the
contents of the file that way.

Also pull in the work Rachel and Zee did to display upload errors inline
within the preview element. Including:

* Allow files up to 8Mb in size
* Don't show the warning by default
* Add error message for small photo sizes

### DIFF
--- a/app/assets/javascripts/upload_setup.js
+++ b/app/assets/javascripts/upload_setup.js
@@ -1,4 +1,4 @@
-/*global uploadSetup */
+/*global $,uploadSetup,Shubox */
 
 var uploadSetup = function (previewTemplate) {
   new Shubox('#document-drop', {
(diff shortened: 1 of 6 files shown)


jayroh:

jayroh:
> When I pulled down this branch, I could still see the s3 URL when inspecting element.

Could you post a screenshot of the DOM inspector where you see the S3 url, @jessieay?

jessieay:
@jayroh sure thing!

screen shot 2017-09-13 at 1 40 38 pm

as far as I can tell, we did not expose it on the UI layer before (when testing on staging, I did not see it there). where in the uI was it?

jayroh:
So here's the thing.

Previously we were displaying the full image post-success in the background of a div (in order to use the `background-size: cover` style magic). So - we *were* making a get request to this image.

As of this PR, we will not. We are now **saving the URL to the image**, not requesting it.  In the future when we lock down permissions of these files in the S3 bucket, we *will* be able to know the URL to the image. We will *not* be able to request that image.

So, back to your statement:

> When I pulled down this branch, I could still see the s3 URL when inspecting element.

Yes. We can see the URL. We just won't be able to view it through the browser. That's what will be changing.

alanjosephwilliams:
Great context. Thank you @jayroh.

PR #235, opened 2017-09-14
Add contract test for MI Bridges driving code

jessieay:
* Run as part of deploy script

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -13,6 +13,19 @@ else
   exit 1
 fi
 
+echo "Running the contract test for MI Bridges driving code..."
+RUN_DRIVING_TEST=true bin/rspec spec/features/mi_bridges_driving_spec.rb
+
+read -p "Did it pass? (y/N) " -n 1 -r
+echo    # (optional) move to a new line
+if [[ $REPLY =~ ^[Yy]$ ]];then
+  echo "Great!"
+else
+  echo "Please make sure the driving code is working before deploying to
+  production!"
+  exit 1
+fi
+
 echo "Diffing the code between staging and production..."
 bin/release-diff
 read -p "Looks ok? (y/N) " -n 1 -r
--- a/bin/release
+++ b/bin/release
@@ -16,16 +16,6 @@ fi
 echo "Running the contract test for MI Bridges driving code..."
 PRE_DEPLOY_TEST=true bin/rspec spec/features/mi_bridges_driving_spec.rb
 
-read -p "Did it pass? (y/N) " -n 1 -r
-echo    # (optional) move to a new line
-if [[ $REPLY =~ ^[Yy]$ ]];then
-  echo "Great!"
-else
-  echo "Please make sure the driving code is working before deploying to
-  production!"
-  exit 1
-fi
-
 echo "Diffing the code between staging and production..."
 bin/release-diff
 read -p "Looks ok? (y/N) " -n 1 -r
--- a/bin/release
+++ b/bin/release
@@ -14,7 +14,7 @@ else
 fi
 
 echo "Running the contract test for MI Bridges driving code..."
-RUN_DRIVING_TEST=true bin/rspec spec/features/mi_bridges_driving_spec.rb
+PRE_DEPLOY_TEST=true bin/rspec spec/features/mi_bridges_driving_spec.rb
 
 read -p "Did it pass? (y/N) " -n 1 -r
 echo    # (optional) move to a new line
--- a/lib/mi_bridges/driver.rb
+++ b/lib/mi_bridges/driver.rb
@@ -94,7 +94,9 @@ module MiBridges
     end
 
     def teardown
-      MiBridges::Driver::BasePage.new(snap_application).close
+      if Rails.env != "test"
+        MiBridges::Driver::BasePage.new(snap_application).close
+      end
     end
 
     def first_driver_application_attempt?
--- a/lib/mi_bridges/driver.rb
+++ b/lib/mi_bridges/driver.rb
@@ -94,7 +94,7 @@ module MiBridges
     end
 
     def teardown
-      if Rails.env != "test"
+      if ENV["PRE_DEPLOY_TEST"] != "true"
         MiBridges::Driver::BasePage.new(snap_application).close
       end
     end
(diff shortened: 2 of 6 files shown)


jessieay:
[Write "contract test" that we can run pre-deploy that runs through the most fullest application possibly but stops short of submission via mi bridges](https://trello.com/c/6m8hOiI6/280-write-contract-test-that-we-can-run-pre-deploy-that-runs-through-the-most-fullest-application-possibly-but-stops-short-of-submis)

bin/release
```diff
@@ -13,6 +13,19 @@ else
   exit 1
 fi
 
+echo "Running the contract test for MI Bridges driving code..."
+RUN_DRIVING_TEST=true bin/rspec spec/features/mi_bridges_driving_spec.rb
```
zspencer: Because we have `set -e` on line 2, I believe if the test fails it will cause the script to stop by default.

lib/mi_bridges/driver.rb
```diff
@@ -94,7 +94,9 @@ def setup
     end
 
     def teardown
-      MiBridges::Driver::BasePage.new(snap_application).close
+      if Rails.env != "test"
```
zspencer: Is this actually about the environment, or is it about whether we want to be debugging the application?

spec/factories/member.rb
```diff
@@ -4,7 +4,7 @@
     last_name "Pad"
     sex "female"
     marital_status "Widowed"
-    ssn "123 12 1234"
+    ssn "123121234"
```
zspencer: _not blocking_ - Since this data format has meaning, do we want to do something in the driver layer that removes the ` `s?

lib/mi_bridges/driver.rb
jessieay: in this case, it was so the test would run. I think the Debugger page stops things if we want to view the state of affairs otherwise? Or do you think this should depend on an env var rather than environment?

bin/release
jessieay: ok cool, so should I get rid of the following section?

spec/spec_helper.rb
```diff
@@ -35,6 +35,13 @@
     "responders"
 
   config.filter_run :focus
+
+  if ENV["RUN_DRIVING_TEST"] == "true"
```
zspencer: _not blocking_ - Ahh this runs *only* the driving test. Maybe a rename could work to make it more about the intent (pre-deploy testing) than the implementation?

lib/mi_bridges/driver.rb
zspencer: Sorry I always forget to do the "review" mode. Yeah, I think making it rely on the same ENV variable as the other stuff that does the pre-deploy would make sense?

bin/release
zspencer: I think it's probably fine unless we want to have the browser stay open for a bit so they can do some manual checking?

zspencer: I just had another thought, apologies... should we be running this on the staging environment via `heroku run` so we're testing we have the right buildpacks and gems in the right location?

jessieay: yeah I thought about that. If we go that direction, we will also need to include all of our test-env-specific gems, but I do think it would be good. 🤔

zspencer: Hmmmm, what if we pulled this out of rspec and into a plain old ruby script that lived in `bin` like `bin/test_mi_bridges_driver`so we don't need to pull in all the test gems?

zspencer: @jessieay yeah... This be complicated. I don't know if the additional work to make it run on Heroku is worth the investment?

zspencer: SOme thoughts on how:

1. Emphasize a driver design of failing fast in places of uncertainty?
2. We could build in checks that verify the data we entered is on the page as part of the standard page actions?
3. ???

lib/mi_bridges/driver.rb
jessieay: >  I think making it rely on the same ENV variable as the other stuff that does the pre-deploy would make sense?

What do you mean by "other stuff" in that sentence? Is that a separate ENV var that releasing depends on? I looked around but didn't see anything.

spec/factories/member.rb
jessieay: Just realized that we include the `SocialSecurityNumber` concern  (which validates the format) in the relevant step classes but not the `Member` AR class. added a ticket for that here: https://trello.com/c/t4tYuREg/310-validate-format-of-ssn-on-member-class

jessieay: reverted factory change here since it was not relevant to this PR 25a6208

spec/spec_helper.rb
jessieay: gooood call - updated name at 5388bce

bin/release
jessieay: I think 2 is a good idea as we move forward with driving code -- on each step, we should know where we are based on the page title / content

lib/mi_bridges/driver.rb
jessieay: ah wait, now I get what you were saying! Yes, totally makes sense to use `ENV["PRE_DEPLOY_TEST"]` here, for consistency. done at 7469836

bin/release
jayroh: I know the proposal is to run it on *staging* but that, for some reason, makes me nervous to make it possible to run anything like this on heroku at all.

spec/features/mi_bridges_driving_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
+
+RSpec.feature "MI Bridges Driving" do
+  scenario "successfully drives a SNAP Application", :js, :driving do
```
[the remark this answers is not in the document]
jayroh: Might be worthwhile to add a comment here that tells you how to enable running this using the full chrome browser?

When it fails I imagine that's the first thing one would want to do is see this run on a browser. Thoughts?

bin/release
zspencer: @jayroh - Aren't we supposed to be running all the driving stuff on Heroku?

jessieay:
@jayroh good idea, added a comment at 572cc0d

jayroh:
@jessieay if you remove the :js it'll run with the full Chrome browser?

PR #236, opened 2017-09-14
Make it easier to drive a given application

zspencer:
Reason for Change
===================
* I kept having to look up docs on how to run the driver against a particular app, so I scripted it?

Minor
=====
* I got rid of that `destroy_all` in the Driver setup, since I don't think we actually want to be destroying all?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -228,13 +228,8 @@ should get this set up for you automatically.
 
 ### How to Drive Applications
 
-To run the driver code, use `rails console` and (where
-`<id>` is the id of a snap application record) run:
-
-```ruby
-snap_application = SnapApplication.find(<id>)
-MiBridges::Driver.new(snap_application: snap_application).run
-```
+To run the driver code, use `bin/drive <id>` (where
+`<id>` is the id of a snap application record):
 
 There are three extra environment variables you can set to modify the behavior
 of the script.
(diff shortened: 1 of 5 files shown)


lib/mi_bridges/driver.rb
```diff
@@ -89,7 +89,6 @@ def run
     attr_reader :snap_application
 
     def setup
-      DriverApplication.destroy_all
```
zspencer: @jayroh / @jessieay - This in particular worried me a bit; I'm assumin we only want to be deleting driver applications in particular cases, not every time we drive?

jessieay: you need to have this line if you are going to re-drive for the same snap app multiple times (which is what we are doing now).

We will want to remove for prod, though, so maybe put it in an env check in the meantime? or remove scoped to the id of the current snap app?

CONTRIBUTING.md
```diff
@@ -249,6 +244,6 @@ Example:
 
 ```sh
 DRIVER_SPEED=slow \
-             WEB_DRIVER=chrome \
-             DEBUG_DRIVE=true \
-             ./bin/rails runner "MiBridges::Driver.new(snap_application: SnapApplication.find(ID)).run"
+  WEB_DRIVER=chrome \
```
jessieay: you know what else would be cool? to have a `bin/demo_drive` script that does this for us!

lib/mi_bridges/driver.rb
jayroh: This *could* remain if we scope it to Rails.env.development only - because (at least for NOW) we only want to run the driver for brand new applications. Otherwise this would pick up previous runs.

zspencer: I moved it into the `bin/drive` script; but I like the notion of renaming `bin/drive` to `bin/dev_drive` and `bin/demo_drive` with the default environment variables set for that.

PR #237, opened 2017-09-14
Add SNAP explanation and Michigan Benefits section

rachelcope:
Reason for Change
===================
* Separate SNAP explanation from information about MichiganBenefits.
* Add info sectino on MichiganBenefits and privacy concerns
* https://trello.com/c/fO8I930e/225-home-as-a-user-i-want-a-clear-explanation-of-what-snap-is-and-how-this-site-helps

fireshot capture 39 - michigan benefits - http___localhost_3000_

### DIFF
--- a/app/views/pages/_home_page_bottom_section.html.erb
+++ b/app/views/pages/_home_page_bottom_section.html.erb
@@ -1,26 +1,51 @@
-<section class="slab" id="scroller-target">
+<section class="slab slab--white slab--padded" id="scroller-target">
   <div class="grid">
     <div class="grid__item width-seven-twelfths">
       <p>
-        MichiganBenefits.org is an easy, safe and secure way to apply for
-        Michigan's Food Assistance Program. Information you provide will be
-        securely submitted as an official application with the Michigan
-        Department of Health and Human Services.
-
-        MichiganBenefits.org is a non-profit that offers online application
-        assistance to any Michigan resident. For additional assistance email our
-        team at <a href="mailto:hello@michiganbenefits.org">hello@michiganbenefits.org</a>.
-      </p>
-
-      <p>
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/atoms/_layout.scss
```diff
@@ -35,6 +35,10 @@
   }
 }
 
+.slab--centered {
+  text-align: center;
```
jessieay: there is a `text--centered` class that does this as well. would the atomic way of doing this be to apply that class to the div as well and not create another class?

app/views/pages/_home_page_bottom_section.html.erb
```diff
@@ -24,3 +13,40 @@
...
+
+        <hr>
+
+        <p>
+          Private and Secure
+        </p>
+        <p class="text--help">
+          Your information is kept private and secure. MichiganBenefits
+          maintains strict security guidelines to protect your privacy and indentity.
+          For details, read the full <a href="/privacy" target="_blank">Privacy Policy</a>.
```
jessieay: `For details, read the full <%= link_to "Privacy Policy", privacy_path, target: "_blank" %>.`

app/assets/stylesheets/atoms/_layout.scss
rachelcope: @jessieay yes, I think it makes sense to use the text--centered class. I was referring to the GetCalFresh website that has a `slab--centered` class. But I like reusing the class that we already have. Will update

app/views/pages/_home_page_bottom_section.html.erb
rachelcope: 👍

PR #239, opened 2017-09-14
Expose the office name in the dashboard

zspencer:
Reason for Change
===================
* When working on another ticket, I had noticed that we said "Michigan Bridges" as the recipient which is... annoying at best and misleading at worst. I wanted to make sure we could more easily answer the question of "Which office did the fax go to?" and "Which office did should the fax go to?"

##### Before
screen shot 2017-09-14 at 1 18 57 pm

##### After
screen shot 2017-09-14 at 1 32 50 pm

### DIFF
--- a/app/models/fax_recipient.rb
+++ b/app/models/fax_recipient.rb
@@ -12,12 +12,23 @@ class FaxRecipient
   end
 
   def name
-    ENV.fetch("FAX_RECIPIENT", "Michigan Bridges")
+    office["name"].to_s.titleize
+  end
+
+  def office
+    if office_location.present?
+      office = self.class.offices[office_location]
+      office["name"] = office_location
+      office
+    else
+      self.class.find_by(zip: residential_address.zip)
+    end
   end
 
   def self.find_by(zip:)
     name = covered_zip_codes[zip] || configuration["fallback_office"]
     Rails.logger.debug("Mapped zip #{zip} to office #{name}")
+    offices[name]["name"] = name
     offices[name]
   end
 
--- a/app/models/fax_recipient.rb
+++ b/app/models/fax_recipient.rb
@@ -4,31 +4,24 @@ class FaxRecipient
   end
 
   def number
-    if office_location.present?
-      self.class.offices[office_location]["fax_number"]
-    else
-      self.class.find_by(zip: residential_address.zip)["fax_number"]
-    end
+    office["fax_number"]
   end
 
   def name
-    office["name"].to_s.titleize
+    office["name"]
   end
 
   def office
     if office_location.present?
-      office = self.class.offices[office_location]
-      office["name"] = office_location
-      office
+      self.class.offices[office_location]
     else
       self.class.find_by(zip: residential_address.zip)
     end
   end
 
   def self.find_by(zip:)
-    name = covered_zip_codes[zip] || configuration["fallback_office"]
+    name = covered_zip_codes[zip.to_s] || configuration["fallback_office"]
     Rails.logger.debug("Mapped zip #{zip} to office #{name}")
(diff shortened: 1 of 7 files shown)


app/models/fax_recipient.rb
```diff
@@ -12,12 +12,23 @@ def number
   end
 
   def name
-    ENV.fetch("FAX_RECIPIENT", "Michigan Bridges")
+    office["name"].to_s.titleize
+  end
+
+  def office
```
jessieay: seems like this could use a unit test

app/models/fax_recipient.rb
```diff
@@ -12,12 +12,23 @@ def number
...
+      office
+    else
+      self.class.find_by(zip: residential_address.zip)
+    end
   end
 
   def self.find_by(zip:)
     name = covered_zip_codes[zip] || configuration["fallback_office"]
     Rails.logger.debug("Mapped zip #{zip} to office #{name}")
+    offices[name]["name"] = name
```
jessieay: this feels a bit confusing -- why this change?

alanjosephwilliams:
great catch and significant mprovement

app/models/snap_application.rb
```diff
@@ -65,6 +65,18 @@ def residential_address
     addresses.where.not(mailing: true).first || NullAddress.new
   end
 
+  def zip
+    residential_address.zip
+  end
+
+  def receiving_office
```
jessieay: thoughts on putting this below `receiving_office_name` so that the methods in this file are in dependence order? (methods are above other methods that they depend on). This is something I picked up from Avdi Grimm's ruby tapas!

app/models/fax_recipient.rb
zspencer: This was hella annoying. It's because our yml file doesn't include the name of the office in the data, just as the key. It would be better to do as part of processing the YML file initially, not this far downstream.

app/models/snap_application.rb
zspencer: Nice call! I wish there was a rubocop for that!

app/models/fax_recipient.rb
jessieay: interesting -- so is there a test we can write that confirms that change does what we want?

zspencer:
@jessieay  - Resolved your feedback, I think?

config/offices.yml
```diff
@@ -2,7 +2,7 @@
 defaults: &defaults
   # Which office to send the fax to in the event the submitted
   # zip code is not in our mapping
-  fallback_office: "union"
+  fallback_office: "clio"
```
jessieay: good change! the different receiving office fallback in different envs was confusing to me when I was working on the `office_location` stuff

spec/models/fax_recipient_spec.rb
```diff
@@ -1,135 +1,77 @@
...
+        snap_application: app(from_zip_covered_by: :testing),
+      )
 
-        expect(fax_recipient.number).to eq clio_fax_number
-      end
+      expect(fax_recipient.name).to eq fax_recipient.office["name"]
     end
   end
-
   def sandbox_fax_number
```
jessieay: looks like we removed tests for sandbox fax number but did not re-add any?

zspencer: Yeah, I removed them because they weren't as important as the non-production ones, and forgot to clean out this method.

zspencer:
[Improve admin overview by exposing which office we faxed the 1171 to](https://trello.com/c/5jKZ5tbM/313-improve-admin-overview-by-exposing-which-office-we-faxed-the-1171-to)

PR #240, opened 2017-09-14
Downloading public *and* private S3 files

jayroh:
Soon all files being uploaded to S3 will be private. Requesting them
directly via their S3 url will usually (unless it's in development env)
not work.

As a result, we need a fallback in those cases when those files are
private. That fallback is the AWS SDK. Using the requisite credentials
we can download them that way and use them in our PDF creation workflow.

This commit extracts some of the work done previously in
`VerificationDocument` and houses it in `RemoteDocument`.

Also:

* Add dotenv gem (not sure why it wasn't included)
* Add FileMagic gem to inspect file mimetypes
* Install libmagic in `bin/setup` as it is needed for FileMagic


Gemfile
```diff
@@ -23,6 +23,7 @@ gem "prawn"
 gem "puma"
 gem "rails", "~> 5.1"
 gem "responders"
+gem "ruby-filemagic"
```
jayroh: Per the note in the commit - don't forget to run `brew install libmagic` for this one.

zspencer: Do we need to add a buildpack for that to our app.json, staging, and production environments?

bin/setup
```diff
@@ -29,6 +29,10 @@ chdir APP_ROOT do
     end
   end
 
+  step "Installing libmagic, a dependency for ruby-filemagic" do
+    system!("brew install libmagic")
```
jessieay: this is happening after `bundle install` so the gem install of `ruby-filemagic` fails when I run the setup script -- switch up order?

also, do we want to skip install if this is already installed?

app/models/verification_document.rb
```diff
@@ -5,53 +5,29 @@ def initialize(url:, snap_application:)
...
+    pdf.render_file(tempfile.path)
+    tempfile
   end
 
   private
 
   attr_reader :url, :snap_application
 
-  def create_magick_object
-    Rails.logger.debug("Creating magick object from #{local_file.path}")
```
jessieay: do we want to keep these logging statements around?

app/models/remote_document.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def initialize(url)
+    @url = url
+    @tempfile = Tempfile.new.tap(&:binmode)
+  end
+
+  def download
+    open(url) { |url_file| tempfile.write(url_file.read) }
+    self
+  rescue OpenURI::HTTPError => _e
```
jessieay: is there any way to know this is a private s3 doc without a `rescue` ? This seems to work / make sense, but it's always nice to do this kind of logic without exceptions, if possible

Gemfile
jayroh: I'll have to double-check whether this is in the default ruby buildpack but if not, I found this:

`heroku buildpacks:add --index 1 https://github.com/NetsoftHoldings/heroku-buildpack-magic.git`

Gemfile
```diff
@@ -50,6 +51,7 @@ group :development, :test do
   gem "bundler-audit"
   gem "climate_control"
   gem "codeclimate-test-reporter"
+  gem "dotenv-rails"
```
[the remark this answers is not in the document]
jayroh: If you try to access some of these envs through `Rails.application.secrets` then this does not work.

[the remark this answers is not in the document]
jessieay: in which environment? I was testing Twilio stuff accessed via `Rails.application.secrets` and I believe it worked in dev, staging, and prod.

bin/setup
jayroh: Good catch! I'll fix this up

app/models/verification_document.rb
jayroh: I don't know - do we? I don't have an opinion

app/models/remote_document.rb
jayroh: As far as I know, no. The only thing I can think of is to use the SDK to check - and that would mean another round-trip request. To me that would defeat the purpose.

Gemfile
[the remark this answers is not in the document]
zspencer: @jayroh - Foreman automatically loads the `.env` file, if you use foreman to run the app locally everything should work for yah.

app/models/remote_document.rb
zspencer: Nice fall back :). Do we intend to write a rake task to go through and update all the past uploaded documents to make them private again?

app/models/verification_document.rb
zspencer: They were put in to try and diagnose an error that... uhh.. proved to not be programmatic in nature. Feel fee to ditch them!

Gemfile
zspencer: OK, let's make sure we test this in a review app before merging it then, and then add this buildpack to `app.json` if it is necessary.

zspencer:

Gemfile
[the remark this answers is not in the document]
jayroh: @zspencer @jessieay: What if you're *not* using foreman? I often don't use foreman locally for debugging reasons -- foreman + pry don't mix well.

app/models/remote_document.rb
jayroh: I don't think we need to, @zspencer. It's a very quick and easy process to do so w/an S3 client like [Transmit]. It'd take me less than a few minutes.

[Transmit]:https://panic.com/transmit/

Gemfile
jayroh: michigan-benefits_ _pipeline___heroku

All good on that front.

spec/models/remote_document_spec.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
jayroh: Addressed inconsistency in PR #244

Gemfile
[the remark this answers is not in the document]
zspencer: Ahh, in that case you can prefix your commands with `dotenv` which will load the `.env` file. To be honest though, I don't mind having dotenv-rails as a development/test dependency because it's quite light and (I think?) brings no additional dependencies into the mix.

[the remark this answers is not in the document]
jessieay: ah ok -- we install foreman in the setup script and tell people to run foreman in the CONTRIBUTING doc so I assumed we were all using foreman. Not problematic from my side to include here, perhaps just give a tiny comment about it being useful for running server without foreman?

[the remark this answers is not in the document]
jayroh: > perhaps just give a tiny comment about it being useful for running server without foreman

@jessieay - where would you expect/like to see a comment wrt this? Because I can definitely do that.  In the commit? In CONTRIBUTING? Just want to make sure I get it right.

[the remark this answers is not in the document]
jessieay: just right here in the gemfile next to the gem works for me!

PR #241, opened 2017-09-14
Add SNAP Benefits Section to Homepage

rachelcope:
Reason for Change
===================
* Add section that explains the SNAP benefits
* https://trello.com/c/rbHdbRcN/312-home-as-a-user-i-want-to-understand-the-benefits-of-snap

fireshot capture 40 - michigan benefits - http___localhost_3000_


rachelcope:
[[Home] As a user, I want to understand the benefits of SNAP](https://trello.com/c/rbHdbRcN/312-home-as-a-user-i-want-to-understand-the-benefits-of-snap)

app/assets/stylesheets/atoms/_emoji.scss
```diff
@@ -156,4 +156,12 @@
 
 .emoji--finger-down {
   background-image: image-url('emojis/1f447.png')
-}
\ No newline at end of file
+}
+
+.emoji--broccoli {
```
jessieay: SO CUTE

app/views/pages/_home_page_bottom_section.html.erb
```diff
@@ -41,10 +75,15 @@
...
           Private and Secure
         </p>
-        <p class="text--help">
+        <p class="text--help with-padding-large">
           Your information is kept private and secure. MichiganBenefits
           maintains strict security guidelines to protect your privacy and indentity.
           For details, read the full <%= link_to "Privacy Policy", privacy_path, target: "_blank" %>.
+        </p>
+        <%= link_to step_path(StepNavigation.first), class: "button button--cta button--fully-width" do %>
+          Apply now
```
jessieay: we want this link to go to a different path depending on if it is an office page.

One way to do this would be to pass a var to the `home_page_buttom` partial in each view 

eg, on `index`, `clio` and `union` it would be these 3 (one line in each file):

```
<% render "home_page_bottom_section", office_location: "" %>
<% render "home_page_bottom_section", office_location: "clio" %>
<% render "home_page_bottom_section", office_location: "union" %>
```

and then here it would be

```
<%= link_to step_path(StepNavigation.first, office_location: office_location) class: "button button--cta button--fully-width" do %>
```

jessieay:
as for the specs that are failing, they are unhappy because there are now two links on the page with the same text "Apply now" -- to fix that, replace with

```
within(".slab--hero") do
  click_on "Apply now"
end
```

danielladevera:
A quick google verified (for me) that it's `Farmers Market` (no apostrophe) - after that copy change, I give my thumbs up!

app/views/pages/_home_page_bottom_section.html.erb
rachelcope: @jessieay That's a great point. I'll add this in!

PR #242, opened 2017-09-15
Persist debug information when exporting an application to MIBridges

zspencer:
Reason for Change
===================
* We want to make it easier for developers or admins to actually review what happened for a given export. To do so, I'm considering directing the log statements to a logger, that on a successful (or unsuccessful) export attaches the results to the database for review.

Risks
=====
* What happens if we start putting in PII like the SSN? Can we make sure not to log that?

Example Output
============
screen shot 2017-09-18 at 10 39 41 am


zspencer:
@jessieay / @jayroh - This is some spike code I wrote this morning to see if I could start persisting the log messages for a given export so that devs and/or admins can more easily review what happened with a given attempt to submit an application via the MiBridges driver.

Would love an initial response to the direction here? Especially some of @jessieay's awesome code smells.

app/models/enqueuer.rb
```diff
@@ -0,0 +1,34 @@
...
+        EmailApplicationJob.perform_later(export: export)
+      when :sms
+        ApplicationSubmittedSmsJob.perform_later(export: export)
+      else
+        raise UnknownExportTypeError, export.destination
+      end
+      export.transition_to(new_status: :queued)
+    end
+  end
+end
```
zspencer: The `Export` class is getting way too big, so I wanted to ull out the thing that actually does the enqueuing.

zspencer:
[Create a "Failed Session" bundle that is persisted so we can diagnose and resolve failed MI Bridges Exports](https://trello.com/c/jytJD7h5/276-create-a-failed-session-bundle-that-is-persisted-so-we-can-diagnose-and-resolve-failed-mi-bridges-exports)

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,12 +17,12 @@ class SnapApplicationsController < Admin::ApplicationController
...
     # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
     # for more information
-    #
     def resend_fax
       application = SnapApplication.find(params[:id])
-      Export.create_and_enqueue!(snap_application: application,
-                                 destination: :fax, force: true)
-      flash[:notice] = "Resent fax for #{application.signature}!"
+      Enqueuer.create_and_enqueue_export!(snap_application: application,
+                                          destination: :fax, force: true)
```
jayroh: Small nit I'm going to pick here.  For consistency and legibility's sake I would break these params out to their own lines. EG:

```ruby
Enqueuer.create_and_enqueue_export!(
  snap_application: application,
  destination: :fax,
  force: true,
)
```

app/models/enqueuer.rb
```diff
@@ -0,0 +1,34 @@
+class Enqueuer
+  def enqueue_faxes
+    SnapApplication.faxable.untouched_since(DELAY_THRESHOLD.minutes.ago).
+      find_each do |snap_application|
+      create_and_enqueue_export(destination: :fax,
+                                snap_application: snap_application)
+    end
+  end
```
jayroh: The indentation here makes things a little hard to follow.

app/models/enqueuer.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def create_and_enqueue_export(params)
+    enqueue(Export.create(params))
+  end
+
+  def create_and_enqueue_export!(params)
+    enqueue(Export.create(params))
+  end
```
jayroh: What's the difference between the normal and bang methods? Is it necessary to have both since the method bodies are the same?

app/models/export.rb
```diff
@@ -82,6 +59,24 @@ def execute
     snap_application.pdf.try(:unlink)
   end
 
+  def logger
+    return @_logger if @_logger.present?
+
+    logger = ActiveSupport::Logger.new(STDOUT)
+    logger.level = Logger::DEBUG
+    logger.formatter = Rails.application.config.log_formatter
+    @_logger = ActiveSupport::TaggedLogging.new(logger)
```
jayroh: A nice pattern I've seen for multi-line memoization is to wrap things in a `begin`/`end` block to wrap everything up together:

```ruby
def logger
  @_logger ||= begin
    logger = ActiveSupport::Logger.new(STDOUT)
    logger.level = Logger::DEBUG
    logger.formatter = Rails.application.config.log_formatter
    ActiveSupport::TaggedLogging.new(logger)
  end
end
```

lib/mi_bridges/driver.rb
```diff
@@ -86,9 +87,22 @@ def run
...
+                     end
+      logger.formatter = Rails.application.config.log_formatter
+      @logger = ActiveSupport::TaggedLogging.new(logger)
+    end
+
+    attr_reader :snap_application, :logger
 
     def setup
+      # Do we really want to destroy all the stored driver applications?
       DriverApplication.destroy_all
```
jayroh: What do you think about - 

`DriverApplication.destroy_all if Rails.env.development?`

\- ?

lib/mi_bridges/driver/base_page.rb
```diff
@@ -5,13 +5,14 @@ class Driver
     class BasePage
       include Capybara::DSL
 
-      def initialize(snap_application)
-        log("filling out page", self.class.to_s) if logging?
+      def initialize(snap_application, logger:)
+        @logger = logger
+        log("filling out page")
```
jayroh: Wouldn't having the page you're filling out be helpful when tracing what's going on with the driving?

app/models/enqueuer.rb
zspencer: ! is supposed to `create!`, raising an exception in the event the export isn't valid.

lib/mi_bridges/driver.rb
zspencer: I think this is left over from before we removed it.

lib/mi_bridges/driver/base_page.rb
zspencer: the way this works is using tagged logging, so `log` messages will be prefixed by the page name.

app/models/enqueuer.rb
jayroh: Should this be `enqueue(Export.create!(params))` (the `!` isn't there in the line above)

lib/mi_bridges/driver/base_page.rb
jayroh: A-HA ☝️ !

app/models/enqueuer.rb
zspencer: Yep! Will fix Monday :)

zspencer:
@jayroh / @jessieay - Ready for feedback and then (hopefully!) merge.

app/controllers/admin/snap_applications_controller.rb
```diff
@@ -17,12 +17,13 @@ class SnapApplicationsController < Admin::ApplicationController
...
-    #
     def resend_fax
       application = SnapApplication.find(params[:id])
-      Export.create_and_enqueue!(snap_application: application,
-                                 destination: :fax, force: true)
-      flash[:notice] = "Resent fax for #{application.signature}!"
+      Enqueuer.create_and_enqueue_export!(snap_application: application,
+                                          destination: :fax,
+                                          force: true)
+      flash[:notice] = "Resent fax to #{application.office.number} " \
```
jessieay: should we have a test for this? Sometimes I've added flash messages in the wrong place and they don't show up but it's hard to keep track of that without a spec.

app/jobs/application_submitted_sms_job.rb
```diff
@@ -3,7 +3,7 @@ def perform(export:)
     export.execute do |snap_application|
       Sms.new(snap_application).deliver_application_submitted_message
 
-      "Sent SMS to #{snap_application.phone_number}"
+      logger.info("Sent SMS to #{snap_application.phone_number}")
```
jessieay: nice! I find this much more intuitive with `logger.info` than just having the string

app/jobs/submit_application_via_mi_bridges_job.rb
```diff
@@ -0,0 +1,9 @@
+class SubmitApplicationViaMiBridgesJob < ApplicationJob
+  def perform(export:)
+    export.execute do |snap_application, logger|
+      MiBridges::Driver.new(logger: logger,
+                            snap_application: snap_application).run
+      logger.info("Exported via MiBridges successfully")
```
jessieay: I think the name is `MI Bridges`

app/models/enqueuer.rb
```diff
@@ -0,0 +1,49 @@
+class Enqueuer
+  DELAY_THRESHOLD = 30
+  def enqueue_faxes
+    SnapApplication.
+      faxable.
+      untouched_since(DELAY_THRESHOLD.minutes.ago).
+      find_each do |snap_application|
+        create_and_enqueue_export(destination: :fax,
```
jessieay: small style nit:

I find this kind of spacing easier to read

```
 create_and_enqueue_export(
   destination: :fax,
   snap_application: snap_application,
)
```

app/models/enqueuer.rb
```diff
@@ -0,0 +1,49 @@
...
+                                  snap_application: snap_application)
+      end
+  end
+
+  def create_and_enqueue_export(params)
+    enqueue(Export.create(params))
+  end
+
+  def self.create_and_enqueue_export(params)
```
jessieay: can we put class methods above instance methods here? That's my general rule of thumb for class organization: class methods, instance methods (also dependency order, where methods that call other methods are above all methods they depend on)

app/models/export.rb
```diff
@@ -18,62 +15,32 @@ class UnknownExportTypeError < StandardError; end
...
 
-    if snap_application.exports.succeeded.
-        for_destination(destination).present? && !force
+    if snap_application.exports.for_destination(destination).
+        successful_or_in_flight.without(self).present? && !force
 
-      transition_to new_status: :failed
-      update(metadata: "Export failed because a previous export succeeded",
+      transition_to new_status: :unnecessary
+      update(metadata: "Another export for this destination and app occurred",
```
jessieay: this metadata message is a bit confusing to me. Maybe this is related to me not understanding when a status of `unnecessary` would occur

app/models/export.rb
```diff
@@ -82,6 +49,24 @@ def execute
     snap_application.pdf.try(:unlink)
   end
 
+  def logger
+    return @_logger if @_logger.present?
+
+    logger = ActiveSupport::Logger.new(logger_output)
+    logger.level = Logger::DEBUG
```
jessieay: We are using a logger level of "info" on prod, which I believes means this will only show up in development / test envs. Is that expected?

lib/mi_bridges/driver.rb
```diff
@@ -86,15 +87,27 @@ def run
 
     private
 
-    attr_reader :snap_application
+    def logger=(logger)
```
jessieay: seems like we defined almost the same method in `Export.rb` -- should it be extracted?

app/models/enqueuer.rb
```diff
@@ -0,0 +1,49 @@
...
+    new.create_and_enqueue_export(params)
+  end
+
+  def create_and_enqueue_export!(params)
```
jessieay: this `Enqueuer` class does appear to be doing more than enqueueing -- its also creating an export! did you think about using a different name or dependency injection to do the creation outside of the class?

app/controllers/admin/snap_applications_controller.rb
zspencer: It depends on how much we want to invest in testing our admin panel; I've been working from the assumption that the admin panel is significantly lower important/priority than the other stuff; so tests covering all the details is not as important; given that it will likely be in significant flux.

That said, I'm always happy to add more tests, especially if someone else is validating their value.

app/models/enqueuer.rb
zspencer: I don't see creation of the export that needs to be enqueued as an independent piece of responsibility from enqueueing; but I am happy to hear name suggestions that you feel are more reasonable for this extraction?

jessieay: Someone very Ruby smart once told me that whenever there is an `and` in a class or method name, that means that two things are happening. Not terrible to have two things here (much better than 3+!) was just wondering if it would make sense to pass in the export, so it would be something like

```
Enqueuer.enqueue_export(
  destination: :fax,
  snap_application: current_snap_application,
  export: Export.create!(params),
)
```

zspencer: I think what you're smelling here is totally valid; but I would rather we came up with a solution that maintains the current level of encapsulation as opposed to one that moves responsibility up.

So yes, naming is hard. However, my impression is that we're trying to avoid over-investing at this time; and I'm feeling like this is a six in one half dozen in another given that the results would shove a bunch of responsibility into the controller for making sure that jobs can be audited and tracked correctly.

app/models/export.rb
zspencer: Good point. Going to rephrase it to include more useful information even though it makes it a bit longer.

app/models/export.rb
zspencer: This is it's own unique logger which will track all stuff at the debug level. It will not pass them through to the normal logs, however. I considered making a multi-logger adapter but decided it wasn't worth the investment. Happy to add it tho so that we can store the debug information in the database for easier debugging and have it pass through info to the normal papertrail logger.

lib/mi_bridges/driver.rb
zspencer: @jessieay - Yea, the seams aren't quite there, and since there are only two and not three I kicked it back.

zspencer: Wound up extracting a LoggerFactory, but I am somewhat filled with shame for it.

app/jobs/application_submitted_sms_job.rb
zspencer: Yea, I was really happy how passing the logger down into the Driver allowed me to backfill a more explicit way of storing information in the export.

app/models/export.rb
```diff
@@ -18,62 +15,35 @@ class UnknownExportTypeError < StandardError; end
...
 
-    if snap_application.exports.succeeded.
-        for_destination(destination).present? && !force
+    if snap_application.exports.for_destination(destination).
+        successful_or_in_flight.without(self).present? && !force
 
-      transition_to new_status: :failed
-      update(metadata: "Export failed because a previous export succeeded",
+      transition_to new_status: :unnecessary
+      update(metadata: "There is already another successful or in progress " \
```
jessieay: 👏

PR #243, opened 2017-09-15
OK, maybe this won't fail on first build?

zspencer:
Reason for Change
===================
* Heroku kept saying deploys failed, which was making me worried we can't use review apps very effectively when they are useful. I think it's because `db:setup` does a create operation which fails.

### DIFF
--- a/app.json
+++ b/app.json
@@ -2,7 +2,7 @@
   "name": "michigan-benefits",
   "description": "Providing SNAP better, stronger, and faster",
   "scripts": {
-    "postdeploy": "bin/rails db:setup"
+    "postdeploy": "bin/rails db:migrate db:seed"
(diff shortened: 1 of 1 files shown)


zspencer:
Excellent. I was wondering why the first build was failing and that review apps were having weird issues when I visited them the first time

jayroh:
So this worked, @zspencer?  If so - merge it in, man!

Somewhat related question - are you aware if a webhook event is fired off per creation of a review app? *IF* that's the case - I could really make use of that over on the Shubox side to make things work just right across the board.

zspencer:
Yep! Just want a second pair of :eyes:. I  don't like merging without at least someone else taking a gander.

PR #245, opened 2017-09-15
Limit MI Bridges secret question answer char count

jessieay:
* During sign IN, MI Bridges only lets you answer security questions
  with answers up to 20 chars
* 32 is default string length for SecureRandom

### DIFF
--- a/lib/mi_bridges/driver/create_account_page.rb
+++ b/lib/mi_bridges/driver/create_account_page.rb
@@ -3,6 +3,10 @@
 module MiBridges
   class Driver
     class CreateAccountPage < BasePage
+      MAXIMUM_SECRET_QUESTION_ANSWER_CHAR_COUNT = 20
+      MAXIMUM_USER_ID_CHAR_COUNT = 20
+      MAXIMUM_PASSWORD_CHAR_COUNT = 16
+
       def setup; end
 
       def fill_in_required_fields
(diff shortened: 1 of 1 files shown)


lib/mi_bridges/driver/create_account_page.rb
```diff
@@ -49,19 +53,19 @@ def create_driver_application
...
       end
 
       def generate_secret_question_1_answer
-        SecureRandom.hex
+        SecureRandom.hex(MAXIMUM_SECRET_QUESTION_ANSWER_CHAR_COUNT / 2)
       end
 
       def generate_secret_question_2_answer
-        SecureRandom.hex
+        SecureRandom.hex(MAXIMUM_SECRET_QUESTION_ANSWER_CHAR_COUNT / 2)
```
jayroh: Why the need to divide by 2?

jayroh:
I meant to click approve instead of comment, so this is 👍 .  (Just wondering why the division)

lib/mi_bridges/driver/create_account_page.rb
jessieay: > The argument n specifies the length of the random length. The length of the result string is twice of n.

That's how secure random works - thoughts on way to clarify?

https://ruby-doc.org/stdlib-1.9.3/libdoc/securerandom/rdoc/SecureRandom.html

jayroh: So weird. The other approaches I can think of are just too verbose and wordy to really make it worthwhile.

All good in the hood.

PR #246, opened 2017-09-15
Add Optional tags to non-required fields

rachelcope:
Reason for Change
===================
* https://trello.com/c/P6e3PbqN/228-application-as-a-user-i-want-to-know-which-fields-are-required-throughout-the-application 
* Add Optional tag to application questions that are not required

image

### DIFF
--- a/app/views/contact_information/edit.html.erb
+++ b/app/views/contact_information/edit.html.erb
@@ -15,7 +15,7 @@
         autofocus: true,
         options: { maxlength: 10 },
         notes: "No dashes or spaces (example: 5555555555)." %>
-      <%= f.mb_input_field :email, "What is your email address?", notes: ["If you don’t have one or you don’t want to receive emails from us you can leave this blank."] %>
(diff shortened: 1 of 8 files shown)


app/views/contact_information/edit.html.erb
```diff
@@ -15,7 +15,7 @@
         autofocus: true,
         options: { maxlength: 10 },
         notes: "No dashes or spaces (example: 5555555555)." %>
-      <%= f.mb_input_field :email, "What is your email address?", notes: ["If you don’t have one or you don’t want to receive emails from us you can leave this blank."] %>
+      <%= f.mb_input_field :email, "What is your email address?<span class='form-card__optional'>(optional)</span>", notes: ["If you don’t have one or you don’t want to receive emails from us you can leave this blank."] %>
```
jessieay: to simplify this, we can add an `optional` flag to our `MbFormBuilder` that adds this markup. Let's pair on that after retro!

PR #248, opened 2017-09-18
Use the correct region when retrieving files

zspencer:
Reason for Change
===================
* AWS requires us to tell it which region to check against when downloading files; however Bucketeer's `BUCKETEER_REGION` isn't picked up automatically by the S3 gem.
* https://trello.com/c/G9MsDFD1/325-awsregion-missing-on-staging-prod


jessieay:
[AWS_REGION missing on staging / prod](https://trello.com/c/G9MsDFD1/325-awsregion-missing-on-staging-prod)

PR #249, opened 2017-09-18
Don't close PDFs if they haven't been opened

zspencer:
Reason for Change
===================
* Closing the PDF also opened it, which seems unnecessary.
* Debugging exports was weird because we would say the export succeeded, even if the Job still existed because an exception was raised as part of our ensure statement.
* https://trello.com/c/pcoLnZzS/326-fail-exports-appropriately-when-ensure-fails

Minor Changes
=======
* If we fail at any time, let's mark the export as failed.


zspencer:
[Fail exports appropriately when ensure fails](https://trello.com/c/pcoLnZzS/326-fail-exports-appropriately-when-ensure-fails)

PR #250, opened 2017-09-18
Check page title on each driver step

jessieay:
**WHY**: This helps with debugging and is an incremental step toward
filling out pages based on which page you are on rather than assuming a
certain order (which would be tricky since it's possible that there are
many different page orders depending on snap app data)

### DIFF
--- a/lib/mi_bridges/driver/additional_information_page.rb
+++ b/lib/mi_bridges/driver/additional_information_page.rb
@@ -1,6 +1,12 @@
 module MiBridges
   class Driver
     class AdditionalInformationPage < ClickNextPage
+      def setup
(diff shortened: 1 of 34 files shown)


jessieay:
[Determine current MI Bridges page based on page title](https://trello.com/c/uSaMSnex/324-determine-current-mi-bridges-page-based-on-page-title)

jessieay:
@jayroh thoughts on the code climate error? seems weird that it's seeing duplication for two pages only.

jayroh:
@jessieay considering that's only a complaint about duplication I wouldn't give it *too* much credit. Can we tell CodeClimate to ignore those in particular?

PR #251, opened 2017-09-19
Send email to field office after submission

jayroh:
Once a snap application is submitted, in order to make the review of the
application as quick as possible we should email the field office that best
corresponds to the address of the submitting client.

In order to do so:

* Change :email to :client_email and add :office_email. Until now we've only
  emailed the application to a client, but now want to email the application to
  the appropriate field office when the application is done. This changes
  :email to :client_email to be more specific, and adds :office_email as a
  possible destination for the export process.
* Rename `EmailApplicationJob` to `ClientEmailApplicationJob`.  This is one use
  case for emailing the generate PDF and should be qualified as such. The next
  commit will introduce an `OfficeEmailApplicationJob` that will send an email
  to an MDHHS office for their use instead of receiving a Fax.
* SnapApplication admin index updated w/addtl funnel

### DIFF
--- a/app/models/export.rb
+++ b/app/models/export.rb
@@ -1,15 +1,32 @@
 class Export < ApplicationRecord
+  VALID_STATUS = %i(
+    new
+    queued
+    in_process
+    succeeded
+    failed
+    unnecessary
+  ).freeze
+
+  VALID_DESTINATIONS = %i(
+    client_email
+    office_email
+    fax
+    sms
+    mi_bridges
+  ).freeze
+
   belongs_to :snap_application
-  validates :status, inclusion: { in: %i(new queued in_process succeeded failed
-                                         unnecessary) }
-  validates :destination, inclusion: { in: %i(fax email sms mi_bridges) }
+  validates :status, inclusion: { in: VALID_STATUS }
+  validates :destination, inclusion: { in: VALID_DESTINATIONS }
 
   attribute :destination, :symbol
   attribute :status, :symbol
 
   default_scope { order(updated_at: :desc) }
   scope :faxed, -> { where(destination: :fax) }
-  scope :emailed, -> { where(destination: :email) }
+  scope :emailed_client, -> { where(destination: :client_email) }
+  scope :emailed_office, -> { where(destination: :office_email) }
   scope :succeeded, -> { where(status: :succeeded) }
   scope :for_destination, ->(destination) { where(destination: destination) }
   scope :completed, -> { where.not(completed_at: nil) }
--- a/app/views/application_mailer/office_application_notification.html.erb
+++ b/app/views/application_mailer/office_application_notification.html.erb
@@ -1,0 +1,3 @@
+Hello!
+
+A new SNAP application has been submitted and is attached here.
--- a/config/offices.yml
+++ b/config/offices.yml
@@ -45,19 +45,25 @@ defaults: &defaults
 development:
   <<: *defaults
   offices:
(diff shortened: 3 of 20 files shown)


jayroh:
[Email a single PDF application (1171 + at least 1 doc) to both office inboxes](https://trello.com/c/0kWIanwQ/315-email-a-single-pdf-application-1171-at-least-1-doc-to-both-office-inboxes)

app/models/export.rb
```diff
@@ -1,15 +1,32 @@
 class Export < ApplicationRecord
+  VALID_STATUS = %i(
```
jessieay: 🤗

app/models/fax_recipient.rb
```diff
@@ -11,6 +11,10 @@ def name
     office["name"]
   end
 
+  def email
```
jessieay: now that we have this, I wonder if the class name should no longer be `FaxRecipient` ?

app/views/application_mailer/office_application_notification.html.erb
```diff
@@ -0,0 +1,3 @@
+Hello!
+
+A new application has been submitted and is attached here.
```
jessieay: I know this is just placeholder text, but perhaps we should indicate that it's a SNAP application?

config/offices.yml
```diff
@@ -45,19 +45,25 @@ defaults: &defaults
...
+      email: "union@example.com"
     clio:
       fax_number: "+16173963015"
+      email: "clio@example.com"
 production:
   <<: *defaults
   offices:
     staging_sfax:
       fax_number: "+18888433549"
+      email: "gcfengineering+heroku@codeforamerica.org"
```
jessieay: I don't have access to this inbox -- should we use it anyway?

spec/features/admin_dashboard_spec.rb
```diff
@@ -1,17 +1,19 @@
...
-    1.times { create(:snap_application, :emailed, signed_at: nil) }
+    1.times { create(:snap_application, :emailed_client, signed_at: nil) }
     9.times { create(:snap_application, signed_at: 5.days.ago) }
     1.times { create(:snap_application, :faxed, signed_at: 5.days.ago) }
     visit admin_root_path
     expect(page).to have_content "21 Created"
     expect(page).to have_content "10 Signed (47.6%)"
-    expect(page).to have_content "1 Emailed (4.8%)"
+    expect(page).to have_content "1 Emailed Client (4.8%)"
+    expect(page).to have_content "0 Emailed Office (0.0%)"
```
jessieay: perhaps we should add an office email in the setup here so we can be sure the stats are working?

jayroh:
[Spike on Emailing PDFs to MDHHS](https://trello.com/c/8eGN7LY8/304-spike-on-emailing-pdfs-to-mdhhs)

app/models/export.rb
jayroh: :raised_hands:

app/models/fax_recipient.rb
jayroh: Great call! Agree 100%

config/offices.yml
jayroh: @jessieay is (maybe?) the Front mailbox the best address for this right now?

Maybe we should make a temporary @cylinder.work address or alias that this sort of stuff gets funneled through?

jessieay: oh yeah, front might be good! we'd just want to make it clear why something was coming from staging into front

PR #252, opened 2017-09-19
Use first name and age for driving code

jessieay:
* So we can differentiate between household members


jessieay:
[Make it possible to differentiate members based on first name](https://trello.com/c/KA1zH1bX/321-make-it-possible-to-differentiate-members-based-on-first-name)

PR #254, opened 2017-09-19
Handle failed attempts getting non-existent S3 object

jayroh:
In the cases where an object in S3 no longer exists we need to handle
the effects that has in our system. This means we should NOT try to
attach this non-existent document.

This commit rescues the error from the AWS SDK and bubbles up the
resulting nils to the DHS1171Pdf class where it does not include them if
they are nil.

### DIFF
--- a/spec/models/remote_document_spec.rb
+++ b/spec/models/remote_document_spec.rb
@@ -10,6 +10,7 @@ describe RemoteDocument do
 
         downloaded = RemoteDocument.new(document_url).download
 
+        expect(downloaded).not_to eq nil
         expect(file_type(downloaded.tempfile.path)).to eq "image/jpeg"
       end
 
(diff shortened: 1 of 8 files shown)


jayroh:

spec/models/remote_document_spec.rb
```diff
@@ -10,6 +10,7 @@
 
         downloaded = RemoteDocument.new(document_url).download
 
+        expect(downloaded).not_to eq nil
```
adarsh: Minor: `expect(downloaded).not_to be_nil`?

spec/models/remote_document_spec.rb
```diff
@@ -24,6 +25,7 @@
         downloaded = RemoteDocument.new(document_url).download
         imagemagick_file = MiniMagick::Image.open(downloaded.tempfile.path)
 
+        expect(downloaded).not_to eq nil
```
adarsh: Same

spec/models/verification_document_spec.rb
```diff
@@ -44,6 +44,20 @@
...
+      it "returns nil" do
+        remote_document = double("remote_document", download: nil)
+        allow(RemoteDocument).to receive(:new).and_return(remote_document)
+
+        document = VerificationDocument.new(
+          url: "doc url",
+          snap_application: "snap application",
+        )
+
+        expect(document.file).to eq nil
```
adarsh: Minor: `expect(document.file).to be_nil`?

PR #255, opened 2017-09-19
Fill out current page based on title, not order

jessieay:
**WHY**: If we know what page we're on, we know what to do on that page.
This way, our MI Bridges flow is not dependent on a specific order and
will be more flexible as we add more of the pages found during Gap
analysis.

### DIFF
--- a/lib/mi_bridges/driver.rb
+++ b/lib/mi_bridges/driver.rb
@@ -67,7 +67,14 @@ module MiBridges
 
     def run
       setup
+      sign_up_for_account
+      complete_application
+      teardown
+    end
 
+    private
+
+    def sign_up_for_account
       SIGN_UP_FLOW.each do |klass|
         begin
           page = klass.new(@snap_application, logger: logger)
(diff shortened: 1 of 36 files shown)


jessieay:
[Determine current MI Bridges page based on page title](https://trello.com/c/uSaMSnex/324-determine-current-mi-bridges-page-based-on-page-title)

lib/mi_bridges/driver.rb
```diff
@@ -66,25 +68,33 @@ def initialize(snap_application:, logger: nil)
...
         rescue StandardError => e
-          # rubocop:disable Debugger
-          binding.pry if ENV["DEBUG_DRIVE"]
-          # rubocop:enable Debugger
-          throw e
+          debug(e)
         end
       end
 
+      complete_flow_steps
```
jayroh: Above we iterate through the sign up flow, then we while loop through the rest of the pages - is that right?

What do you think about breaking out the iterator above to its own method so it maintains a familiar pattern of implementation as `complete_flow_steps`. I mention that because I was having a hard time determining the difference between what goes on in the `each` and what gets tucked away in `complete_flow_steps`. The former being implemented in the body of this method, versus the latter in a private method threw me off a little bit.

something like

```
sign_up_for_account
complete_application
```

jessieay: great suggestion! done at e7be38d

jessieay:
@jayroh was driving the code during development to make sure it worked, but just confirmed again that the smoke test works!

PR #259, opened 2017-09-20
Pages for mailing addr, pregnancy, self-employment

jayroh:
Page objects for "Mailing Address", "Pregnancy Information", and
"More about <PERSON>'s Self Employment". A bigger change as a result of
the "Self-Employment" section is that the string we're testing against
in the H1 can now be dynamic - it can reference a member of the
application as per this example.

As a result this commit will check for both equality of the string, and
whether there is a match in the chance we make `TITLE` a regular
expression.

### DIFF
--- a/lib/mi_bridges/driver.rb
+++ b/lib/mi_bridges/driver.rb
@@ -23,10 +23,13 @@ module MiBridges
       BenefitsSelectorPage,
       ProgramBenefitsPage,
       PersonalInformationPage,
+      MailingAddressPage,
       BasicInformationSummaryPage,
       PeopleListedPage,
+      PregnancyInformationPage,
       HouseholdMembersSummaryPage,
       JobIncomeInformationPage,
+      MoreAboutSelfEmploymentPage,
       JobIncomeSummaryPage,
       MoneyOtherSourcesPage,
       MoneyOtherSourcesSummaryPage,
@@ -124,7 +127,8 @@ module MiBridges
 
     def find_page_klass
       APPLY_FLOW.detect do |page|
-        page::TITLE == page_title
+        page_title == page::TITLE ||
+          page_title.match?(page::TITLE)
       end
     end
 
--- a/lib/mi_bridges/driver/more_about_self_employment_page.rb
+++ b/lib/mi_bridges/driver/more_about_self_employment_page.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module MiBridges
+  class Driver
+    class MoreAboutSelfEmploymentPage < BasePage
+      TITLE = /More About .*'s Self-Employment/
+
+      delegate :members, to: :snap_application
+
+      def setup; end
+
+      def fill_in_required_fields
+        members.each do |member|
+          fill_in_all_programs_for(member)
+          fill_in_fap_program_benefits_for(member)
+        end
+      end
+
+      def continue
+        click_on "Next"
+      end
+
+      private
+
+      def fill_in_all_programs_for(member)
+        name = member.first_name_and_age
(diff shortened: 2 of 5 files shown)


jayroh:
[Residential / Mailing address in MI Bridges](https://trello.com/c/HesEPXI7/335-residential-mailing-address-in-mi-bridges)

jayroh:
[Pregnancy pages in MI Bridges](https://trello.com/c/VdNIcRWK/334-pregnancy-pages-in-mi-bridges)

jayroh:

lib/mi_bridges/driver.rb
```diff
@@ -124,7 +127,8 @@ def setup
 
     def find_page_klass
       APPLY_FLOW.detect do |page|
-        page::TITLE == page_title
+        page_title == page::TITLE ||
+          page_title.match?(page::TITLE)
```
jessieay: nice! Was wondering when we were going to have to deal with page names that include peoples' FIRST NAME

lib/mi_bridges/driver.rb
```diff
@@ -23,10 +23,13 @@ class Driver
       BenefitsSelectorPage,
       ProgramBenefitsPage,
       PersonalInformationPage,
+      MailingAddressPage,
```
jessieay: now that the order of these doesn't matter for iteration, I wonder if we should alpha order?

lib/mi_bridges/driver/more_about_self_employment_page.rb
```diff
@@ -0,0 +1,45 @@
...
+  class Driver
+    class MoreAboutSelfEmploymentPage < BasePage
+      TITLE = /More About .*'s Self-Employment/
+
+      delegate :members, to: :snap_application
+
+      def setup; end
+
+      def fill_in_required_fields
+        members.each do |member|
```
jessieay: if we have a member who is NOT self employed, will those code error (bc looking for fields that do not exist)?

lib/mi_bridges/driver/pregnancy_information_page.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      def continue
+        click_on "Next"
+      end
+
+      private
+
+      def click_no_one
+        check_in_section "starPregnancyAllPrograms"
```
jessieay: I had to read this a couple of times before I was sure it was working as expected (it is!) 

Wonder if we should have a separate method called `click_no_one` in `BasePage` that just takes a single arg (for clarity) eg:

```
def check_no_one(section)
  check_in_section(section)
end
```

lib/mi_bridges/driver.rb
jayroh: I don't see why not? In a wholly separate commit/PR would be pretty nice. Could make for some rough rebasing if there's another branch in flight that's changing the array ;)

lib/mi_bridges/driver/more_about_self_employment_page.rb
jayroh: I don't think this will error since it just won't find the labels associated with the member's name.

jayroh:
> As we add pages like this, thoughts on making sure we are adding test Snap Apps that will take us to these pages and using that Snap app to run the pre deploy test?

Funny you should say that - the "more comprehensive" sample SNAP application that's generated with `db:seed` is what I was using to traverse through MI Bridges to discover all these 😄 .  So I guess we inadvertantly got ahead of the game with that one.

To address the point of your statement, however - I completely agree. Any sample data we create to get to these pages should be encompassed within the seeded objects.

PR #260, opened 2017-09-21
Add personal information for additional household members

jessieay:
**WHY**:
If a household has more than 1 member, the "People listed" form needs to
be filled out for each additional member.

For non-primary members, required fields are:
- first name
- last name
- marital status
- gender
- DOB
- whether they live in the same household as the primary member

To test, run the driving code with `bin/drive_dev` using the ID of a Snap Application that has more than one household member.

We are determining which household members have not yet been added by checking to see whose "first_name_and_age" is not already on the page. See:

screen shot 2017-09-20 at 4 38 30 pm

### DIFF
--- a/lib/mi_bridges/driver/base_page.rb
+++ b/lib/mi_bridges/driver/base_page.rb
@@ -83,10 +83,10 @@ module MiBridges
         end
       end
 
-      def fill_in_birthday_fields(member)
-        month = padded(member.birthday.month)
-        day = padded(member.birthday.day)
-        year = member.birthday.year
+      def fill_in_birthday_fields(birthday)
+        month = padded(birthday.month)
+        day = padded(birthday.day)
+        year = birthday.year
 
         fill_in "monthgroupDateOfBirth", with: month
         fill_in "dategroupDateOfBirth", with: day
--- a/lib/mi_bridges/driver/base_page.rb
+++ b/lib/mi_bridges/driver/base_page.rb
@@ -82,6 +82,24 @@ module MiBridges
           logger.debug("#{description.upcase}: #{text.join(', ')}")
         end
       end
+
+      def fill_in_birthday_fields(member)
+        month = padded(member.birthday.month)
+        day = padded(member.birthday.day)
+        year = member.birthday.year
+
+        fill_in "monthgroupDateOfBirth", with: month
+        fill_in "dategroupDateOfBirth", with: day
+        fill_in "yeargroupDateOfBirth", with: year
+
+        fill_in "monthconfirmGroupDateOfBirth", with: month
+        fill_in "dateconfirmGroupDateOfBirth", with: day
+        fill_in "yearconfirmGroupDateOfBirth", with: year
+      end
+
+      def padded(int)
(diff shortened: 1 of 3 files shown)


lib/mi_bridges/driver/people_listed_page.rb
```diff
@@ -5,18 +5,70 @@ class Driver
...
+                  end
+      end
 
       def fill_in_required_fields
-        select primary_member.marital_status.titleize, from: "maritalStatus"
-        fill_in "peopleInYourHome", with: snap_application.members.size
+        select member.marital_status.titleize, from: "maritalStatus"
+
+        if for_next_household_member?
+          fill_in "firstName", with: member.first_name_and_age
```
jessieay: note that these fields are required for the primary member, but they are pre-populated after already having been added via `lib/mi_bridges/driver/personal_information_page.rb`

jessieay:
[Add household members in MI Bridges](https://trello.com/c/SgVkgw15/339-add-household-members-in-mi-bridges)

jessieay:
NOTE that this PR does not break the deploy spec because that uses a record with one household member. 

But things are not quite working for households with multiple members because we still need to add relationship info, see https://trello.com/c/bq90t4KV/340-relationship-information-in-mi-bridges

lib/mi_bridges/driver/base_page.rb
```diff
@@ -78,6 +78,24 @@ def log(description, *text)
           logger.debug("#{description.upcase}: #{text.join(', ')}")
         end
       end
+
+      def fill_in_birthday_fields(member)
```
adarsh: Looks like this method requires `member.birthday` more than it needs the full `member` - what do you think about passing in that instead?

lib/mi_bridges/driver/people_listed_page.rb
adarsh: Noted! It would be great to document this in a nice method name somehow, but I can't see an obvious way to do that. Any ideas?

lib/mi_bridges/driver/people_listed_page.rb
```diff
@@ -5,18 +5,52 @@ class Driver
...
+      attr_reader :member
+
+      def for_next_household_member?
+        has_content? "Please tell us about the next person in your home."
+      end
+
+      def find_not_yet_added_member
+        snap_application.members.detect do |member|
+          member_first_name = Regexp.new(
+            Regexp.escape(member.first_name_and_age),
```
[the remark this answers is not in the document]
adarsh: What do you think about extracting this to something like `def escape_parens_in_formatted_name` or the like?

lib/mi_bridges/driver/base_page.rb
jessieay: I like it! done at a1ab73d

lib/mi_bridges/driver/people_listed_page.rb
jessieay: this is a case where TESTS would make great documentation :/

PR #261, opened 2017-09-21
Update Gems

adarsh:
Reason for Change
=================
* [We should keep our dependencies up to date by regularly bundling.][1]
* When doing so, `aws-sdk` jumped up to version 3, which seems drastic.

[1]: http://adarsh.io/save-money-and-be-happier-by-updating-your-gems-every-monday-morning/

Changes
=======
* `bundle update --conservative`
* Lock `aws-sdk` to roughly what we have now.

Minor
-----
* Update `.rubocop.yml` to fit the new naming conventions.


jessieay:
did you consider doing this with something like bummr so that each gem is updated in a separate commit in case of weird failures? https://github.com/lpender/bummr

adarsh:
@jessieay oh I forgot about Bummr! I'll use it next time!

PR #263, opened 2017-09-21
Fix rubocop errors

jessieay:
* These popped up after we upgraded the gem

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,12 +18,12 @@ Style/ArrayJoin:
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#array-join'
   Enabled: false
 
-Naming/AsciiComments:
+Style/AsciiComments:
   Description: 'Use only ascii symbols in comments.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#english-comments'
   Enabled: false
 
-Style/AsciiIdentifiers:
+Naming/AsciiIdentifiers:
   Description: 'Use only ascii symbols in identifiers.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#english-identifiers'
   Enabled: false
--- a/app/controllers/expenses_additional_sources_controller.rb
+++ b/app/controllers/expenses_additional_sources_controller.rb
@@ -6,11 +6,7 @@ class ExpensesAdditionalSourcesController < StandardStepsController
   def step_params
     super.tap do |params_hash|
       params_hash.each do |key, val|
-        params_hash[key] = if val == "true"
-                             true
-                           else
-                             false
-                           end
+        params_hash[key] = val == "true"
       end
     end
   end
--- a/app/helpers/mb_form_builder.rb
+++ b/app/helpers/mb_form_builder.rb
@@ -179,11 +179,11 @@ class MbFormBuilder < ActionView::Helpers::FormBuilder
 
   def label_contents(label_text, notes, optional = false)
     notes = Array(notes)
-    if optional
-      optional_text = "<span class='form-card__optional'>(optional)</span>"
-    else
-      optional_text = ""
-    end
+    optional_text = if optional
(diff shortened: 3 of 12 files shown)


.rubocop.yml
```diff
@@ -18,12 +18,12 @@ Style/ArrayJoin:
...
   Enabled: false
 
-Naming/AsciiComments:
+Style/AsciiComments:
   Description: 'Use only ascii symbols in comments.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#english-comments'
   Enabled: false
 
-Style/AsciiIdentifiers:
+Naming/AsciiIdentifiers:
```
adarsh: Whoops!

app/controllers/expenses_additional_sources_controller.rb
```diff
@@ -6,11 +6,7 @@ class ExpensesAdditionalSourcesController < StandardStepsController
   def step_params
     super.tap do |params_hash|
       params_hash.each do |key, val|
-        params_hash[key] = if val == "true"
-                             true
-                           else
-                             false
-                           end
+        params_hash[key] = val == "true"
```
adarsh: I'm not sure I follow this...

app/helpers/mb_form_builder.rb
```diff
@@ -179,11 +179,11 @@ def mb_checkbox(method, label_text, options = {})
...
-    if optional
-      optional_text = "<span class='form-card__optional'>(optional)</span>"
-    else
-      optional_text = ""
-    end
+    optional_text = if optional
+                      "<span class='form-card__optional'>(optional)</span>"
+                    else
+                      ""
+                    end
```
adarsh: I personally prefer the previous version of this, but I suppose that's another PR.

spec/features/snap_application_with_maximum_info_spec.rb
```diff
@@ -213,10 +213,10 @@ def upload_documents
   def add_document_photo(url)
     input = %(<input type="hidden" name="step[documents][]" value="#{url}">)
     page.execute_script(
-      <<~eos
+      <<~JAVASCRIPT
         document.querySelector('[data-documents-form]').
           insertAdjacentHTML('beforeend', '#{input}')
-      eos
+      JAVASCRIPT
```
adarsh: oh interesting... I prefer this, I believe.

app/controllers/expenses_additional_sources_controller.rb
jessieay: this happened via rubocop's autocorrect. here it is in action:

```
params_hash = { test: "true" }
[10] pry(main)> params_hash.each do |key, val|
[10] pry(main)*   params_hash[key] = val == "true"
[10] pry(main)* end
=> {:test=>true}
[11] pry(main)> params_hash = { test: "blah" }
=> {:test=>"blah"}
[12] pry(main)> params_hash.each do |key, val|
[12] pry(main)*   params_hash[key] = val == "true"
[12] pry(main)* end
=> {:test=>false}
```

jessieay: agree that it's more confusing than the former!

app/helpers/mb_form_builder.rb
jessieay: 🤗

spec/features/snap_application_with_maximum_info_spec.rb
jessieay: Yeah now rubocop wants 'meaningful' heredoc delimiters https://github.com/bbatsov/rubocop/pull/4467/files

(so anything other than EOS)

app/controllers/expenses_additional_sources_controller.rb
jessieay: basically, it's evaluating `val == "true"` and setting the param key to whatever the result of that evaluation is (which is true or false)

PR #265, opened 2017-09-21
Add relationship details for all household members

jessieay:
* This was a doozy!
* The way MI Bridges handles this is that it starts with a page for the
primary member and asks how that person is related to every other
household member.
* The next page asks how the second household member is related to every
other member who is not the primary member
* The third page asks how the third household member is related to ever
member who is not the first or second member
* Etc etc

I tested this locally using a household with 4 members so I have pretty
high confidence that it will work for a household of any size.

NOTE that there are a few weird things here:
1. The relationship mapping is flipped. In our app, we ask 'What is their relationship to you?',
but MI Bridges asks the relationship of the first person to the second.
So if I am the primary member, I would report my child as "Child", but
in MI Bridges I would enter in "Mother" because it asks how Primary
relates to Secondary.
2. Relationship is not a validated field, so we fall back to "Other"
3. We only collect relationship info for primary to other members, so
for any relationship info page that is not the first one, we select "is Related to another way to"
as the relationship.
4. Used a lot of crazy XPATH stuff here because each section is defined
by an id that uses the position of a household member, which felt
brittle.

Example output for non-primary member:
screen shot 2017-09-21 at 3 36 50 pm

Example for primary member (where we actually have relationship data and don't just fall back to other)
screen shot 2017-09-21 at 2 28 30 pm

### DIFF
--- a/lib/mi_bridges/driver/relationship_information_page.rb
+++ b/lib/mi_bridges/driver/relationship_information_page.rb
@@ -1,0 +1,132 @@
+# frozen_string_literal: true
+
+module MiBridges
+  class Driver
+    class RelationshipInformationPage < BasePage
+      TITLE = "Relationship Information"
+
+      def setup
+        @first_member = find_first_member
+        @second_members = find_second_members
+      end
+
+      def fill_in_required_fields
+        second_members.each do |member|
+          fill_in_relationship(member)
+          buy_food_with(member)
+        end
+      end
+
+      def continue
+        click_on "Next"
+      end
+
+      private
+
+      attr_reader :first_member, :second_members
+
+      def fill_in_relationship(member)
+        selector = find(:xpath, relationship_xpath(member))
+        selector.select(mi_bridges_relationship(member))
+      end
+
+      def relationship_xpath(member)
+        second = member.mi_bridges_formatted_name
+        "//*[@title=\"* #{first_member_name}\'s Relationship to #{second}\"]"
+      end
+
+      def buy_food_with(member)
+        buy_food_with_section =
+          find("fieldset", text: buy_food_with_label(member))
+
+        within buy_food_with_section do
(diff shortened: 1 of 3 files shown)


jessieay:
[Relationship information in MI Bridges](https://trello.com/c/bq90t4KV/340-relationship-information-in-mi-bridges)

lib/mi_bridges/driver/relationship_information_page.rb
```diff
@@ -0,0 +1,132 @@
...
+          merge(other_relationship_mapping)
+        mapping[relationship]
+      end
+
+      def female_relationship_mapping
+        {
+          "Child" => "is the Mother of",
+          "Parent" => "is the Daughter of",
+          "Sibling" => "is the Sister of",
+          "Spouse" => "is the Wife of",
```
adarsh: Oh my word.

lib/mi_bridges/driver/relationship_information_page.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      def setup
+        @first_member = find_first_member
+        @second_members = find_second_members
+      end
+
+      def fill_in_required_fields
+        second_members.each do |member|
+          fill_in_relationship(member)
+          buy_food_with(member)
```
adarsh: Since you're passing a `member` to all of these private methods, I wonder if these two procedures might want to be their own object, like `SubmitRelationship.new(member).run` and `BuyFoodWith.new(member).who` or something?

Not a huge deal, just passing the same type of object to lots of private methods always smells like "new object" to me.

jessieay: yeah I totally agree with that, and I think we should do it! I am going to merge this as-is and then do a follow up PR

PR #267, opened 2017-09-22
Improvements to release script

jessieay:
* Grep for "deploy" in heroku releases (if the last change to heroku was
updating an ENV var, the `bin/release-diff` command opens up an invalid
URL because there is not a commit SHA in the latest heroku "release"
* Set up "staging" app release stage for faxes and emails.
* Send staging test emails to hello@michiganbenefits.org, which goes to
front so we all have access to it / can view it.

Screenshot of invalid diff (before "GREP"):
screen shot 2017-09-22 at 9 31 16 am

### DIFF
--- a/bin/release-diff
+++ b/bin/release-diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -e
 
-staging_sha=$(heroku releases -r staging | head -n 2 | tail -n 1 | cut -d " " -f 4)
-prod_sha=$(heroku releases -r production | head -n 2 | tail -n 1 | cut -d " " -f 4)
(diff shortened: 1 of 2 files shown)


bin/release-diff
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -e
 
-staging_sha=$(heroku releases -r staging | head -n 2 | tail -n 1 | cut -d " " -f 4)
-prod_sha=$(heroku releases -r production | head -n 2 | tail -n 1 | cut -d " " -f 4)
+staging_sha=$(heroku releases -r staging | grep "Deploy" | head -n 1 | tail -n 1 | cut -d " " -f 4)
+prod_sha=$(heroku releases -r production | grep "Deploy" | head -n 1 | tail -n 1 | cut -d " " -f 4)
```
adarsh: Oh I like this 💻

PR #268, opened 2017-09-22
Extract relationship/ buy food with sections into classes

jessieay:
**WHY**: There is a lot of logic here. Better to encapsulate in a
service obj.

### DIFF
--- a/lib/mi_bridges/driver/services/submit_buy_food_with.rb
+++ b/lib/mi_bridges/driver/services/submit_buy_food_with.rb
@@ -1,0 +1,48 @@
+module MiBridges
+  class Driver
+    module Services
+      class SubmitBuyFoodWith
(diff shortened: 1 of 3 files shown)


lib/mi_bridges/driver/services/submit_buy_food_with.rb
```diff
@@ -0,0 +1,48 @@
...
+          if second_member.buy_food_with?
+            "Yes"
+          else
+            "No"
+          end
+        end
+      end
+    end
+  end
+end
```
adarsh: Nice.

jessieay: quite a lot of nesting! Wasnt 100% sure on this location for the classes, but figure we can move them later if we come up with a better idea

PR #271, opened 2017-09-25
Delete child records when deleting a snap app

jessieay:
* When we delete a snap app, we don't want the addresses, members, and
exports for that snap app floating around.


jessieay:
[Add dependent: destroy to relationships](https://trello.com/c/6SZD3sTP/351-add-dependent-destroy-to-relationships)

PR #275, opened 2017-09-25
Add basic unit tests for Driver code classes

jessieay:
* Confirms that all classes have certain methods
* Should we keep or remove simplecov filter?


lib/mi_bridges/driver/base_page.rb
```diff
@@ -5,7 +5,7 @@ class Driver
     class BasePage
       include Capybara::DSL
 
-      def initialize(snap_application, logger:)
+      def initialize(snap_application, logger: nil)
```
adarsh: Is there a default logger which is more appropriate to set here than `nil`?

Aside: `nil` and I are not friends.

lib/mi_bridges/driver/base_page.rb
```diff
@@ -78,8 +78,10 @@ def first_name_section(member)
       end
 
       def log(description, *text)
-        logger.tagged(self.class.to_s) do
-          logger.debug("#{description.upcase}: #{text.join(', ')}")
+        if logger.present?
```
adarsh: This is why `nil` and I are not friends.

spec/lib/mi_bridges/driver/pages_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "rails_helper"
+
+RSpec.describe "Driver pages" do
+  MiBridges::Driver::APPLY_FLOW.each do |page|
```
adarsh: This is nice. The alternative is a shared example, maybe, but this feels better since it drives through whatever the actual flow pages are. 👍

spec/spec_helper.rb
```diff
@@ -10,7 +10,7 @@
 end
 
 SimpleCov.start("rails") do
-  add_filter "/lib/mi_bridges/"
+  # add_filter "/lib/mi_bridges/"
```
adarsh: What about something like this as a comment?

```
SimpleCov.start("rails") do
  # To exclude files from coverage analysis:
  # add_filter "/app/path/to/directory/"
end
```

lib/mi_bridges/driver/base_page.rb
jessieay: We could create a NullLogger. 

The reason I did this was that I was constantly instantiating pages in rails console and getting a valid logger each time was a lot of typing and remembering each time. We could have a smarter default, but I am not 100% sure if there is a large benefit to that, since we only do one nil check on this...

spec/spec_helper.rb
jessieay: I commented this rather than deleted because I wasn't sure if we should exclude or not (see PR description). What do you think we should do?

adarsh: Either way is fine, I'm a 1/10 on my suggestion. I trust your discretion.

lib/mi_bridges/driver/base_page.rb
adarsh: I like the `NullLogger`, but I love all null objects.

To me, I feel like `nil` always multiplies so I try to root it out whenever I see it. Not a huge deal though, I'm a 2/10 on this opinion.

spec/spec_helper.rb
jessieay: the question I was (not very clearly) asking was whether you think we should continue to filter out `lib` in our test coverage? downside is that our test coverage is still low when we include it, upside is that it encourages us to add more tests when possible.

adarsh: We should not continue to filter. Let's keep testing, I think!

lib/mi_bridges/driver/base_page.rb
jessieay: I would def go for a null object if we were going to pass this around in multiple places, but for now it's just the one so I am OK with a nil check.

PR #276, opened 2017-09-25
Add "Other" as relationship option

jessieay:
* We do not require clients to submit an answer for this question, but
it's good to give them the option

### DIFF
--- a/spec/steps/household_add_member_spec.rb
+++ b/spec/steps/household_add_member_spec.rb
@@ -10,4 +10,35 @@ RSpec.describe HouseholdAddMember do
   end
 
   include_examples "social security number"
+
+  context "relationship is included in list" do
+    it "is valid" do
+      valid_relationship = "Roommate"
+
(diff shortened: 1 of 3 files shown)


spec/steps/household_add_member_spec.rb
```diff
@@ -10,4 +10,35 @@
...
   include_examples "social security number"
+
+  context "relationship is included in list" do
+    it "is valid" do
+      valid_relationship = "Roommate"
+
+      step = HouseholdAddMember.new(relationship: valid_relationship)
+      step.valid?
+
+      expect(step.errors[:relationship]).to eq []
```
adarsh: I believe you can do `expect(step).to be_valid` also and it might be the same thing and you can avoid calling `.valid?`.

http://www.rubydoc.info/gems/rspec-rails/RSpec/Rails/Matchers#be_valid-instance_method

jessieay: ah yeah -- the thing is, this is not a valid step because there are presence validations on other fields :)

PR #277, opened 2017-09-26
Add "More about <user>'s income'" page

jessieay:
In addition, add more information about when we hit the specific error
when a page's title is not found in our registry of page objects.

Why? Because it will tell us exactly what the page title is that is
being searched for, therefore allowing us to add a page that adheres to
the pattern.

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -7,6 +7,7 @@ class Member < ApplicationRecord
     "Every Two Weeks",
     "Twice a Month",
     "Monthly",
+    "Yearly",
     "Other",
   ].freeze
 
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -1,7 +1,20 @@
 class Member < ApplicationRecord
   AVERAGE_WEEKS_PER_MONTH = 4.33
   MONTHS_PER_YEAR = 12
+  PAYMENT_INTERVALS = [
+    "Hourly",
+    "Weekly",
+    "Every Two Weeks",
+    "Twice a Month",
+    "Monthly",
+    "Other",
+  ].freeze
+
   belongs_to :snap_application
+
+  validates :employed_pay_interval,
+    inclusion: { in: PAYMENT_INTERVALS },
+    allow_nil: true
 
   attribute :ssn
   attr_encrypted(
@@ -60,8 +73,8 @@ class Member < ApplicationRecord
     return if employed_pay_quantity.nil?
     if employed_pay_interval == "Hourly"
       employed_hours_per_week * employed_pay_quantity * AVERAGE_WEEKS_PER_MONTH
-    elsif employed_pay_interval == "Daily"
-      employed_pay_quantity * (AVERAGE_WEEKS_PER_MONTH * 5)
+    elsif ["Every Two Weeks", "Twice a Month"].include? employed_pay_interval
+      employed_pay_quantity * (AVERAGE_WEEKS_PER_MONTH / 2)
     elsif employed_pay_interval == "Weekly"
       employed_pay_quantity * AVERAGE_WEEKS_PER_MONTH
     elsif employed_pay_interval == "Monthly"
--- a/lib/mi_bridges/driver/job_income_more_about_page.rb
+++ b/lib/mi_bridges/driver/job_income_more_about_page.rb
@@ -10,9 +10,15 @@ module MiBridges
       def setup; end
 
       def fill_in_required_fields
-        fill_in "Name of Employer", with: current_member.employed_employer_name
-        select current_member.employed_pay_interval,
-          from: "How often does #{name} get paid? This is #{name}'s pay period"
+        fill_in(
+          "Name of Employer",
+          with: current_member.employed_employer_name,
+        )
+
+        select(
+          current_member.employed_pay_interval,
+          from: "How often does #{name} get paid? This is #{name}'s pay period",
+        )
 
         if hourly?
           fill_in_hourly
@@ -32,15 +38,22 @@ module MiBridges
       end
 
       def fill_in_hourly
-        fill_in "If #{name} gets paid by the hour, ",
-          with: current_member.employed_pay_quantity
-        fill_in "Please tell us how many hours #{name} works",
-          with: current_member.employed_hours_per_week
+        fill_in(
+          "If #{name} gets paid by the hour, ",
+          with: current_member.employed_pay_quantity,
+        )
+
+        fill_in(
+          "Please tell us how many hours #{name} works",
+          with: current_member.employed_hours_per_week,
+        )
       end
 
       def fill_in_salary
-        fill_in "If #{name} earns a salary instead",
-          with: current_member.employed_pay_quantity
+        fill_in(
+          "If #{name} earns a salary instead",
+          with: current_member.employed_pay_quantity,
+        )
       end
 
       def current_member
--- a/lib/mi_bridges/driver/job_income_more_about_page.rb
+++ b/lib/mi_bridges/driver/job_income_more_about_page.rb
@@ -12,8 +12,7 @@ module MiBridges
       def fill_in_required_fields
         fill_in "Name of Employer", with: current_member.employed_employer_name
         select current_member.employed_pay_interval,
-          from: "How often does Complete (40) get paid? "\
-                "This is Complete (40)'s pay period"
+          from: "How often does #{name} get paid? This is #{name}'s pay period"
 
         if hourly?
           fill_in_hourly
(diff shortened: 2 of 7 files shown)


jessieay:

lib/mi_bridges/driver/job_income_more_about_page.rb
```diff
@@ -0,0 +1,56 @@
...
+      def fill_in_required_fields
+        fill_in "Name of Employer", with: current_member.employed_employer_name
+        select current_member.employed_pay_interval,
+          from: "How often does Complete (40) get paid? "\
+                "This is Complete (40)'s pay period"
+
+        if hourly?
+          fill_in_hourly
+        else
+          fill_in_salary
```
jessieay: if someone is not hourly, that doesn't mean that they have a salary. The interval options we provide are: 

```
["Hourly", "Daily", "Weekly", "2 weeks", "Monthly", "Yearly"],
```
I'd imagine that many hourly workers are paid every 2 weeks, for example.

I wonder if this is a case where we need MichiganBenefits.org to map more closely to MI Bridges (eg: a question that asks whether someone is paid hourly or with a salary)... cc @alanjosephwilliams

lib/mi_bridges/driver/job_income_more_about_page.rb
```diff
@@ -0,0 +1,56 @@
...
+          member.mi_bridges_formatted_name == name
+        end.first
+      end
+
+      def hourly?
+        current_member.employed_pay_interval == "Hourly"
+      end
+
+      def fill_in_hourly
+        fill_in "If Complete (40) gets paid by the hour, ",
```
jessieay: I think we need to replace `Complete (40)` with the `current_member.mi_bridges_formatted_name` in this file

lib/mi_bridges/driver/job_income_more_about_page.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+      end
+
+      def continue
+        click_on "Next"
+      end
+
+      private
+
+      def current_member
```
jessieay: can we put this lower on the page so that it is below the methods that call it? I find that more readable!

lib/mi_bridges/driver/job_income_more_about_page.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        if hourly?
+          fill_in_hourly
+        else
+          fill_in_salary
+        end
+      end
+
+      def continue
+        click_on "Next"
```
jessieay: I keep thinking that maybe this should just go in `BasePage`

lib/mi_bridges/driver/job_income_more_about_page.rb
```diff
@@ -0,0 +1,56 @@
...
+    class JobIncomeMoreAboutPage < BasePage
+      TITLE = /More About .*'s Job/
+
+      delegate :members, to: :snap_application
+
+      def setup; end
+
+      def fill_in_required_fields
+        fill_in "Name of Employer", with: current_member.employed_employer_name
+        select current_member.employed_pay_interval,
```
jessieay: are all of the payment intervals that we have in MI Bridges as well?

jayroh: They are now :)

lib/mi_bridges/driver/job_income_more_about_page.rb
jayroh: I took the info directly from MI Bridges and implemented it in the context of MB.org so we're doing a more close 1:1 match in this area.

lib/mi_bridges/driver/job_income_more_about_page.rb
jayroh: I think it's common enough where it probably should be. I *think* there's just one page that has a different implementation?

lib/mi_bridges/driver/job_income_more_about_page.rb
jayroh: Ugh. Brain-fart

app/models/member.rb
```diff
@@ -1,8 +1,21 @@
 class Member < ApplicationRecord
   AVERAGE_WEEKS_PER_MONTH = 4.33
   MONTHS_PER_YEAR = 12
+  PAYMENT_INTERVALS = [
```
jessieay: nice! We should maybe be doing those for more fields in general. Hard to keep track of validations when they're all in the steps! Although are we worried about having duplicate validations and them getting out of sync?

app/models/member.rb
```diff
@@ -1,8 +1,21 @@
...
+    "Hourly",
+    "Weekly",
+    "Every Two Weeks",
+    "Twice a Month",
```
jessieay: isn't it kind of repetitive and confusing to have twice a month and every 2 weeks as options?

lib/mi_bridges/driver/job_income_more_about_page.rb
```diff
@@ -0,0 +1,73 @@
...
+          "Name of Employer",
+          with: current_member.employed_employer_name,
+        )
+
+        select(
+          current_member.employed_pay_interval,
+          from: "How often does #{name} get paid? This is #{name}'s pay period",
+        )
+
+        if hourly?
```
jessieay: I think my comment got swallowed by GitHub before-- if someone is hourly, they are not paid a salary. But if someone is paid weekly, they could still be a non-salaried employee (eg: me!) -- thoughts on how to handle this?

app/models/member.rb
jayroh: It is. I don't know who's in charge of coming up with these options but I do know they crazy

lib/mi_bridges/driver/job_income_more_about_page.rb
jayroh: I think technically you ARE paid hourly, but bill weekly?

The way through this is to take the word of the applicant. If they say "hourly" we go along with that.

jessieay: yes my point exactly -- having a conditional check if someone is hourly and otherwise assuming a salary seems incorrect?

app/models/member.rb
jessieay: do we want to include this data in our model, then? or should we just ignore than option for MI Bridges?

jayroh:
I don't understand how code climate says the check isn't successful and then when I click on "Details" I don't see anything saying what the issue is? @jessieay @adarsh - any ideas?

jayroh:
http://quick.as/lqvdi0pkv

/gif head scratch

lib/mi_bridges/driver/job_income_more_about_page.rb
jayroh: What recourse do we have? Is there another way?

jessieay: If someone is paid yearly, I think we can safely assume they have a salary. Otherwise, we can assume hourly? At least that way we will be reporting their hours per week, which we ask of all ppl but MI Bridges only asks for people who work hourly

app/models/member.rb
jessieay: ah I see you removed "Yearly" as an option. I think we want to keep that for our integration with the PDF, which has a checkbox for "Yearly" when you report salary.

If we keep that, then we can flip the conditional and assume that Yearly payment means salary and anything else is hourly, and we can report the payment interval and hours per week for anyone not on a salary.

This validation is awesome, but I do think we should remove the options we don't want to make available in our app, right?

jayroh: > do we want to include this data in our model, then? or should we just ignore that option for MI Bridges? It would be nice to not allow this kind of insane data in our DB

I mean ... there's a big overlap when you consider the two - "every 2 weeks" versus "twice a month", but I've been at jobs where it's been one and not the other. And then at another job where it's the opposite. It's dumb, but I think it's valid.

I do remember more than a handful of times where I'd get paid 3 times in a month bc of the every 2 weeks thing and think "YEAH!"

And then at the twice a month jobs where sometimes those long ass 31-day months would seriously harsh my mellow.

jayroh:
@jessieay 👍 / 👎 

Missing anything here?

app/models/member.rb
jessieay: ok that's fine -- I guess I'd maybe check with someone on the design side

made a card for that here so we can punt for now https://trello.com/c/PLcgU178

lib/mi_bridges/driver/job_income_more_about_page.rb
jessieay: what do you think of this suggestion?

jessieay:
@jayroh this looks good - approved!!

but I do think we should heavily consider flipping the hourly / salary logic per my comment at https://github.com/codeforamerica/michigan-benefits/pull/277#discussion_r141143473

jayroh:
> but I do think we should heavily consider flipping the hourly / salary logic per my comment

Yup! Did that yesterday: https://github.com/codeforamerica/michigan-benefits/pull/277/commits/a516495374fbad4a5eb18065259de631ff63ed4f

PR #279, opened 2017-09-26
Add default response for Child support question

jessieay:
**WHY**: I didn't catch this before because MI Bridges only shows this
question if there is someone under 18 in the household.

The tricky thing about this is that we know if *anyone* receives child
support income in a household, but MI Bridges wants to know which child.

After you select a name, it follows up with a queston about the amount
of child support income for each name selected.i

Options are:
- always say "no one" (easiest - already done via this PR)
- always select first child (second easiest)
- collect more granular child support data in MI Benefits so we can fill
this out accurately (hardest)

In order to know which is best, it would be good to get context on the
importance of this question for the county workers.

question:

screen shot 2017-09-26 at 10 57 38 am

screen you see if you check a box next to a name:
screen shot 2017-09-26 at 11 14 08 am

### DIFF
--- a/lib/mi_bridges/driver/money_other_sources_page.rb
+++ b/lib/mi_bridges/driver/money_other_sources_page.rb
@@ -7,6 +7,7 @@ module MiBridges
 
       delegate(
         :income_other?,
+        :income_child_support?,
         :income_ssi_or_disability?,
         :primary_member,
         to: :snap_application,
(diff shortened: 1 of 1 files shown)


jessieay:
[Other types of income in MI Bridges](https://trello.com/c/FE7Ia91i/338-other-types-of-income-in-mi-bridges)

jayroh:
[Cash On Hand in MI Bridges](https://trello.com/c/6VJHpEXr/355-cash-on-hand-in-mi-bridges)

jessieay:
it looks like @alanjosephwilliams thinks this is a good move for now, so ready for review from ya @jayroh !

PR #280, opened 2017-09-26
Add cash on hand page object

jayroh:
Cash on hand page shows up amongst all the income questions so this
takes care of that.

Also changing the regex in the title to capture the user's name allows
us to not have to redefine that regex below in the "More about Job
Income" page, just like in this "cash on hand" page.


lib/mi_bridges/driver/cash_on_hand_page.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module MiBridges
+  class Driver
+    class CashOnHandPage < BasePage
```
jessieay: for the other pages like this I've been naming them`MoreAbout XUZ` to try and be as consistent with the title as possible. Thoughts on doing that here too? Or should we rename all?

lib/mi_bridges/driver/job_income_more_about_page.rb
```diff
@@ -3,7 +3,7 @@
 module MiBridges
   class Driver
     class JobIncomeMoreAboutPage < BasePage
-      TITLE = /More About .*'s Job/
+      TITLE = /More About (.*)'s Job/
```
jessieay: oooh should we make this regex change to the other "More" pages as well?

lib/mi_bridges/driver.rb
```diff
@@ -32,6 +32,7 @@ class Driver
       JobIncomeInformationPage,
       JobIncomeMoreAboutPage,
       JobIncomeSummaryPage,
+      CashOnHandPage,
```
jessieay: alpha order?

lib/mi_bridges/driver/job_income_more_about_page.rb
jayroh: We should - I'll make note and open another PR.

lib/mi_bridges/driver.rb
jayroh: If we alpha here, then we should alpha in each of these arrays. If that's cool w/you then I'll do it in a commit here.

jayroh:
Will wait to merge until the base branch is merged.

jayroh:
[Cash On Hand in MI Bridges](https://trello.com/c/6VJHpEXr/355-cash-on-hand-in-mi-bridges)

PR #281, opened 2017-09-26
Add community partner ID when applying

jessieay:
* So they know it's in partnership with Civilla
* On start page

screen shot 2017-09-26 at 12 23 20 pm


jessieay:
[Include Community Partner ID in Drive Attempt](https://trello.com/c/X2AagCKC/319-include-community-partner-id-in-drive-attempt)

PR #282, opened 2017-09-26
Validate employer name for employed members

jessieay:
**WHY**: This is a required field in MI Bridges, so we need to collect
the info in order to drive applications successfully.

screen shot 2017-09-26 at 3 23 05 pm

### DIFF
--- a/app/steps/many_members_step.rb
+++ b/app/steps/many_members_step.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class ManyMembersStep < Step
+  step_attributes :members
+
+  def valid?
+    members.each do |member|
+      member.errors.clear
+      validate_household_member(member)
+    end
+
+    members.map(&:errors).all?(&:blank?)
+  end
+end
--- a/spec/steps/income_details_per_member_spec.rb
+++ b/spec/steps/income_details_per_member_spec.rb
@@ -1,17 +1,28 @@
 require "rails_helper"
 
 RSpec.describe IncomeDetailsPerMember do
-  describe "#working_members" do
-    it "returns members who are employed or self employed" do
-      working = create(:member, employment_status: "employed")
-      not_working = create(:member, employment_status: "not_employed")
-      self_employed = create(:member, employment_status: "self_employed")
-      members = [working, not_working, self_employed]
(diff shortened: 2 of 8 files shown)


jessieay:

app/steps/many_members_step.rb
```diff
@@ -0,0 +1,14 @@
...
+class ManyMembersStep < Step
+  step_attributes :members
+
+  def valid?
+    members.each do |member|
+      member.errors.clear
+      validate_household_member(member)
+    end
+
+    members.map(&:errors).all?(&:blank?)
```
jayroh: If the errors are not empty/blank I think we should bubble up the contents of those objects up to the step.

spec/steps/income_details_per_member_spec.rb
```diff
@@ -1,17 +1,28 @@
...
       snap_application = create(:snap_application, members: members)
 
       step = IncomeDetailsPerMember.new(members: snap_application.members)
+      step.valid?
 
-      expect(step.working_members).to match_array([working, self_employed])
+      expect(step.members.last.errors[:employed_employer_name]).
+        to eq([])
+      expect(step.members.first.errors[:employed_employer_name]).
+        to eq(["Make sure to answer this question"])
```
jayroh: I think(?) the form objects automatically handle displaying errors, yeah? If we bubble up the errors to the step object itself I believe it'll handle it without us reaching down into the associated `members` objects to check for errors. From the step's perspective that would mean all errors would be accessible via the step object itself.

```ruby
expect(step.errors[:employed_employer_name]).
  to eq(["Make sure to answer this question"])
```

I'll pull this down and play with it a little bit to see if I'm wrapping my head around this completely. I might be missing a bit by just looking at the diff :).

jessieay: in order for the errors to show correctly on the form, it needs to be on a `member`

jayroh: Ah-ha! Ok, cool

jessieay: the reason it is bubbling up to the parent `step` is that we do this check inside the `validate` method

PR #283, opened 2017-09-26
Check "homeless" box if address is unstable

jessieay:
**WHY**
* We do not validate residential street / city / zip if unstable address
* MI Bridges has same behavior/
* If someone's address is unstable and we do not mark as homeless,
driving could fail because we may not have street/city/zip

TO TEST:
* re run rake db:seed
* run driver with minimal application


jessieay:
[Check "homeless" for those without a stable address](https://trello.com/c/QvIzD0ti/353-check-homeless-for-those-without-a-stable-address)

PR #284, opened 2017-09-26
Remove foster care income option

jessieay:
**WHY**: as far as we can tell, it is not needed


jessieay:
[Other types of income in MI Bridges](https://trello.com/c/FE7Ia91i/338-other-types-of-income-in-mi-bridges)

PR #285, opened 2017-09-27
Add "More About <client>'s Checking Account" page

jayroh:
Since we don't specifically ask how much is in individual accounts, this
page will continue with an "I don't know" answer.


lib/mi_bridges/driver/vehicles_page.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module MiBridges
+  class Driver
+    class VehiclesPage < BasePage
+      TITLE = "Vehicles"
+
+      def setup; end
+
+      def fill_in_required_fields
+        click_id "assetOther_1"
```
jessieay: maybe we can alias this into a private method so it's clearer what we are clicking on?

lib/mi_bridges/driver/real_properties_page.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module MiBridges
+  class Driver
+    class RealPropertiesPage < BasePage
+      TITLE = "Real Properties"
+
+      def setup; end
+
+      def fill_in_required_fields
+        click_id "assetOther_1"
```
jessieay: maybe we can alias this into a private method so it's clearer what we are clicking on?

lib/mi_bridges/driver/real_properties_page.rb
```diff
@@ -0,0 +1,20 @@
+
```
jessieay: extra empty line :)

jessieay:
maybe add a quick trello card for acceptance step?

jessieay:
[More About Checking Account on MI Bridges](https://trello.com/c/BrW20VVy/356-more-about-checking-account-on-mi-bridges)

PR #287, opened 2017-09-27
Prevent infinite looping over the same page

jayroh:
Occasionally, some pages will be submitted and redirect back to
themselves, which will be filled in the same way, and result in an
infinite loop of:

1. Detect
2. Submit
3. Fail
4. Re-render same page
5. GOTO 1

This commit will keep a count of number of times a particular child
class page object is instantiated, and will raise an exception with that
class's name so that we can tell _which_ page is getting looped over.

### DIFF
--- a/lib/mi_bridges/driver/base_page.rb
+++ b/lib/mi_bridges/driver/base_page.rb
@@ -3,6 +3,8 @@
 module MiBridges
   class Driver
     class BasePage
+      INSTANTIATION_LIMIT = 5
+
       include Capybara::DSL
 
       @@run_counts = {}
@@ -13,7 +15,7 @@ module MiBridges
         @snap_application = snap_application
 
         @@run_counts[self.class.to_s] ||= 0
-        if @@run_counts[self.class.to_s] >= 5
+        if @@run_counts[self.class.to_s] >= INSTANTIATION_LIMIT
           raise MiBridges::Errors::TooManyAttempts.new(self.class.to_s)
         else
           @@run_counts[self.class.to_s] += 1
--- a/lib/mi_bridges/driver/base_page.rb
+++ b/lib/mi_bridges/driver/base_page.rb
@@ -83,7 +83,7 @@ module MiBridges
       end
 
       def skip_instance_limitation?
-        self.class.methods(false).include? :skip_instance_limitation
+        self.class.instance_methods(false).include?(:skip_instance_limitation)
       end
 
       def check_instance_limitation
--- a/lib/mi_bridges/driver/base_page.rb
+++ b/lib/mi_bridges/driver/base_page.rb
@@ -13,8 +13,8 @@ module MiBridges
         @logger = logger
         log("filling out page")
         @snap_application = snap_application
-        return if skip_instance_limitation?
-        check_instance_limitation
+        return if skip_infinite_loop_check?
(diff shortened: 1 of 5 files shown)


lib/mi_bridges/driver/base_page.rb
```diff
@@ -5,10 +5,19 @@ class Driver
...
 
+      @@run_counts = {}
+
       def initialize(snap_application, logger: nil)
         @logger = logger
         log("filling out page")
         @snap_application = snap_application
+
+        @@run_counts[self.class.to_s] ||= 0
+        if @@run_counts[self.class.to_s] >= 5
```
jessieay: technically, a page could show up 5 times and still be valid (eg: there are more than 5 people in a family, so we have a "Personal Information" page for each one". Thoughts on either looking for a higher number *or* potentially checking for that error text /div on the page?

screen shot 2017-09-27 at 10 18 27 am

spec/lib/mi_bridges/driver/pages_spec.rb
```diff
@@ -17,4 +17,28 @@
       end
     end
   end
+
+  context "when instantiating a page too many times" do
+    it "raises an exception" do
+      snap_application = double
+
+      expect do
+        6.times { MiBridges::Driver::BasePage.new(snap_application) }
```
jessieay: would be good to reference a constant or var here so `6` feels less like a magic number

lib/mi_bridges/driver/base_page.rb
jayroh: How many multi-member pages are there where this could pose an issue?  We could "opt out" of those particular pages by resetting the count for that page ... or white-listing them in the base class.

I'm a little iffy on checking for the error div because I'm not sure what to check for that would be universal across them all.

jessieay: Yeah you are right on the "error div" although it might be good to check and see if there is a universal structure, since that seems to be working for the "title" piece of our code (title is always in an h1). 

I think the only multi-member pages are RelationshipInformationPage and PeopleListedPage...so you are thinking we could just not do the looping logic there?

jayroh:
[Prevent driven pages from infinitely looping during the driving process](https://trello.com/c/fSmAxpY0/359-prevent-driven-pages-from-infinitely-looping-during-the-driving-process)

lib/mi_bridges/driver/personal_information_page.rb
```diff
@@ -11,6 +11,8 @@ class PersonalInformationPage < BasePage
         to: :snap_application,
       )
 
+      def skip_instance_limitation; end
```
jessieay: nice! I think we need this for the Relationship page as well

lib/mi_bridges/driver/base_page.rb
```diff
@@ -68,6 +74,27 @@ def next
...
+            logger.debug("#{description.upcase}: #{text.join(', ')}")
+          end
+        end
+      end
+
+      def skip_instance_limitation?
+        self.class.instance_methods(false).include?(:skip_instance_limitation)
+      end
+
+      def check_instance_limitation
```
jessieay: maybe this could be renamed to something like `check_infinite_loop` so the meaning is clearer?

PR #289, opened 2017-09-27
Actually enroll people in FAP :)

jessieay:
* This tricky little optional checkbox needs to be checked to enroll
members in food assistance.

screen shot 2017-09-27 at 3 22 02 pm

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -40,15 +40,15 @@ minimal_application.update!(
 
 minimal_mailing = Address.find_or_initialize_by(
   snap_application: minimal_application,
-  street_address: "",
 )
 
 minimal_mailing.update!(
-  city: "",
-  zip: "",
+  street_address: "123 Main St.",
+  city: "Flint",
+  zip: "12345",
   county: "Genesee",
   state: "MI",
-  mailing: false,
+  mailing: true,
 )
 
 minimal_primary = Member.find_or_initialize_by(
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -40,15 +40,15 @@
...
 minimal_mailing.update!(
-  city: "",
-  zip: "",
+  street_address: "123 Main St.",
+  city: "Flint",
+  zip: "12345",
   county: "Genesee",
   state: "MI",
-  mailing: false,
+  mailing: true,
```
jessieay: updated this to make it more in line with reality: we get a mailing address for all people, but a residential address for some only. 

We validate all fields are present on a mailing address

lib/mi_bridges/driver.rb
```diff
@@ -142,7 +142,6 @@ def debug(e)
       # rubocop:disable Debugger
       binding.pry if ENV["DEBUG_DRIVE"]
       # rubocop:enable Debugger
-      raise e
```
jessieay: by not re-raising we can continue to step through the app even if a debugger statement arises. Is that problematic?

jessieay:
[Check necessary boxes for FAP](https://trello.com/c/UEqmh1V5/360-check-necessary-boxes-for-fap)

lib/mi_bridges/driver.rb
jayroh: I would suggest, instead, something like

```ruby
if ENV["DEBUG_DRIVE"]
  binding.pry
else
  raise e
end
```

That way we can continue only if we're in debug mode. Which I think is fair.

jessieay: done!

PR #291, opened 2017-09-28
Drive medical bills/expenses pages

jayroh:
The medical bills page provides a collection of check boxes to state
what sort of medical bills the client pays. Subsequent pages are shown
depending on what was checked on that medical bills page.

This commit handles those pages. It also adds ALL possible medical
expenses that correspond to the MI Bridges options into the
"complete_application" object we create in the db seed task.

Note that all of the derivitive "more information" pages are "click
next" pages since we do not collect any more information beyond whether
or not the client is paying something for them.


db/seeds.rb
```diff
@@ -122,7 +122,14 @@
   monthly_medical_expenses: 200,
   monthly_court_ordered_expenses: nil,
   care_expenses: ["childcare"],
-  medical_expenses: ["health_insurance", "prescriptions"],
+  medical_expenses: [
```
jessieay: I thought it was cool the way you added an inclusion validation on `member` for a collection last time we did this -- was it for income types? I can't remember. 

Anyway, what do you think about doing that? And/or adding a constant with all possible values?

Without those types of validations I worry that our seed data will get out of sync with our real data

lib/mi_bridges/driver/medical_bills_page.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      def check_health_insurance
+        check_in_section(
+          "#{first_name_section(primary_member)}sMedicalBills",
+          condition: medical_expenses.include?("health_insurance"),
+          for_label: "Health/hospitalization insurance premiums",
+        )
+      end
+
+      def check_co_pays
```
jessieay: looks like we aren't calling this method

lib/mi_bridges/driver/medical_bills_page.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+module MiBridges
+  class Driver
+    class MedicalBillsPage < BasePage
+      TITLE = "Medical Bills"
+
+      delegate(
+        :medical_expenses,
+        :primary_member,
```
jessieay: we are filling in all of this info for primary member rather than for specific members. Let's keep track of that via this spreadsheet: https://docs.google.com/spreadsheets/d/1je97jHCu98-yFM5xYerOCb8pf1rPIp6k8QQE3PVc_lE/edit#gid=0

(Alan asked me to record places where MI Bridges and our app are inconsistent for future review)

lib/mi_bridges/driver/medical_bills_page.rb
jessieay: and once we do, we may need a `ClickNextPage` for its follow up...

app/models/snap_application.rb
```diff
@@ -135,4 +160,36 @@ def signed_at_est
...
+  def court_ordered_expenses_values
+    court_ordered_expenses.each do |value|
+      if !COURT_ORDERED_EXPENSES.include?(value)
+        errors.add(
+          :court_ordered_expenses,
+          "'#{value}' is an invalid court-ordered expense",
+        )
+      end
+    end
+  end
```
jayroh: I'd post that Code Climate's complaint that these three methods are too similar, while it's valid, would become more complicated if you jump through the hoops to DRY it up. This is one of those cases where repetition is a by-product of clarity.

Thoughts, @jessieay?

lib/mi_bridges/driver/medical_bills_page.rb
jayroh: @jessieay for michigan benefits we collect this info only for the primary member (it's on the SnapApplications model).  In MI Bridges it only asks for that information for the primary member. It doesn't present those questions per member.

jessieay: just gave you access - sorry for the delay!

Are you sure it isn't only showing this questions for the primary member because we indicate that the primary member has medical expenses?

jayroh:
[Medical Bills on MI Bridges](https://trello.com/c/pfHisfcr/358-medical-bills-on-mi-bridges)

PR #292, opened 2017-09-28
Fix MI Bridges flow for those with an unstable address

jessieay:
* In MI Bridges, someone with an unstable address (Homeless) is not able
to have a residential address as well. The page surfaces an error if
both "Homeless" is checked and a residential address is entered.
* To solve this, we are now returning a `NullAddress` for Snap Apps with
an unstable address.


jessieay:
[Fix issue with residential address / homeless](https://trello.com/c/vQNrpMzP/362-fix-issue-with-residential-address-homeless)

PR #294, opened 2017-09-29
Calculate hourly wage based on pay interval

jessieay:
**WHY**:
- we known pay quantity "per interval". Before, we were filling in the
hourly rate as whatever the pay quantity was, which is only accurate if
the interval happens to be hourly. For all other intervals, slightly
more complex calculations are required.
- took this opp to move the monthly income calculator into its own class
as well.

### DIFF
--- a/lib/mi_bridges/driver/more_about_job_income_page.rb
+++ b/lib/mi_bridges/driver/more_about_job_income_page.rb
@@ -65,7 +65,7 @@ module MiBridges
       end
 
       def current_member
-        @current_member ||= members.select do |member|
+        @_current_member ||= members.select do |member|
           member.mi_bridges_formatted_name == name
         end.first
       end
--- a/lib/mi_bridges/driver/more_about_job_income_page.rb
+++ b/lib/mi_bridges/driver/more_about_job_income_page.rb
@@ -20,7 +20,7 @@ module MiBridges
           from: "How often does #{name} get paid? This is #{name}'s pay period",
         )
 
-        if hourly?
+        if not_salaried?
           fill_in_hourly
         else
           fill_in_salary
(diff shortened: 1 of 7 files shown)


jessieay:
[How to calculate hourly pay for non-salaried people?](https://trello.com/c/PLcgU178/357-how-to-calculate-hourly-pay-for-non-salaried-people)

spec/models/member_spec.rb
```diff
@@ -31,11 +31,16 @@
...
+        ).and_return(calc_double)
 
-        weekly_pay_times_average_weeks_per_month = 433.0
-        expect(member.monthly_income).to eq(
-          weekly_pay_times_average_weeks_per_month,
-        )
+        member.monthly_income
+
+        expect(calc_double).to have_received(:run)
       end
```
jayroh: The test says that "it returns monthly income", but we are asserting that `calc_double` has received `run`. Could we either:

1. Change the description to align with the test's expectation
2. Or, split this out into two tests where one tests the value of the monthly income, and the other tests that calc_double calls run?

lib/mi_bridges/driver/more_about_job_income_page.rb
```diff
@@ -57,7 +65,7 @@ def fill_in_salary
       end
 
       def current_member
-        members.select do |member|
+        @current_member ||= members.select do |member|
```
jayroh: `@_current_member` to keep up with convention?

spec/models/member_spec.rb
jessieay: since we have comprehensive tests for the MonthlyCalculator itself, went with 1 (change description) at 251c58a 🎐

lib/mi_bridges/driver/more_about_job_income_page.rb
jessieay: good catch! 🐟  done at  060d4b2

jessieay:
@jayroh ready for re-review!

PR #296, opened 2017-10-02
Check that "From" header in emails have a name

jayroh:
When emails are sent to an office or a client, the "FROM" field in the
email should have a more approachable value instead of just
"hello@somedomain.com".

This commit changes the FROM header and adds some tests to accompany the
resulting emails' headers, asserting their validity.

### DIFF
--- a/spec/mailers/application_mailer_spec.rb
+++ b/spec/mailers/application_mailer_spec.rb
@@ -1,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe ApplicationMailer do
+  describe ".snap_application_notification" do
+    it "sets the correct headers" do
+      with_modified_env EMAIL_DOMAIN: "example.com" do
+        email = ApplicationMailer.snap_application_notification(
+          file_name: "#{Rails.root}/spec/fixtures/image.jpg",
+          recipient_email: "user@example.com",
+        )
+        from_header = email.header.select do |header|
+          header.name == "From"
+        end.first.value
+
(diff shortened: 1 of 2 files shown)


spec/mailers/application_mailer_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      expect(from_header).to eq %("Michigan Benefits" <hello@example.com>)
+      expect(email.from).to eq(["hello@example.com"])
+      expect(email.to).to eq(["user@example.com"])
+      expect(email.subject).to eq("Your SNAP application")
+    end
+  end
+end
+
+RSpec.describe ApplicationMailer, "office_application_notification" do
```
jessieay: I don't usually see this format in specs (two `RSpec.describe` blocks) -- interesting! Is this something you do often?

jessieay:
loooks good except for Circle

spec/mailers/application_mailer_spec.rb
jayroh: Good catch :). Sometimes when a method has a lot of assertions I'll do it to get a really nice clear bifurcation of responsibilities.

I like that word - "bifurcation".

In any case - here it's probably not necessary. It just kinda showed up off of my fingers and can def be merged. I'll do some tidying up here. Those method names should infer that they're class methods too, which I'll fix.

adarsh:
[Display name for "from" email](https://trello.com/c/fFVfk6jw/300-display-name-for-from-email)

PR #297, opened 2017-10-02
Add high voltage for static pages

jessieay:
* Now, creating any file within the `app/views/pages/FILENAME` directory will
create a view at the route`/pages/FILENAME`
* Moved existing static pages to `static_pages` dir to avoid namespace
conflict with High Voltage
https://github.com/thoughtbot/high_voltage


app/controllers/static_pages_controller.rb
```diff
@@ -1,4 +1,4 @@
-class PagesController < ApplicationController
+class StaticPagesController < ApplicationController
```
jayroh: Does `ApplicationController` get extended by high_voltage? Or should this be subclassing a controller provided by high_voltage?

jessieay: this controller is separate from HighVoltage, by default HV uses `PagesController`, which is why I renamed this existing controller

PR #300, opened 2017-10-03
Save error info for all failed driving attempts

jayroh:
When driving errors out, we should get some visibility into _why_ things
failed. The information we have available upon exceptions being caught
are the exceptions themselves, the message property in the exception,
the current page object we are attempting to run, and the contents of
the page we are on currently.


jayroh:
[As a developer, I want to know to know relevant information about why a drive failed](https://trello.com/c/op1BJO6U/346-as-a-developer-i-want-to-know-to-know-relevant-information-about-why-a-drive-failed)

PR #301, opened 2017-10-03
Medicaid application, Introduction pages

rachelcope:
Changes
=======
* Initial styles for Medicaid application
* Add initial pages for Location, Add primary member, Add household member, Household overview
* Add styles for yes/no, next buttons.

screen shot 2017-10-03 at 3 06 33 pm
screen shot 2017-10-03 at 3 06 39 pm
screen shot 2017-10-03 at 3 06 46 pm
screen shot 2017-10-03 at 3 06 52 pm

### DIFF
--- a/app/views/pages/medicaid/intro_household.erb
+++ b/app/views/pages/medicaid/intro_household.erb
@@ -1,0 +1,33 @@
+<h4 class="step-header step-header__title--centered">Introduction</h4>
+
+<div class="form-card">
+  <header class="form-card__header">
+    <div class="form-card__title">Now tell us about any other people in your household.</div>
+    <p class="text--help text--centered">
+      This includes everyone who lives in your home (including you), even if they are not seeking benefits today.
+    </p>
+  </header>
+
+  <div class="form-card__content">
+    <div class="slab slab--not-padded">
+      <div class="card card--narrow">
+        <p>
+          <i class="button__icon--left icon-check icon-check--color"></i>
+          Harriett Fletcher (That’s you!)
+        </p>
+
+        <a href="/pages/medicaid/intro_household_member" class="button button-full-width button-bold">
+          <i class="button__icon--left icon-add"></i>
+          Add a member
+        </a>
+
+      </div>
+    </div>
+  </div>
+
+  <footer class="form-card__button_right">
+    <a href="#" class="button button--nav button--nav-center button--cta button--full-width">
+      Next
+    </a>
+  </footer>
+</div>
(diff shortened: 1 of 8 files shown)


app/views/pages/medicaid/intro_household.erb
```diff
@@ -0,0 +1,33 @@
...
+      This includes everyone who lives in your home (including you), even if they are not seeking benefits today.
+    </p>
+  </header>
+
+  <div class="form-card__content">
+    <div class="slab slab--not-padded">
+      <div class="card card--narrow">
+        <p>
+          <i class="button__icon--left icon-check icon-check--color"></i>
+          Just me!
```
jessieay: will this change dynamically if there is more than one person? Just curious / thinking ...

app/views/pages/medicaid/intro_household.erb
```diff
@@ -0,0 +1,33 @@
...
+          <i class="button__icon--left icon-add"></i>
+          Add a member
+        </a>
+
+      </div>
+    </div>
+  </div>
+
+  <footer class="form-card__button_right">
+    <a href="" class="button button--nav button--nav-center button--cta button--full-width">
```
jessieay: I *think* that setting the href value to `"#"` is perhaps better? Maybe it doesn't really matter...

app/views/pages/medicaid/intro_household_member.erb
```diff
@@ -0,0 +1,32 @@
...
+  </header>
+
+  <div>
+    <label class="form-question" for="example_text">What is their first name?</label>
+    <input type="text" class="text-input" name="example_text">
+
+    <label class="form-question" for="example_text">What is their last name?</label>
+    <input type="text" class="text-input"  name="example_text">
+
+    <label class="form-question" for="example_text">What is their gender?</label>
```
jessieay: didn't we want some helper text on this re: "Gender on license" or something? I know all of us would love to make this non-binary but we need to collect the info the gov't wants

rachelcope: 💯  @alanjosephwilliams do you have some good copy/helper text for this question?

app/views/pages/medicaid/intro_household.erb
rachelcope: Yes, it will change to Member Name (Me!). If we want to avoid that dynamic change, I can update it with that text even for one person

danielladevera:
my thumbs up with the copy change for gender help text

PR #302, opened 2017-10-03
Use different email subject/body when client in office

jessieay:
**WHY**: to alert office staff


jessieay:
[Change our email subject line and body to copy to reflect origin of the application (in-office or no, which office)](https://trello.com/c/zNa6NF3P/367-change-our-email-subject-line-and-body-to-copy-to-reflect-origin-of-the-application-in-office-or-no-which-office)

jessieay:
@jayroh agreed!

PR #303, opened 2017-10-04
Display driving-related errors in the dashboard

jayroh:
After an application has been driven and runs into a problem, we save
that information in the `driver_errors` table. To get better vsibility
into these instances we should be able to see what's happened in the
admin/dashboard.

Snap applications now have a different color background based on the
status of their previously attempted drives.

snap_applications_-_michigan_benefits

show_snap_application__2_-_michigan_benefits

show_drivererror__1_-_michigan_benefits


jayroh:
[Inspect failed driving sessions from the admin dashboard](https://trello.com/c/ldj6gV8S/318-inspect-failed-driving-sessions-from-the-admin-dashboard)

spec/models/snap_application_spec.rb
```diff
@@ -183,4 +183,33 @@
...
+      it "returns :drive_none" do
+        snap_application = create(:snap_application)
+
+        expect(snap_application.drive_status).to eq :drive_none
+      end
+    end
+
+    context "when there are drive attempts and no errors" do
+      it "returns :drive_in_flight" do
+        driven_application = create(:driver_application)
```
jessieay: the naming in this spec is a bit confusing -- a driven application is an instance of drive application and driven application has a snap application. 

Maybe it should be `completed_driver_application` or something like that instead?

app/models/snap_application.rb
```diff
@@ -54,6 +55,16 @@ class SnapApplication < ApplicationRecord
     where.not(id: Export.emailed_client.succeeded.application_ids)
   end)
 
+  def drive_status
+    if driver_applications.any? && driver_errors.empty?
+      :drive_in_flight
```
jessieay: given the conditionals I am reading above, I wouldn't expect this status to be `in flight` -- maybe something like `successfully completed` instead?  What do you think, is there a clearer way we could know that a drive is in progress?

app/dashboards/snap_application_dashboard.rb
```diff
@@ -59,26 +59,26 @@ class SnapApplicationDashboard < Administrate::BaseDashboard
     vehicle_income: Field::Boolean,
     zip: Field::String,
     members: Field::HasMany,
+    driver_errors: Field::HasMany,
   }.freeze
 
   COLLECTION_ATTRIBUTES = %i[
     id
-    signature
```
jessieay: good call, don't think this added much 🍓

spec/models/snap_application_spec.rb
```diff
@@ -183,4 +183,33 @@
...
+        snap_application = driven_application.snap_application
+
+        expect(snap_application.drive_status).to eq :drive_in_flight
+      end
+    end
+
+    context "when there are drive attempts and no errors" do
```
jessieay: I think this spec description needs an update 🐦 👁

app/models/snap_application.rb
```diff
@@ -54,6 +55,16 @@ class SnapApplication < ApplicationRecord
...
+    elsif driver_applications.any? && driver_errors.any?
```
jessieay: hat if an earlier drive had errors but a newer drive does not?

jayroh:
> Would love to see a screenshot

Ask and ye shall receive!

app/models/snap_application.rb
jayroh: What do you think about - upon re-drive we delete the existing driver errors?

app/models/snap_application.rb
jayroh: yeah good call!

spec/models/snap_application_spec.rb
jayroh: aye aye aye

app/models/snap_application.rb
jessieay: I like that, although it *could* prove challenging if we re-drive something and want to remember why it failed previously (eg: we manually update the data to make it pass but want to pass without manual intervention in the future). Perhaps we do not delete and look at the latest driver application instead?

app/models/snap_application.rb
```diff
@@ -54,6 +55,16 @@ class SnapApplication < ApplicationRecord
     where.not(id: Export.emailed_client.succeeded.application_ids)
   end)
 
+  def drive_status
+    if driver_applications.any? && driver_errors.empty?
```
jessieay: `driver_applications.any? && driver_application.last.errors.empty?`

app/models/snap_application.rb
```diff
@@ -54,6 +55,16 @@ class SnapApplication < ApplicationRecord
...
+      :drive_success
+    elsif driver_applications.any? && driver_errors.any?
```
jessieay: `driver_applications.any? && driver_application.last.errors.any?`

app/views/admin/snap_applications/index.html.erb
```diff
@@ -58,6 +58,26 @@ It renders the `_table` partial to display details about the resources.
...
 
+<section class="main-content__snap-application-legend">
+  <strong>Legend:</strong>
+
+  <span class="drive_success">
+    Successful Drive
+  </span>
+
+  <span class="drive_in_flight">
+    Driving In Progress
```
jessieay: seems like we don't have anything to put in this column for now -- remove or update logic to have in flight be a thing?

jayroh: I'll remove this for now bc, yeah - if it's not a thing, it shouldn't be here!

jayroh:
Ready for re-review / acceptance, @jessieay

jayroh:
> Before you merge, just wanted to make sure you saw these comments (swallowed by GH: #303 (comment))

Yup. I made the changes you suggested.

> Do you foresee any issues with the way we are looking for errors in a world where we are doing multiple drives?

> I like the idea of preserving errors from multiple drives so we can more effectively debug errors that we manually fix on subsequent drives, so maybe that means we will need to look at errors per-drive somehow?

I think preserving them is a good idea. If we look at when the errors occurred we can essentially clump them together?  Or maybe add another field to the errors where it's a timestamp for the time a particular parent "driver_application" is run?

jessieay:
> Or maybe add another field to the errors where it's a timestamp for the time a particular parent "driver_application" is run?

I like that! Then we can find the latest errors easily

jayroh:
Going to merge this in and get another PR out.

PR #304, opened 2017-10-04
Drive all new Snap Apps

jessieay:
* When success page is reached, app is driven
* If app was already driven, do not re-drive
* We are not submitting apps on MI Bridges yet, we only get to the
submit page, but this will allow us to test our driving
code with real data so we can feel more confident when we do start
submitting.

Minor:
* Add helper for running background jobs immediately, h/t https://robots.thoughtbot.com/process-jobs-inline-when-running-acceptance-tests

### DIFF
--- a/app/jobs/drive_application_job.rb
+++ b/app/jobs/drive_application_job.rb
@@ -1,0 +1,7 @@
+class DriveApplicationJob < ApplicationJob
+  def perform(snap_application:)
+    if snap_application.driver_applications.empty?
+      MiBridges::Driver.new(snap_application: snap_application).run
+    end
+  end
+end
--- a/app/controllers/success_controller.rb
+++ b/app/controllers/success_controller.rb
@@ -21,6 +21,10 @@ class SuccessController < StandardStepsController
       destination: :office_email,
       snap_application: current_snap_application,
     )
+
+    DriveApplicationJob.perform_later(
+      snap_application: current_snap_application,
+    )
   end
 
   def after_successful_update_hook
--- a/spec/controllers/success_controller_spec.rb
+++ b/spec/controllers/success_controller_spec.rb
@@ -21,25 +21,26 @@ RSpec.describe SuccessController do
         with(snap_application: current_app, destination: :office_email)
     end
 
+    it "drives the app once" do
+      run_background_jobs_immediately do
+        run_double = double(run: true)
+        allow(MiBridges::Driver).to receive(:new).
+          with(snap_application: current_app).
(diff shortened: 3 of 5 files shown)


jessieay:
[Drive (but do not submit) existing Snap Apps and record any errors we see](https://trello.com/c/SY0ntG6g/380-drive-but-do-not-submit-existing-snap-apps-and-record-any-errors-we-see)

app/jobs/drive_application_job.rb
```diff
@@ -0,0 +1,7 @@
+class DriveApplicationJob < ApplicationJob
+  def perform(snap_application:)
+    if snap_application.driver_applications.empty?
+      MiBridges::Driver.new(snap_application: snap_application).run
+    end
```
jayroh: I was just thinking about this.  Instead of us having two approaches to driving any application, those being:

1. Drive for the first time. IE: `.run`
2. Re-drive / re-run a drive. IE: `.re_run`

What do you think about changing it to push the logic down into the driver where:

1. If a driver_application exists - re-run using the log in flow
2. If no driver_application exists - run the sign-up flow

jessieay: That sounds right -- I originally wanted to have them be entirely different because it seemed simpler to ensure that `run` can only happen once, but that isn't really happening so we may as well add some smart logic. Should I do that via a separate PR?

jessieay: Should also add that this approach would not change the fact that we'd need logic in here to not run or re run, since we don't want to re-attempt a drive or re-drive just because someone reloads the page

jayroh: > Should also add that this approach would not change the fact that we'd need logic in here to not run or re run, since we don't want to re-attempt a drive or re-drive just because someone reloads the page

I think we might still be able to make it idempotent. Like say we did it like:

1. find or create the first driver application for a given snap application.
2. enqueue the job with that driver application
3. add a "driven_at" timestamp column that's nil by default
4. When the job is run, 
   * if `driven_at` is nil, set the timestamp, and run it.
   * if `driven_at` is some time in the last 24 hours, don't run it.

jessieay: ok yeah, or really if there is *any* timestamp we wouldnt want to re-run, since re-running will be an admin only thing.

is this a change you want to see in this PR? I am looking for which feedback you want for this to get merged :)

jayroh: All this can be totally in for further PR's

PR #306, opened 2017-10-05
Record when apps (and its errors) are driven

jayroh:
When we create and drive an application we are saving the information about
the drive to a `driver_applications` record. When these drives occur we
should be recording the date and time WHEN the drive is occurring. In
addition, the associated errors should have the same time set to them so
that errors can be grouped by the time they were all run by its parent
`DriverApplication` object.

### DIFF
--- a/spec/factories/driver_errors.rb
+++ b/spec/factories/driver_errors.rb
@@ -4,6 +4,7 @@ FactoryGirl.define do
     error_message "runtime error"
     page_class "MiBridges::Driver::BasePage"
     page_html "<html></html>"
+    driven_at DateTime.current
 
     driver_application
   end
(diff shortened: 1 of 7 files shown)


spec/factories/driver_errors.rb
```diff
@@ -4,6 +4,7 @@
     error_message "runtime error"
     page_class "MiBridges::Driver::BasePage"
     page_html "<html></html>"
+    driven_at DateTime.current
```
jessieay: do we need `driven_at` on here if we have a FK for the driver application and a created at timestamp for this error?

jayroh: We do because this can be a different time (for older errors) from the parent driver application. Older errors would not be updated along with a recently updated driver application.

I thought the same thing as I was working on this actually but when I considered (_if_ you referent the parent driver application record) that changing the driven at for a driver_application would change driver_errors from a day or two ago, the redundancy made sense.

PR #307, opened 2017-10-05
Scrape Tracking number from post-submit page and save

jessieay:
**WHY**: This is what MDHHS uses to track cases


jessieay:
[Scrape and persist confirmation number after MI Bridges app is submitted](https://trello.com/c/5An1cXny/378-scrape-and-persist-confirmation-number-after-mi-bridges-app-is-submitted)

PR #308, opened 2017-10-05
Add MiBridges module to fix autoloading

jayroh:
Rails' auto-loading might be having a hard time pulling in all of the
driving code because the MiBridges module, which we namespace all our
driving code inside of, is not defined in its own file.

Also:

* In `save_error`, if latest_drive_attempt is null we should not try to
  tap into it.

### DIFF
--- a/lib/mi_bridges/driver.rb
+++ b/lib/mi_bridges/driver.rb
@@ -160,6 +160,8 @@ module MiBridges
     end
 
     def save_error(e, page)
+      return if latest_drive_attempt.nil?
+
       latest_drive_attempt.
         driver_errors.
         create(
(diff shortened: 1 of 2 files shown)


lib/mi_bridges/driver.rb
```diff
@@ -160,6 +160,8 @@ def level
     end
 
     def save_error(e, page)
+      return if latest_drive_attempt.nil?
```
jessieay: might also be a good time to add a `NullDriveAttempt`

jayroh:
> Any good links/resources on the autoloading issue? (wondering what lead you to this conclusion)

When I used the console instaging and tried to drive from there, it said that `MiBridges` wasn't defined and I was all ... wtf?

lib/mi_bridges/driver.rb
jayroh: Let's explore that and see about getting that in soon. I do want to get this in before I sign off for the day. So I'll merge this in and see how things are.

PR #310, opened 2017-10-06
Add chrome and chromedriver buildpacks to app.json

jayroh:
There are a number of dependencies needed to get things driving properly. This PR gets everything set up so that things work.

1. Add chrome and chromedriver buildpacks
2. Use the ENV provided by the Chrome buildpack for the chrome binary
3. Fix issue where we require a concern twice. It breaks due to the callback run upon requiring the file
4. Set `WED_DRIVER` ENV so that the driver doesn't try to run it with the full chrome browser experience
5. Use a fork of the PDFTK buildpack due to a bug that causes conflicts with the other buildpacks
6. Add `no-sandbox` to the list of options provided to the headless chrome driver configuration. Otherwise it crashes
7. Update app.json to inherit attr_encrypted secrets instead of generating them

### DIFF
--- a/app.json
+++ b/app.json
@@ -48,10 +48,10 @@
       "generator": "secret"
     },
     "SECRET_KEY_FOR_DRIVER_APPLICATION": {
-      "generator": "secret"
+      "required": true
     },
     "SECRET_KEY_FOR_SSN_ENCRYPTION": {
-      "generator": "secret"
+      "required": true
     },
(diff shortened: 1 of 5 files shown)


jayroh:
[Drive every application we receive beginning now, but don't submit them](https://trello.com/c/JKEK5GoA/376-drive-every-application-we-receive-beginning-now-but-dont-submit-them)

PR #314, opened 2017-10-10
Move static medicaid pages to where they will live

jayroh:
Keeping @rachelcope's naming as is but just moving them to where they
would live under corresponding controller spaces. Also making them all
`edit.html.erb`


jayroh:
I feel like a wizard.

```sh
for file in ../pages/medicaid/**/*(.); do basename $file | sed 's/\.erb//' | xargs mkdir; done
for file in ../pages/medicaid/**/*(.); do basename $file | sed 's/\.erb//' | xargs -I '{}' git mv ../pages/medicaid/'{}'.erb '{}'/edit.html.erb; done
```

PR #316, opened 2017-10-10
Small refactor of IntroLocationController

jayroh:
Let `existing_attributes` and `step_class` defer to their super classes.
Existing attributes can have the nil object problem fixed with the
lonely `&` operator. This well return nil and get wrapped by the
HashWithIndifferentAccess ... returning an empty hash

```ruby
HashWithIndifferentAccess.new(nil) => {}
```

`step_class` successfully infers the correct, namespaced, class from the
controller path as it is -- not sure why it didn't work for us in
development mode earlier.

### DIFF
--- a/app/controllers/medicaid/intro_location_controller.rb
+++ b/app/controllers/medicaid/intro_location_controller.rb
@@ -7,13 +7,5 @@ module Medicaid
     def ensure_application_present
       true
     end
-
-    def existing_attributes
-      {}
-    end
-
-    def step_class
-      Medicaid::IntroLocation
-    end
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/medicaid/intro_location_controller.rb
```diff
@@ -7,13 +7,5 @@ class IntroLocationController < StandardStepsController
     def ensure_application_present
       true
     end
-
-    def existing_attributes
-      {}
-    end
-
-    def step_class
-      Medicaid::IntroLocation
```
jessieay: the page loads properly without this now?

jayroh: It does.

PR #320, opened 2017-10-11
Put medicaid steps in correct order

jessieay:
* Comment steps that are for multiple members
* Add in forgotten College step


app/steps/medicaid/step_navigation.rb
```diff
@@ -4,41 +4,42 @@ module Medicaid
...
+        Medicaid::ContactHomeAddressController,
+        Medicaid::ContactOtherAddressController,
+        Medicaid::ContactHomelessController,
+        Medicaid::ContactCurrentAddressController,
+        Medicaid::ContactPhoneController,
+        Medicaid::ContactTextMessagesController,
+        Medicaid::ContactEmailController,
+        Medicaid::ContactSsDobController,
+        Medicaid::ContactSocialSecurityController,
+        Medicaid::ConfirmationController,
```
jayroh: Do we know enough yet where we could divide these up into their own subsections? As in - not within the "Introduction" key?

jessieay: good idea! done at d7c7758

PR #321, opened 2017-10-11
Add medicaid_application model & integrate with the form object

jayroh:
Medicaid information must be stored, of course, so this commit
    introduces that first migration and model to get things started.

   In addition, an update to the form builder here allows us to abstract
   the common pattern of the "yes" and "no" buttons on the Medicaid side of
   the application. Clicking one or the other will pass the boolean value
   to the update action, as well as allowing us to determine which page is
   the next page for either option.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -10,31 +10,32 @@
 // Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
 // about supported directives.
 //
-//= require jquery
-//= require jquery_ujs
-//= require handlebars
-//= require_tree .
-var textWell = (function() {
+// = require jquery
+// = require jquery_ujs
+// = require handlebars
+// = require_tree .
+
+var textWell = (function () {
   return {
-    init: function() {
-      $(document).on('click', '.textwell .textwell__expander', function() {
+    init: function () {
+      $(document).on('click', '.textwell .textwell__expander', function () {
         $(this).parent('.textwell').addClass('textwell--expanded')
       })
     }
   }
-})();
+})()
 
-var radioSelector = (function() {
+var radioSelector = (function () {
   var rs = {
-    init: function() {
-      $('.radio-button').each(function(index, button){
-        if($(this).find('input').is(':checked')) {
-          $(this).addClass('is-selected');
+    init: function () {
+      $('.radio-button').each(function (index, button) {
+        if ($(this).find('input').is(':checked')) {
+          $(this).addClass('is-selected')
         }
 
-        $(this).find('input').click(function(e) {
-          $(this).parent().siblings().removeClass('is-selected');
-          $(this).parent().addClass('is-selected');
+        $(this).find('input').click(function (e) {
+          $(this).parent().siblings().removeClass('is-selected')
+          $(this).parent().addClass('is-selected')
         })
       })
     }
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/application.js
```diff
@@ -66,10 +66,30 @@ var checkboxSelector = (function() {
...
-  radioSelector.init();
-  checkboxSelector.init();
-  textWell.init();
-});
+$(document).ready(function () {
+  radioSelector.init()
+  checkboxSelector.init()
+  textWell.init()
+  yesNoButtons.init()
+})
```
jayroh: for what it's worth, my vim linters did a number on this file, so - the important parts are everything related to `yesNoButtons`

app/controllers/steps_controller.rb
```diff
@@ -35,7 +35,7 @@ def step_attrs
   end
 
   def current_path(params = nil)
-    step_path(self.class, params)
+    step_path(self.class, params).gsub("%2F", "/")
```
jayroh: Turns out our namespacing causes this weird side effect where the `/` between medicaid and the rest of the path gets encoded to `%2F` so this is here to fix that.

app/views/medicaid/intro_location/edit.html.erb
```diff
@@ -9,24 +9,30 @@
...
-      <a href="/pages/medicaid/intro_name" class="button button--nav button--cta button--full-width">
-        Yes
-      </a>
-    </footer>
-  </div>
-</div>
+    <div class="form-card__content">
+      <footer class="form-card__buttons">
+        <%= f.mb_boolean_buttons :michigan_resident,
+          no_path: "/pages/medicaid/intro_location_help",
```
jessieay: the way we do this logic in the snap appl flow is that the next step is always `intro_location_help` but that controller has a `skip?` method that skips based on the previous answer - I think we should stick to that pattern for consistency, which will mean that both yes and no will have the same path and we don't need to worry about `redirect_path` stuff in here

db/migrate/20171011140150_create_medicaid_applications.rb
```diff
@@ -0,0 +1,9 @@
+class CreateMedicaidApplications < ActiveRecord::Migration[5.1]
+  def change
+    create_table :medicaid_applications do |t|
+      t.boolean :michigan_resident, null: false
+
+      t.timestamps
```
jessieay: super minor: I prefer timestamps first -- but this will be irrelevant in schema when you merge your other PR :)

app/assets/javascripts/application.js
```diff
@@ -66,10 +66,30 @@ var checkboxSelector = (function() {
   return {
     init: cs.init
   }
-})();
+})()
+
+var yesNoButtons = (function () {
```
jessieay: this is a very cool solution! would be nice to move these JS functions into different files at some point if we can

db/migrate/20171011140150_create_medicaid_applications.rb
jayroh: Rebased it onto that merged branch and re-ran the schema dump so this should be settled up :)

app/steps/medicaid/intro_location.rb
```diff
@@ -2,5 +2,9 @@
 
 module Medicaid
   class IntroLocation < Step
+    step_attributes(
+      :michigan_resident,
+      :redirect_path,
```
jessieay: we don't need this attribute anymore,I think

app/views/medicaid/intro_location/edit.html.erb
```diff
@@ -9,24 +9,28 @@
...
 
-      <a href="/pages/medicaid/intro_name" class="button button--nav button--cta button--full-width">
-        Yes
-      </a>
-    </footer>
-  </div>
-</div>
+    <div class="form-card__content">
+      <footer class="form-card__buttons">
+        <%= f.mb_boolean_buttons :michigan_resident %>
```
jessieay: beauty! 🌹

spec/models/medicaid_application_spec.rb
```diff
@@ -0,0 +1,4 @@
+require "rails_helper"
+
+RSpec.describe MedicaidApplication, type: :model do
```
jessieay: remove this file if it's gonna be blank?

PR #322, opened 2017-10-11
Add skip logic to medicaid controller

jessieay:
* This will be a blueprint for future skips

### DIFF
--- a/app/controllers/concerns/medicaid_flow.rb
+++ b/app/controllers/concerns/medicaid_flow.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module MedicaidFlow
+  def ensure_application_present
+    if current_medicaid_application.blank?
+      redirect_to medicaid_intro_location_steps_path
+    end
+  end
+
+  def current_medicaid_application
+    MedicaidApplication.find_by(id: current_medicaid_application_id)
+  end
+
+  def current_medicaid_application_id
+    session[:medicaid_application_id]
+  end
+end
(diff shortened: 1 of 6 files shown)


app/controllers/medicaid/intro_location_help_controller.rb
```diff
@@ -2,5 +2,16 @@
 
 module Medicaid
   class IntroLocationHelpController < StandardStepsController
+    include MedicaidFlow
+
+    private
+
+    def skip?
```
jessieay: this will be tested via feature spec when we add one -- when should we do that, @jayroh ?

app/controllers/concerns/medicaid_flow.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def current_medicaid_application
+    MedicaidApplication.find_by(id: current_medicaid_application_id)
+  end
+
+  def current_medicaid_application_id
+    session[:medicaid_application_id]
+  end
+end
```
jayroh: If we change the references to `current_medicaid_application` to `current_application` then the two different flows (both `MedicaidFlow` and `SnapFlow`) always have the same method as the entry point to that object. The two of them just access the records in their own way - same interface, different objects returned :)

app/controllers/medicaid/intro_location_help_controller.rb
jayroh: If we have unit tests for the controllers we _could_ delay the feature spec until we have a little more meat in the medicaid side. I think it depends on what we would consider the happy path?

OR we can just start now? 😄

app/controllers/concerns/medicaid_flow.rb
jessieay: yes I want to do this in the next PR to separate that change from these changes -- that work for you?

jayroh: AH! Yeah!

PR #323, opened 2017-10-11
Add yes/no form partials & assoc fields to steps

jayroh:
The views that ask for a yes or no answer can all display those forms
through one shared partial. These views have all been updated to pull in
that (new) partial.

The fields for each of those steps have also all been added to the
related step classes.

Last but not least, after looking at things, it probably makes more
sense to surface any HTML to the views, that CAN be displayed in the
view. The form builder method is just an obfuscation at this point if we
can handle this in the partial.

### DIFF
--- a/app/steps/medicaid/health_disability.rb
+++ b/app/steps/medicaid/health_disability.rb
@@ -2,5 +2,6 @@
 
 module Medicaid
   class HealthDisability < Step
+    step_attributes(:disabled)
   end
 end
--- a/app/steps/medicaid/health_pregnancy.rb
+++ b/app/steps/medicaid/health_pregnancy.rb
@@ -2,5 +2,6 @@
 
 module Medicaid
   class HealthPregnancy < Step
+    step_attributes(:new_mom)
   end
 end
--- a/app/steps/medicaid/income_job.rb
+++ b/app/steps/medicaid/income_job.rb
@@ -2,5 +2,6 @@
 
 module Medicaid
   class IncomeJob < Step
+    step_attributes(:employed)
   end
 end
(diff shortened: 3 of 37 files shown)


app/steps/medicaid/health_disability.rb
```diff
@@ -2,5 +2,6 @@
 
 module Medicaid
   class HealthDisability < Step
+    step_attributes(:has_a_disability)
```
jessieay: `disabled`?

app/steps/medicaid/health_pregnancy.rb
```diff
@@ -2,5 +2,6 @@
 
 module Medicaid
   class HealthPregnancy < Step
+    step_attributes(:recently_pregnant)
```
jessieay: for `snap_applications` we call this `new_mom` because the actual question is about whether someone is pregnant or has been recently pregnant...thoughts on using that here, too?

app/steps/medicaid/income_job.rb
```diff
@@ -2,5 +2,6 @@
 
 module Medicaid
   class IncomeJob < Step
+    step_attributes(:has_a_job)
```
jessieay: `employed` ?

app/steps/medicaid/insurance_current_type.rb
```diff
@@ -2,5 +2,6 @@
 
 module Medicaid
   class InsuranceCurrentType < Step
+    step_attributes(:has_health_insurance)
```
jessieay: `insured` ?

app/steps/medicaid/intro_citizen.rb
```diff
@@ -2,5 +2,6 @@
 
 module Medicaid
   class IntroCitizen < Step
+    step_attributes(:us_citizen)
```
jessieay: this may be looking too far ahead, but later this will be about whether everyone in the household is a citizen. in snap apps we call this `everyone_a_citizen` -- should we use that here, too?

PR #325, opened 2017-10-11
Add dual application homepage

rachelcope:
Changes
=======
* Add dual application homepage and copy.

fireshot capture 41 - michigan benefits - http___localhost_3000_dual-index


jessieay:
tests are failing because you moved the `index` page to `dual_index` - assuming we want both, you'd want to make sure we preserve the existing static page that serves as a home page (and preserve that "bottom partial" as well)

adarsh:
What do you think about renaming the files from `1f34e.png` to `apple.png` or similar? Same with the shield icon. Might be easier to fuzzy-find them.

rachelcope:
@adarsh all the image files are named with the emojione code. I think I'll stick with to remain consistent

adarsh:
@rachelcope makes sense, did not know about that code. Doesn't matter since it's in the `/emojis` directory. 👍

PR #327, opened 2017-10-12
Remove fax rake task

adarsh:
Reason for Change
=================
* We are now emailing applications to MDHHS instead of faxing them.
* We would also like to keep our valuable fax machinery intact, just in case.

Changes
=======
* Remove the rake task which sends unfaxed applications.
* The Heroku scheduler task had already been shut off - this is just to make sure.

Minor
=======
* Rename the spec to match the method name.

### DIFF
--- a/lib/tasks/queue_faxes.rake
+++ b/lib/tasks/queue_faxes.rake
@@ -1,4 +1,0 @@
-desc "Queues faxes for signed, unfaxed applicants"
-task queue_faxes: :environment do
-  ExportFactory.export_unfaxed_snap_applications
-end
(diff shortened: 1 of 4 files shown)


lib/tasks/queue_faxes.rake
```diff
@@ -1,4 +0,0 @@
-desc "Queues faxes for signed, unfaxed applicants"
-task queue_faxes: :environment do
-  ExportFactory.export_unfaxed_snap_applications
```
jessieay: remove this method too?

adarsh: I wasn't sure if we should keep it. Let's burn it down 🔥 🔥 🔥 (done in b976e37)

adarsh: Whoops! Yep, I neglected to save and commit that change. Done in 8206ea1

PR #329, opened 2017-10-12
Hook up more pieces of medicaid flow to database

jessieay:
* Name / gender fields
* Income types
* Job number

For now I am testing this all manually as I go. Hope is to get all steps functional and then write a large feature spec to cover it all

### DIFF
--- a/app/views/medicaid/income_job_number/edit.html.erb
+++ b/app/views/medicaid/income_job_number/edit.html.erb
@@ -12,30 +12,16 @@
     <div class="form-card__title">Tell us how many jobs you currently have.</div>
   </header>
 
-  <div class="form-card__content">
-    <div>
-      <label for="example_radio" class="radio-button">
-        <input type="radio" name="example_radio" id="example_radio">
-        1 job
-      </label>
-
-      <label for="example_radio" class="radio-button">
-        <input type="radio" name="example_radio" id="example_radio">
-        2 jobs
-      </label>
-
-      <label for="example_radio" class="radio-button">
-        <input type="radio" name="example_radio" id="example_radio">
-        3 or more jobs
-      </label>
-    </div>
-    <!-- Replace with partial -->
-    <footer class="form-card__button_right">
-      <a href="/pages/medicaid/income_self_employment" class="button button--nav button--nav-center button--cta button--full-width">
-        Next
-      </a>
-    </footer>
+  <div>
+    <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
+      <%= f.mb_radio_set :number_of_jobs,
+        '',
+        [
+          { value: "1 job", label: "1 job" },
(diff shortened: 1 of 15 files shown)


app/views/medicaid/income_job_number/edit.html.erb
```diff
@@ -12,30 +12,16 @@
...
-      </a>
-    </footer>
+  <div>
+    <%= form_for @step, as: :step, builder: MbFormBuilder, url: current_path, method: :put do |f| %>
+      <%= f.mb_radio_set :number_of_jobs,
+        '',
+        [
+          { value: "1 job", label: "1 job" },
+          { value: "2 jobs", label: "2 jobs" },
+          { value: "3 or more jobs", label: "3 or more jobs" },
```
adarsh: Random question: If you omit the `label`, will it just show the `value`?

app/views/medicaid/income_other_income_type/edit.html.erb
```diff
@@ -11,41 +11,23 @@
   <header class="form-card__header">
     <div class="form-card__title">What type of income do you receive that’s not from a job?</div>
     <p class="text--help text--centered">
-      Check all that apply.
+    Check all that apply.
```
adarsh: Style: shouldn't this be indented?

db/migrate/20171012163307_add_input_fields_to_database.rb
```diff
@@ -0,0 +1,17 @@
...
+  def change
+    add_column :medicaid_applications, :first_name, :string
+    add_column :medicaid_applications, :last_name, :string
+    add_column :medicaid_applications, :gender, :string
+
+    add_column :medicaid_applications, :income_unemployment, :boolean
+    add_column :medicaid_applications, :income_pension, :boolean
+    add_column :medicaid_applications, :income_social_security, :boolean
+    add_column :medicaid_applications, :income_retirement, :boolean
+    add_column :medicaid_applications, :income_alimony, :boolean
```
adarsh: You guys probably already looked at making these a `jsonb` datatype with a hash in it, right? I'm assuming there's some good reason not to go that way...

jessieay: here I am just copying the way we did it for the SNAP flow. I think the benefit of having separate columns is it's easier to render the correct form state when re-visiting a page (rails does it for us ratherthan having to fill in fields based on a hash of data)

app/views/medicaid/income_other_income_type/edit.html.erb
jessieay: yes! will fix in follow up

app/views/medicaid/income_job_number/edit.html.erb
jessieay: good q! not sure. do not want to prod the sleeping beast jus now but going to look into this later

PR #331, opened 2017-10-12
Add income and expenses fields for medicaid

jessieay:


app/controllers/medicaid/amounts_expenses_controller.rb
```diff
@@ -6,8 +6,8 @@ class AmountsExpensesController < StandardStepsController
 
     private
 
-    def ensure_application_present
-      true
+    def step_class
```
adarsh: I wonder if we should add some kind of shared example for these controllers to ensure they all have this private method? Is that too invasive?

app/views/medicaid/income_job_number/edit.html.erb
```diff
@@ -10,7 +10,7 @@
       <%= f.mb_radio_set :number_of_jobs,
         '',
         [
-          { value: "1 job", label: "1 job" },
+          { value: "1 job" },
           { value: "2 jobs", label: "2 jobs" },
```
adarsh: If this worked for `1 job`, can we do it for the 2 and 3 also?

app/views/medicaid/income_job_number/edit.html.erb
```diff
@@ -10,7 +10,7 @@
       <%= f.mb_radio_set :number_of_jobs,
         '',
         [
-          { value: "1 job", label: "1 job" },
+          { value: "1 job" },
```
jessieay: answer to what this looks like with a value and no label: 
screen shot 2017-10-12 at 11 05 57 am

jessieay: reverted at b0dbc0a

app/views/medicaid/income_job_number/edit.html.erb
adarsh: Ah I guess not :)

app/controllers/medicaid/amounts_expenses_controller.rb
jessieay: we do have one in `app/controllers/steps_controller.rb` but it doesn't work for plural step class names because it calls `controller_path.classify.constantize`

PR #332, opened 2017-10-12
Add logic for address flow

jessieay:
* For full explanation, see https://trello.com/c/QwrCdRgF/29-do-you-receive-mail-at-your-residential-address


app/controllers/medicaid/contact_homeless_controller.rb
```diff
@@ -3,5 +3,20 @@
 module Medicaid
   class ContactHomelessController < StandardStepsController
     include MedicaidFlow
+
+    private
+
+    def skip?
+      mail_sent_to_residential? ||
+        reliable_mail_address?
```
adarsh: Super Minor Style: does this need to be 2 lines?

app/controllers/medicaid/contact_homeless_controller.rb
```diff
@@ -3,5 +3,20 @@
...
+    end
+
+    def mail_sent_to_residential?
+      current_application&.mail_sent_to_residential?
+    end
+
+    def reliable_mail_address?
+      current_application&.reliable_mail_address?
```
adarsh: Good use of the lonely operator!

app/controllers/medicaid/contact_mailing_address_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    def skip?
+      mail_sent_to_residential? ||
+        no_reliable_mail_address?
+    end
+
+    def mail_sent_to_residential?
+      current_application&.mail_sent_to_residential?
+    end
+
+    def no_reliable_mail_address?
```
adarsh: These methods seem to be duplicated a lot. I wonder if it makes sense to move into the `StandardStepsController` or `MedicaidFlow`? Not a huge deal, just thought it was worth asking.

app/controllers/medicaid/contact_homeless_controller.rb
jessieay: nope! fixed at 9e16339

app/controllers/medicaid/contact_mailing_address_controller.rb
jessieay: this method in particular just appears once, as far as I can tell. All of these are pretty simple, but they are suppose to help with reading the logic rather than throwing the `!` and `&` stuff all into one big method. I think for now it will be easier to understand if we keep with a little duplication but I will keep my eye on it

PR #333, opened 2017-10-12
Fix steps controller

jessieay:
* Was broken because we weren't including the concern
* Also clear session vars for both types of flows on static pages so
that peoples' data is not overwritten

### DIFF
--- a/app/controllers/static_pages_controller.rb
+++ b/app/controllers/static_pages_controller.rb
@@ -1,17 +1,21 @@
 class StaticPagesController < ApplicationController
   def index
     session[:snap_application_id] = nil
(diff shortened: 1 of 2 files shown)


app/controllers/static_pages_controller.rb
```diff
@@ -1,17 +1,21 @@
 class StaticPagesController < ApplicationController
   def index
     session[:snap_application_id] = nil
+    session[:medicaid_application_id] = nil
```
adarsh: I wonder if it makes sense to move these session-setters into a `before_action`?

jessieay: great idea! done at 9cefe71

PR #334, opened 2017-10-12
Add correct link for Medicaid on dual index

jessieay:


app/views/static_pages/dual_index.html.erb
```diff
@@ -18,7 +18,7 @@ end %>
       Apply for FAP
     <% end %>
 
-    <%= link_to (), class: "button button--cta button--fully-width" do %>
+    <%= link_to "steps/medicaid/intro-location", class: "button button--cta button--fully-width" do %>
```
adarsh: There's no path helper for this, I'm guessing.

jessieay: so there IS! `step_path(Medicaid::StepNavigation.first)`

but it require decoding, a la https://github.com/codeforamerica/michigan-benefits/blob/448fc6334a740559644deb35e45456b68697a544/app/controllers/steps_controller.rb#L78

because of this rails issue: https://github.com/rails/rails/issues/16058

Which  I have no yet found an elegant solution for. So hardcoding is easiest at the moment

PR #336, opened 2017-10-12
Add forms / DB fields for contact fields (Medicaid)

jessieay:


rachelcope:
🚀

db/migrate/20171012210202_add_contact_fields_to_medicaid_applications.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def up
+    add_column :medicaid_applications, :phone_number, :string
+    add_column :medicaid_applications, :sms_phone_number, :string
+    add_column :medicaid_applications, :email, :string
+
+    add_column :medicaid_applications, :sms_consented, :boolean
+    change_column_default :medicaid_applications, :sms_consented, true
+    MedicaidApplication.in_batches.update_all sms_consented: true
```
adarsh: This should at some point be SQL instead of ActiveRecord, but given our time crunch it's ok.

jessieay: yeah, I added the AR class to this migration so it would work if we rename the class but for sure SQL is better

PR #337, opened 2017-10-12
Make html body consistent

rachelcope:
Changes
=======
* Add `form-card__content` class to make html consistent across views with body content

### DIFF
--- a/app/views/medicaid/health_disability/edit.html.erb
+++ b/app/views/medicaid/health_disability/edit.html.erb
@@ -5,7 +5,6 @@
     <div class="form-card__title">
       Do you have a disability?
     </div>
-
     <p class="text--help text--centered">
       This can include living in a nursing home or other medical facility or a
       physical, mental, or emotional health condition that limits activity.
--- a/app/views/medicaid/insurance_current/edit.html.erb
+++ b/app/views/medicaid/insurance_current/edit.html.erb
@@ -5,7 +5,6 @@
     <div class="form-card__title">
       Are you currently enrolled in a health insurance plan?
     </div>
-
     <p class="text--help text--centered">
       You may be eligible for dual enrollment, which can help cover additional
       medical expenses.
--- a/app/views/medicaid/tax_filing/edit.html.erb
+++ b/app/views/medicaid/tax_filing/edit.html.erb
@@ -5,7 +5,6 @@
     <div class="form-card__title">
       Do you plan on filing a federal tax return next year?
     </div>
-
     <p class="text--help text--centered">
       Some households don't file based on their income.
     </p>
--- a/app/views/medicaid/contact_email/edit.html.erb
+++ b/app/views/medicaid/contact_email/edit.html.erb
@@ -5,7 +5,7 @@
     <div class="form-card__title">What is the best email address at which to contact you? (Optional)</div>
   </header>
 
-  <div>
+  <div class="form-card__content">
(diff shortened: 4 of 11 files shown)


app/views/medicaid/contact_current_address/edit.html.erb
```diff
@@ -0,0 +1,27 @@
+<% content_for :header_title, "Contact Information & Followup" %>
```
jessieay: I renamed this page, so this can be removed now :)

app/views/medicaid/insurance_current/edit.html.erb
```diff
@@ -5,7 +5,6 @@
     <div class="form-card__title">
       Are you currently enrolled in a health insurance plan?
     </div>
-
     <p class="text--help text--centered">
```
jessieay: add `form-card__content` ?

app/views/medicaid/tax_filing/edit.html.erb
```diff
@@ -5,7 +5,6 @@
     <div class="form-card__title">
       Do you plan on filing a federal tax return next year?
     </div>
-
```
jessieay: add `form-card__content` ?

app/views/medicaid/health_disability/edit.html.erb
```diff
@@ -5,7 +5,6 @@
     <div class="form-card__title">
       Do you have a disabilty?
     </div>
-
```
jessieay: add `form-card__content` ?

app/views/medicaid/insurance_current/edit.html.erb
rachelcope: @jessieay I've been adding the `form-card__content` it to views that have additional content outside of the header (e.g., views with inputs, radios, etc)

app/views/medicaid/tax_filing/edit.html.erb
rachelcope: Same as above

app/views/medicaid/health_disability/edit.html.erb
rachelcope: Same as above

app/views/medicaid/contact_current_address/edit.html.erb
rachelcope: @jessieay do you mean this page should be deleted?

app/views/medicaid/insurance_current/edit.html.erb
jessieay: below there is ` <%= render 'shared/yes_no_form', step: @step, field: :insured %>` which is what has the yes and no buttons -- was assuming you wanted those wrapped. you don't?

app/views/medicaid/contact_current_address/edit.html.erb
jessieay: yep, I removed in previously and you re-added it here but it is not used

app/views/medicaid/insurance_current/edit.html.erb
rachelcope: No, I don't think I want the footer wrapped in form-card__content. 

The patterns that I was thinking:
`form-card__header` --> Header/help text
`form-card__content` --> Any content/form in card.
`form-card__footer` --> Yes/No or Next

Does that makes sense? Or do you think footer should be inside of content?

PR #341, opened 2017-10-13
No alert after completing Medicaid application

jayroh:
Add the "dual-index" path to the white-list of paths that will not cause
the "you are exiting the app" message to pop up.


jayroh:
[Back link on confirmation page should be disabled in Medicaid flow](https://trello.com/c/LTNfWZcO/395-back-link-on-confirmation-page-should-be-disabled-in-medicaid-flow)

PR #342, opened 2017-10-13
Do not display back arrow on final submission page

jayroh:
Overriding both the back_path var AND the previous_path method to both
return nil will result in the back arrow not showing up.


jayroh:
[Back link on confirmation page should be disabled in Medicaid flow](https://trello.com/c/LTNfWZcO/395-back-link-on-confirmation-page-should-be-disabled-in-medicaid-flow)

PR #343, opened 2017-10-13
Refactor SNAP / Medicaid controllers

jessieay:
* Seeing an issue on prod related to code loading -
`current_application`, which is defined in a concern that is included in
a child concern, was `nil` when being accessed in a parent controller
* Using inheritance rather than concerns should make loading order more
predictable.


jessieay:
[Documents upload not working](https://trello.com/c/MuszoRsw/401-documents-upload-not-working)

PR #346, opened 2017-10-16
Display exports for snap apps in admin/dashboard

jayroh:
In order to get more visibility as to *where* and *when* applications are sent
to the offices in MI we can display the exports in the admin in the context of
their parent snap application.

This commit adds the resource to the admin and adjusts the seed task to create
some fake data for local testing and demo'ing.

### DIFF
--- a/app/controllers/admin/exports_controller.rb
+++ b/app/controllers/admin/exports_controller.rb
@@ -1,0 +1,4 @@
+module Admin
+  class ExportsController < Admin::ApplicationController
+  end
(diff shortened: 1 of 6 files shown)


jayroh:
[Display exports associated with SNAP Applications.](https://trello.com/c/QbSrrTfH/387-display-exports-associated-with-snap-applications)

jayroh:
There is, yeah. There's one for the main snap application page and the members page.

PR #353, opened 2017-10-16
Fix steps page and display medicaid steps

jayroh:
The steps page had run into some issues sometime in the recent past and
was 500'ing out. This commit fixes that issue, as well as:

1. Adds the medicaid steps to the steps index
2. Displays either/or/both the snap and medicaid steps only if there's
   a respective id present in the session.
3. If there is no id present, user is redirected to the homepage
4. Add a controller spec to ensure this continues to be functional
5. Join the two application sections in the routes

### DIFF
--- a/app/views/steps/index.html.erb
+++ b/app/views/steps/index.html.erb
@@ -1,9 +1,10 @@
 <% content_for :header_title, 'Navigation' %>
 <div class="slab">
-  <% StepNavigation.sections.each do |section, steps| %>
+  <% current_application.class.step_navigation.sections.each do |section, steps| %>
     <h3>
       <%= section %>
     </h3>
+
     <% steps.each do |step| %>
       <div>
         <%= link_to step.to_param.titleize,
--- a/spec/controllers/steps_controller_spec.rb
+++ b/spec/controllers/steps_controller_spec.rb
@@ -1,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe StepsController do
+  context "when no applications have been started" do
+    it "redirects to the homepage" do
+      get :index
+
+      expect(response).to redirect_to(root_path)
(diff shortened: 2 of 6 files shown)


jayroh:
[Set up page for viewing all medicaid steps (similar to `/steps` for SNAP)](https://trello.com/c/n4T2rjfq/397-set-up-page-for-viewing-all-medicaid-steps-similar-to-steps-for-snap)

app/views/steps/index.html.erb
```diff
@@ -1,15 +1,36 @@
...
-    <h3>
-      <%= section %>
-    </h3>
-    <% steps.each do |step| %>
-      <div>
-        <%= link_to step.to_param.titleize,
-          step_path(step),
-          class: 'button button--small button--full-width' %>
-      </div>
+  <% if snap_application? %>
```
jessieay: since we check for whether there is a snap app or a medicaid app in the controller (and do not allow people to visit this page if there is neither), we *could* make this page a little smarter and have a `steps` method on each class that returns either `StepNavigation` or `MedicaidStepNavigation` and then iterate over that.

so this page could be:

```
current_application.class.step_navigation.sections.each do |section, steps|
  ...
end
```

not perfect, but an option if you'd like to take it

spec/controllers/steps_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe StepsController do
+  context "when no applications have been started" do
```
jessieay: yay!

app/views/steps/index.html.erb
jayroh: ooooooooo interesting. I like it

PR #354, opened 2017-10-16
Add field for self employment expenses

jessieay:
**WHY**: people need to report these if they are self employed

screen shot 2017-10-16 at 3 38 15 pm


jessieay:
[Include "Self-Employment" in Expenses as well as Income](https://trello.com/c/X8K5XxYK/420-include-self-employment-in-expenses-as-well-as-income)

PR #356, opened 2017-10-17
Update copy around amount of $ received

jayroh:


jayroh:
[Update help text on "Do you get income that's not from a job"](https://trello.com/c/L5HBd0IL/416-update-help-text-on-do-you-get-income-thats-not-from-a-job)

PR #358, opened 2017-10-17
Add document upload capabilities to medicaid flow

jayroh:
Move the document intro and upload pages to the medicaid flow and
filling in the spots to get things to function the same as the SNAP
side.

* Add boolean and document array to the medicaid application table
* Add controller, steps, views, and specs
* Move document partials to a shared namespace so they can be used
  between the two applicatin flows


jayroh:
[[Medicaid Flow] Upload paperwork](https://trello.com/c/MhaaleGv/448-medicaid-flow-upload-paperwork)

jayroh:
Updates to the flow are that the success page is followed by the documents page. When done uploading paperwork you are sent to the homepage (`dual-index`)

confirmation
success

db/migrate/20171017143139_add_upload_documents_to_medicaid_applications.rb
```diff
@@ -0,0 +1,23 @@
...
+    #
+    add_column \
+      :medicaid_applications,
+      :documents,
+      :string,
+      array: true
+
+    change_column_default :medicaid_applications, :documents, []
+
+    commit_db_transaction
```
jessieay: ooh! I feel like I've been missing this line when adding columns with default values (I guess I read the docs in strong migrations but not closely enough)

config/routes.rb
```diff
@@ -20,7 +20,7 @@
   get "/privacy" => "static_pages#privacy"
   get "/terms" => "static_pages#terms"
   get "/union" => "static_pages#union"
-  get "/dual-index" => "static_pages#dual_index"
+  get "/dual-index" => "static_pages#dual_index", as: :medicaid_root
```
jessieay: 👍

jayroh:
[[Medicaid Flow] Submit paperwork intro](https://trello.com/c/FfcX1uSJ/447-medicaid-flow-submit-paperwork-intro)

app/views/medicaid/documents/edit.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  as: :step,
+  builder: MbFormBuilder,
+  url: current_path,
+  method: :put,
+  html: { data: { 'documents-form' => 'true' } } do |f| %>
+
+  <div class="form-card" id="document-drop">
+    <header class="form-card__header">
+      <div class="form-card__title">
+        Upload your paperwork
```
jessieay: this is more of a design comment than a dev comment, but this page seems perhaps needlessly inconsistent with the docs page for SNAP flow. When moving to multi-member, we found that the gender/sex difference between SNAP and Medicaid required a lot of thinking when combining the two flows with the members table. 

In general, I think we should make them as consistent as possible when adding duplication. Not a blocker, just a comment based on yesterday's experiences.

hartsick:
@jayroh Hi! Question (could you help me route this to the right place/person?): what are we setting as the max file size? Locally the max file size (1mb) is preventing me from uploading photos taken with my phone, which seems a likely use case.

app/views/medicaid/documents/edit.html.erb
jayroh: Could you give me a little more context?  I'm not sure I understand where the inconsistency lies between the two flows.

jessieay: if the pages were exactly the same, I'd expect we could `render template:` in the medicaid docs controller and use the same view for SNAP docs and Medicaid docs. Haven't done an exact diff here, but at the very lease the copy is different (don't remember us calling it "Paperwork" for SNAP)

db/migrate/20171017143139_add_upload_documents_to_medicaid_applications.rb
hartsick: yeah, this is interesting!

app/views/medicaid/documents/edit.html.erb
jayroh: Ah - ok I see. It's a conversation we should bring up then.

jayroh:
> Question (could you help me route this to the right place/person?): what are we setting as the max file size? Locally the max file size (1mb) is preventing me from uploading photos taken with my phone, which seems a likely use case.

Oh yeah. Good question.

When in dev mode (ie: using localhost, 127.0.0.1, 0.0.0.0, etc) the Shubox library uses an s3 bucket that is owned by Shubox. It's a "sandbox" bucket. This allows you to get things working without jumping through all the s3 setup hoops. In order to mitigate ginormous files from being uploading on the regular though, I set a limit of 1MB.

Now, if you're using it on your phone, I presume you're *not* visiting localhost? Is that right? What domain are you visiting?

For "production" domains using shubox (staging, www, etc) you can set the ceiling on the file size in the initializer, which is here:

https://github.com/codeforamerica/michigan-benefits/blob/master/app/assets/javascripts/upload_setup.js#L9

So, it should not complain at that point.

I hope I've added some helpful context here but if you hit me up on slack we can talk about the problem you're running into on your phone - specifically.

hartsick:
> When in dev mode (ie: using localhost, 127.0.0.1, 0.0.0.0, etc) the Shubox library uses an s3 bucket that is owned by Shubox. It's a "sandbox" bucket. This allows you to get things working without jumping through all the s3 setup hoops. In order to mitigate ginormous files from being uploading on the regular though, I set a limit of 1MB.

Thanks for explaining!

PR #360, opened 2017-10-18
Allow multiple members for Medicaid application

hartsick:
For single member flow, it made sense to save some user details on the application.
Now with multiple members, it made sense to move that information from the application and into the member class. To do this, we made applications (Snap and Medicaid) a polymorphic association on the Member class.

To test, create a Medicaid application and confirm you can add multiple members.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 .rspec
 mkmf.log
 Guardfile
+.idea
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -9,7 +9,7 @@ class Member < ApplicationRecord
     "Other",
   ].freeze
 
-  belongs_to :snap_application
+  belongs_to :benefit_application, polymorphic: true
 
   validates :employed_pay_interval,
     inclusion: { in: PAYMENT_INTERVALS },
@@ -26,7 +26,7 @@ class Member < ApplicationRecord
   end
 
   def primary_member?
-    snap_application.primary_member.id == id
+    benefit_application.primary_member.id == id
   end
 
   def monthly_income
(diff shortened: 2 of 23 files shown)


hartsick:
[[Medicaid MM Flow] Intro: Add a member of your household](https://trello.com/c/dWb9ERgX/429-medicaid-mm-flow-intro-add-a-member-of-your-household)

.gitignore
```diff
@@ -14,3 +14,4 @@
 .rspec
 mkmf.log
 Guardfile
+.idea
```
jayroh: oooooh, do tell. What's a dot idea?

app/models/member.rb
```diff
@@ -9,7 +9,7 @@ class Member < ApplicationRecord
     "Other",
   ].freeze
 
-  belongs_to :snap_application
+  belongs_to :benefit_application, polymorphic: true
```
jayroh: 👍

db/migrate/20171017191829_extract_members_from_medicaid_app.rb
```diff
@@ -0,0 +1,60 @@
...
+    safety_assured do
+      execute <<~SQL
+        UPDATE members SET snap_application_id=benefit_application_id;
+      SQL
+
+      remove_column :members, :benefit_application_id
+      remove_column :members, :benefit_application_type
+    end
+  end
+end
```
jayroh: Oh my - nice work!

.gitignore
jessieay: I think this is for Christa's rubymine setup!

db/schema.rb
```diff
@@ -128,7 +125,6 @@
     t.string "residential_zip"
     t.boolean "self_employed"
     t.integer "self_employed_monthly_income"
-    t.integer "self_employment_expenses"
```
jessieay: hmmmm @hartsick it seems like this was removed accidentally -- rebase and re-migrate?

.gitignore
jayroh: Oooooh gotcha!

hartsick: Yep, that's correct! It's a common editor at CfA so figured I'd toss in

db/schema.rb
hartsick: Good catch!

PR #362, opened 2017-10-18
Add "other" insurance type

jessieay:


jessieay:
[[Medicaid Flow] Add "Other" to What time of plan options in single member flow](https://trello.com/c/smMwvanl/452-medicaid-flow-add-other-to-what-time-of-plan-options-in-single-member-flow)

PR #363, opened 2017-10-18
Allow clients to select up to 10 jobs

jessieay:
* This PR adds a "new_job_number" column, which is an integer rather
than a string. The old job number column will be removed in a follow up
PR and `new_job_number` renamed to `job_number`.

* Next up is adding income inputs for each job.

* Open question:

* How do we feel about defaulting the number to 4 on the "continued" page?

screen shot 2017-10-18 at 11 31 29 am

screen shot 2017-10-18 at 11 09 40 am

### DIFF
--- a/app/controllers/medicaid/income_job_number_controller.rb
+++ b/app/controllers/medicaid/income_job_number_controller.rb
@@ -4,6 +4,22 @@ module Medicaid
   class IncomeJobNumberController < MedicaidStepsController
     private
 
+    def existing_attributes
+      HashWithIndifferentAccess.new(number_of_job_attributes)
+    end
+
+    def number_of_job_attributes
+      { new_number_of_jobs: number_of_jobs }
+    end
+
+    def number_of_jobs
+      if current_application.new_number_of_jobs&. > 4
+        4
+      else
+        current_application.new_number_of_jobs
+      end
+    end
+
     def skip?
       not_employed?
     end
--- a/db/migrate/20171018174809_number_of_jobs_to_integer.rb
+++ b/db/migrate/20171018174809_number_of_jobs_to_integer.rb
@@ -7,9 +7,11 @@ class NumberOfJobsToInteger < ActiveRecord::Migration[5.1]
         UPDATE medicaid_applications
         SET new_number_of_jobs=1
         WHERE number_of_jobs='1 job';
+
         UPDATE medicaid_applications
         SET new_number_of_jobs=2
         WHERE number_of_jobs='2 jobs';
+
         UPDATE medicaid_applications
         SET new_number_of_jobs=3
         WHERE number_of_jobs='3 or more jobs';
--- a/db/migrate/20171018174809_number_of_jobs_to_integer.rb
+++ b/db/migrate/20171018174809_number_of_jobs_to_integer.rb
@@ -1,0 +1,23 @@
+class NumberOfJobsToInteger < ActiveRecord::Migration[5.1]
+  def up
+    add_column :medicaid_applications, :new_number_of_jobs, :integer
+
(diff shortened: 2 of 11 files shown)


jessieay:
[[Medicaid Flow - Job number] Update current job numbers to 4 or more jobs](https://trello.com/c/oHtvWVDj/458-medicaid-flow-job-number-update-current-job-numbers-to-4-or-more-jobs)

rachelcope:
💯

app/controllers/medicaid/income_job_number_controller.rb
```diff
@@ -4,6 +4,22 @@ module Medicaid
...
+    def existing_attributes
+      HashWithIndifferentAccess.new(number_of_job_attributes)
+    end
+
+    def number_of_job_attributes
+      { new_number_of_jobs: number_of_jobs }
+    end
+
+    def number_of_jobs
+      if current_application.new_number_of_jobs&. > 4
```
jayroh: `current_application.new_number_of_jobs&. > 4`

Yoooooo. Lonely operator on the `>`?? Look at you PFFT

db/migrate/20171018174809_number_of_jobs_to_integer.rb
```diff
@@ -0,0 +1,23 @@
...
+        UPDATE medicaid_applications
+        SET new_number_of_jobs=1
+        WHERE number_of_jobs='1 job';
+        UPDATE medicaid_applications
+        SET new_number_of_jobs=2
+        WHERE number_of_jobs='2 jobs';
+        UPDATE medicaid_applications
+        SET new_number_of_jobs=3
+        WHERE number_of_jobs='3 or more jobs';
+      SQL
```
jayroh: Lost the semicolons and thought this was one long query. 

Thoughts on splitting them up into several `executes "..."`? Or some newlines in between?

spec/controllers/medicaid/income_job_number_controller_spec.rb
```diff
@@ -1,13 +1,33 @@
 require "rails_helper"
 
-RSpec.describe Medicaid::IncomeJobNumberController, type: :controller do
+RSpec.describe Medicaid::IncomeJobNumberController do
+  let(:step) { assigns(:step) }
```
jayroh: Is there a down side to moving this directly in the context of the test in line 15? Since it's only used in one test I think it should be able to be housed in just that one test

app/controllers/medicaid/income_job_number_controller.rb
jessieay: oh yeah I actually got an error in my specs when I didnt have this!

spec/controllers/medicaid/income_job_number_controller_spec.rb
jessieay: I was getting a `nil` error when calling a method on on `step` when instantiating this in the spec itself (that's what I started with)

putting it in a `let` did some kind of magic that fixed that. Do you know where in a controller spec is ideal to assign an `assigns` ?

db/migrate/20171018174809_number_of_jobs_to_integer.rb
jessieay: sure! done c38c51e

jayroh:
fwiw I like the defaulting to 4 on the second page

spec/controllers/medicaid/income_job_number_controller_spec.rb
jessieay: Done at e97f99c

Turns out you need to assign `assigns` after the `get` in the spec

PR #364, opened 2017-10-18
Add feature spec for single-member medicaid flow

jayroh:


jayroh:
[Add feature spec for medicaid flow](https://trello.com/c/ipRf6Fjf/394-add-feature-spec-for-medicaid-flow)

spec/features/medicaid_spec.rb
```diff
@@ -1,8 +1,142 @@
...
+    on_pages "Current Income" do
+      expect(page).to have_content("Do you currently have a job?")
+      click_on "Yes"
+
+      expect(page).to have_content("Tell us how many jobs you currently have.")
+      choose "1 job"
+      click_on "Next"
+
+      expect(page).to have_content("Are you self-employed?")
+      click_on "No"
```
jessieay: Don't have to do this now, but should we have a test that says "Yes" to all so that we see all possible screens?

spec/features/medicaid_spec.rb
```diff
@@ -1,8 +1,142 @@
...
+
+    on_pages "Income & Expense Amounts" do
+      click_on "Next"
+
+      fill_in "Your Current Job", with: 100
+      fill_in "Your Unemployment", with: 100
+      click_on "Next"
+    end
+
+    on_pages "Contact Information & Followup" do
```
jessieay: we usually call `on_page` for each new page because that method checks that you are on the correct page, which is good for debugging. If we call around a group is it doing that for each visited page?

jayroh: It's not doing it for *each* page. But since these pages' titles aren't ever going to change I thought it would be sufficient to group them. If we want to test *each* page then a view spec with some sort of terse, concise test on the title might be a good idea.

As a run through these pages, however, getting a feel for the interaction seems to be the most important part. Making sure the clicks take us to where they should is the most important part. IMHO.

What do you think?

jessieay: Meh, I am ok with either way. If it becomes a problem we can always add later

PR #365, opened 2017-10-19
Add income inputs for each job

jessieay:
* People can have multiple jobs, now they can report income for each job
individually.


jessieay:
[Add multiple input fields for income earned when a user indicates they work multiple jobs](https://trello.com/c/elfF2Xt6/419-add-multiple-input-fields-for-income-earned-when-a-user-indicates-they-work-multiple-jobs)

PR #366, opened 2017-10-19
Skip pregnancy question if primary client is male

jayroh:


jayroh:
[[Medicaid Flow] Only ask if the user applicant has been pregnant if they indicated they are female](https://trello.com/c/GNNQ98EZ/414-medicaid-flow-only-ask-if-the-user-applicant-has-been-pregnant-if-they-indicated-they-are-female)

app/controllers/medicaid/health_pregnancy_controller.rb
```diff
@@ -2,5 +2,14 @@
...
 module Medicaid
   class HealthPregnancyController < MedicaidStepsController
+    private
+
+    def skip?
+      client_is_male?
+    end
+
+    def client_is_male?
+      current_application.primary_member.sex == "Male"
```
jessieay: what if there are multiple members and one is female?

jayroh: Great idea

PR #367, opened 2017-10-19
Log info about email to both export & logger

jayroh:
Reason for Change
===================

When an export is executed, the block we pass to it can accept a logger
object which, when written/outputted to, will save to the export object
AND log to stdout.

Our email job has not been doing this so this commit changes it to do so
with not only the receiving office, but the id of the client who is
submitting it.


jayroh:
[Add further information to, and properly log, emails that are sent out to field offices.](https://trello.com/c/kQ5RBQRR/463-add-further-information-to-and-properly-log-emails-that-are-sent-out-to-field-offices)

PR #371, opened 2017-10-19
Convert all timestamps in admin dashboard to EST.

hartsick:
Customizes a partial for administrate to convert and format any DateTime fields on detail pages to the same way they're shown elsewhere in the app for SNAP applications (using SnapApplication#signed_at_est as guide). End result should be that all full timestamps in the admin dashboard are now shown in EST rather than UTC.

Does not convert Date objects on index pages to EST.

Pages & fields affected:

**On snap_application#show (e.g. admin/snap_applications/1):**

* Created at
* Updated at
* Signed at

**On export#show (e.g. admin/exports/2):**

* Completed at
* Created at
* Updated at

**On member#show (e.g. admin/member/2):**

* Created at
* Updated at

### DIFF
--- a/app/views/fields/date_time/_index.html.erb
+++ b/app/views/fields/date_time/_index.html.erb
@@ -1,0 +1,24 @@
+<%#
+# DateTime Index Partial
+
+This partial renders a datetime attribute,
+to be displayed on a resource's index page.
+
+By default, the attribute is rendered
+as a localized date & time string.
+
+## Local variables:
+
+- `field`:
+  An instance of [Administrate::Field::DateTime][1].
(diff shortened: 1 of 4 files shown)


hartsick:
[Update timestamps in the Admin dashboard to EST](https://trello.com/c/xTnAV5lZ/462-update-timestamps-in-the-admin-dashboard-to-est)

hartsick:
@jessieay It's not a great reason—just that on index, times are handled as Date objects by default so they lack timezone information. Talking through this made me uncomfortable, though, so I'm going to add a commit that makes sure dates are correct for EST. Thanks for asking!

PR #375, opened 2017-10-20
Verify us citizenship across all household members

jayroh:
If the primary client answers that there is someone in the household that
is not a US citizen then they should be taken to a page that asks them
*who* is not a citizen.

### DIFF
--- a/app/controllers/medicaid/citizens_controller.rb
+++ b/app/controllers/medicaid/citizens_controller.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Medicaid
+  class CitizensController < Medicaid::ManyMemberStepsController
+    private
+
+    def skip?
+      everyone_a_citizen?
+    end
+
+    def member_attrs
+      %i[citizen]
+    end
+
+    def everyone_a_citizen?
+      current_application.citizen?
+    end
+  end
+end
--- a/app/controllers/medicaid/citizens_controller.rb
+++ b/app/controllers/medicaid/citizens_controller.rb
@@ -5,7 +5,8 @@ module Medicaid
     private
 
     def skip?
-      everyone_a_citizen?
+      everyone_a_citizen? ||
+        (single_member_not_a_citizen? && update_single_member)
     end
 
     def member_attrs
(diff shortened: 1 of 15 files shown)


app/views/medicaid/intro_citizen/edit.html.erb
```diff
@@ -2,7 +2,9 @@
 
 <div class="form-card">
   <header class="form-card__header">
-    <div class="form-card__title">Are you currently a US Citizen?</div>
+    <div class="form-card__title">
+      Is everyone in your household currently a US citizen?
```
jessieay: do we want to show the other page title if there is just one person in the household? We did that using i18n pluralization for other features

app/controllers/medicaid/citizens_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    def skip?
+      everyone_a_citizen?
+    end
+
+    def member_attrs
+      %i[citizen]
+    end
+
+    def everyone_a_citizen?
+      current_application.citizen?
```
jessieay: while are are here, does it make sense to rename the `citizen` column to `everyone_a_citizen` on medicaid apps? that's the name it has on on snap apps and we've been trying to standardize as we go when possible

spec/features/medicaid_application_with_multiple_members_spec.rb
```diff
@@ -48,7 +48,9 @@
     end
 
     on_page "Introduction" do
-      expect(page).to have_content("Are you currently a US Citizen?")
+      expect(page).to have_content(
+        "Is everyone in your household currently a US citizen",
+      )
       click_on "Yes"
```
jessieay: Maybe we should say `no` here so we have a chance to indicate who is *not* a citizen in a feature spec?

spec/controllers/medicaid/citizens_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+          )
+        session[:medicaid_application_id] = medicaid_application.id
+
+        get :edit
+
+        expect(response).to redirect_to(subject.next_path)
+      end
+    end
+
+    context "someone in the home is not a use citizen" do
```
jessieay: typo `use` -> `US`

spec/controllers/medicaid/intro_citizen_controller_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "rails_helper"
+
+RSpec.describe Medicaid::IntroCitizenController, type: :controller do
+  describe "#next_path" do
```
jessieay: may want to look at the specs for `IntroCollegeController` -- we make sure that the member for a single member household is updated via this controller, which I think we should be doing for all single member households on these "yes/no" pages.

jayroh:
[[Medicaid MM Flow] Intro: Is everyone in your household currently a US Citizen? + Nested](https://trello.com/c/zouJo1sW/432-medicaid-mm-flow-intro-is-everyone-in-your-household-currently-a-us-citizen-nested)

PR #377, opened 2017-10-23
Redirect to office-specific landing pages after SNAP submitted

jessieay:
**WHY**: This is a better UX and will prevent people from re-applying
from the wrong page when testing the app when in offices.


jessieay:
[Applications started, and then completed, via an office-specific landing page [Clio, Union] they should end up redirecting *back* to the same office-specific landing page](https://trello.com/c/NhFUDQjH/468-applications-started-and-then-completed-via-an-office-specific-landing-page-clio-union-they-should-end-up-redirecting-back-to-th)

PR #380, opened 2017-10-24
Add employed / job number steps for multi-member flow

jessieay:
screen shot 2017-10-23 at 5 34 53 pm

* This is mostly complete but will required a follow up PR to:
- update employment_status, employed_job_number attrs on primary member when only one member present
- remove job_number field from medicaid_applicatiomns table
- rename `employed` on medicaid_applications, since that column is meant to indicate whether *anyone* is employed

### DIFF
--- a/app/controllers/medicaid/income_job_number_continued_controller.rb
+++ b/app/controllers/medicaid/income_job_number_continued_controller.rb
@@ -5,11 +5,17 @@ module Medicaid
     private
 
     def skip?
-      not_employed? || current_application.number_of_jobs < 4
+      multi_member_household? ||
+        not_employed? ||
+        current_application.number_of_jobs < 4
     end
 
     def not_employed?
       !current_application.employed?
+    end
+
+    def multi_member_household?
+      current_application.members.count != 1
     end
   end
 end
--- a/app/controllers/medicaid/income_job_number_continued_controller.rb
+++ b/app/controllers/medicaid/income_job_number_continued_controller.rb
@@ -13,9 +13,5 @@ module Medicaid
     def not_employed?
       !current_application.employed?
     end
-
-    def multi_member_household?
-      current_application.members.count != 1
-    end
   end
 end
--- a/app/controllers/medicaid/income_job_number_controller.rb
+++ b/app/controllers/medicaid/income_job_number_controller.rb
@@ -24,10 +24,6 @@ module Medicaid
       multi_member_household? || not_employed?
     end
 
-    def multi_member_household?
-      current_application.members.count != 1
-    end
-
     def not_employed?
       !current_application&.employed?
     end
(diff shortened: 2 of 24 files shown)


jessieay:
[[Medicaid MM Flow] Current Income: Does anyone in your household currently have a job? + Nested](https://trello.com/c/Mz9hOKuQ/439-medicaid-mm-flow-current-income-does-anyone-in-your-household-currently-have-a-job-nested)

spec/support/feature_helper.rb
```diff
@@ -28,5 +28,11 @@ def on_page(page_title)
     yield
   end
 
+  def select_job_number(full_name:, job_number:)
```
hartsick: nice helper!

app/controllers/medicaid/income_job_number_controller.rb
```diff
@@ -21,7 +21,11 @@ def number_of_jobs
     end
 
     def skip?
-      not_employed?
+      multi_member_household? || not_employed?
+    end
+
+    def multi_member_household?
```
hartsick: we have a similar method in another controller (thinking single_member_household). wonder if we want to just move these up a few levels, maybe just into the StepsController, since they seem generally useful?

app/controllers/medicaid/income_job_number_continued_controller.rb
```diff
@@ -5,11 +5,17 @@ class IncomeJobNumberContinuedController < MedicaidStepsController
...
+      multi_member_household? ||
+        not_employed? ||
+        current_application.number_of_jobs < 4
     end
 
     def not_employed?
       !current_application.employed?
     end
+
+    def multi_member_household?
```
hartsick: I've seen in our current code that there are places where single_member_household is defined as `current_application.members.empty? || current_application.members.count == 1`. I'm not sure how we would get an application without any members (maybe only in tests?), but am thinking we should standardize on single member as either 1 member or 0-1 members.

jessieay: yeah I agree! maybe we decide on one set of logic put the method in `ApplicationController`?

spec/support/feature_helper.rb
jessieay: we have some similar ones in `SnapApplicationFormHelper` if you ever need some clever form methods :)

app/controllers/medicaid/income_job_number_controller.rb
jessieay: agree! see my other comment re: `ApplicationController` ?

app/controllers/medicaid/income_job_number_continued_controller.rb
jessieay: what do you think: should we go for 1 or 0?

spec/support/feature_helper.rb
hartsick: this is good to know

app/controllers/medicaid/income_job_number_continued_controller.rb
jessieay: set to `== 1` and moved to paren controller at b0005974

PR #381, opened 2017-10-24
Ch multi current insurance

hartsick:
Add a multi-member flow for entering information about additional insurance.

This PR ended up being a big one, and I'd love input! I used a lot of the patterns we used before, but kept the insurance type selection in one controller (rather than splitting for single member and multi member). This might be something to refactor—there's a good amount of view logic and some potentially off language for the single-member flow as a result. But, wanted to get this up for review!

Here are the changes, with expected behavior:

**/steps/medicaid/insurance-current**
Asks whether any household members seeking insurance are currently insured 

If single-member, updates `anyone_is_insured` accordingly on primary member.
If multi-member, continues to page to select who is insured.

**/steps/medicaid/insurance-current-member**
Asks who is currently insured. Only shows people seeking insurance.
Shows error if no members are selected.

**/steps/medicaid/insurance-current-type**
For each person requesting insurance with a current insurance plan, asks which kind of insurance they have. The form is currently all on one page (iterate over the eligible members), which we'll want to change going forward—see note below.

And here are some behind-the-scenes changes:
* Moves `insurance_type` and copies `insured` as `is_insured` from `medicaid_applications` to `members` table
* Renames `insured` to `anyone_is_insured` on `medicaid_applications`

screen shot 2017-10-24 at 6 14 38 pm

screen shot 2017-10-24 at 6 14 45 pm

screen shot 2017-10-24 at 6 22 04 pm

screen shot 2017-10-24 at 6 14 59 pm

screen shot 2017-10-24 at 6 15 04 pm

Because this ticket was already fairly big, I punted on breaking the healthcare plan selection into one page per member, which was requested by design. As a result, we'll still need to break out selecting healthcare for multiple members to individual pages, rather than having all members listed on one like it is now (`steps/medicaid/insurance-current-type`). I added a separate ticket for this in Trello: https://trello.com/c/NnQdaKae/434-medicaid-mm-flow-health-coverage-needs-what-type-of-plan-is-member-enrolled-in

### DIFF
--- a/app/controllers/medicaid/insurance_current_controller.rb
+++ b/app/controllers/medicaid/insurance_current_controller.rb
@@ -2,25 +2,16 @@
 
 module Medicaid
   class InsuranceCurrentController < MedicaidStepsController
-    def update
-      @step = step_class.new(step_params)
+    private
 
-      if @step.valid?
-        if single_member_household?
-          current_application.primary_member.update!(member_attrs)
-        end
-
-        current_application.update!(step_params)
-        redirect_to(next_path)
-      else
-        render :edit
+    def after_successful_update_hook
+      if single_member_household?
+        current_application.primary_member.update!(member_attrs)
       end
     end
 
-    private
-
     def member_attrs
-      { is_insured: step_params[:anyone_is_insured] }
+      { insured: step_params[:anyone_insured] }
     end
   end
 end
--- a/app/controllers/medicaid/insurance_current_controller.rb
+++ b/app/controllers/medicaid/insurance_current_controller.rb
@@ -2,5 +2,25 @@
 
 module Medicaid
   class InsuranceCurrentController < MedicaidStepsController
+    def update
+      @step = step_class.new(step_params)
+
+      if @step.valid?
+        if single_member_household?
+          current_application.primary_member.update!(member_attrs)
+        end
+
+        current_application.update!(step_params)
+        redirect_to(next_path)
+      else
+        render :edit
+      end
+    end
+
+    private
+
+    def member_attrs
+      { is_insured: step_params[:anyone_is_insured] }
+    end
   end
 end
--- a/app/controllers/medicaid/insurance_current_type_controller.rb
+++ b/app/controllers/medicaid/insurance_current_type_controller.rb
@@ -5,10 +5,6 @@ module Medicaid
     private
 
     def skip?
-      not_insured?
-    end
-
-    def not_insured?
       current_application.nobody_insured?
     end
 
--- a/app/controllers/medicaid/insurance_current_type_controller.rb
+++ b/app/controllers/medicaid/insurance_current_type_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Medicaid
-  class InsuranceCurrentTypeController < MedicaidStepsController
+  class InsuranceCurrentTypeController < Medicaid::ManyMemberStepsController
     private
 
     def skip?
@@ -9,7 +9,11 @@ module Medicaid
     end
 
     def not_insured?
-      !current_application&.insured?
+      current_application.nobody_insured?
+    end
+
+    def member_attrs
+      %i[insurance_type]
     end
   end
 end
--- a/app/steps/medicaid/insurance_current_type.rb
+++ b/app/steps/medicaid/insurance_current_type.rb
@@ -7,12 +7,12 @@ module Medicaid
       :members,
     )
 
-    def insured_members
-      members.select(&:is_insured)
+    def insured_members_requesting_insurance
+      members.select(&:requesting_health_insurance).select(&:is_insured)
     end
 
     def valid?
-      if insured_members.all? { |m| m.insurance_type.present? }
+      if requesting_members_valid?
         true
       else
         errors.add(
@@ -25,8 +25,14 @@ module Medicaid
 
     private
 
+    def requesting_members_valid?
+      insured_members_requesting_insurance.all? do |m|
+        m.insurance_type.present?
+      end
+    end
+
     def error_message
-      if members.select(&:is_insured).count == 1
+      if insured_members_requesting_insurance.count == 1
         "Please select a plan"
       else
         "Please select a plan for each person"
--- a/app/steps/medicaid/insurance_current_type.rb
+++ b/app/steps/medicaid/insurance_current_type.rb
@@ -1,41 +1,26 @@
 # frozen_string_literal: true
 
 module Medicaid
-  class InsuranceCurrentType < Step
+  class InsuranceCurrentType < ManyMembersStep
     step_attributes(
       :insurance_type,
       :members,
     )
 
     def insured_members_requesting_insurance
-      members.select(&:requesting_health_insurance).select(&:is_insured)
-    end
-
-    def valid?
-      if requesting_members_valid?
-        true
-      else
-        errors.add(
-          :insurance_type,
-          error_message,
-        )
-        false
-      end
+      members.select(&:requesting_health_insurance).select(&:insured)
     end
 
     private
 
-    def requesting_members_valid?
-      insured_members_requesting_insurance.all? do |m|
-        m.insurance_type.present?
-      end
-    end
-
-    def error_message
-      if insured_members_requesting_insurance.count == 1
-        "Please select a plan"
-      else
-        "Please select a plan for each person"
+    def validate_household_member(member)
+      if member.requesting_health_insurance? && member.insured?
+        unless member.insurance_type.present?
+          member.errors.add(
+            :insurance_type,
+            "Please select a plan",
+          )
+        end
       end
     end
   end
--- a/app/steps/medicaid/insurance_current_type.rb
+++ b/app/steps/medicaid/insurance_current_type.rb
@@ -7,15 +7,29 @@ module Medicaid
       :members,
     )
 
+    def insured_members
+      members.select(&:is_insured)
+    end
+
     def valid?
-      if members.select(&:insurance_type).any?
+      if insured_members.all? { |m| m.insurance_type.present? }
         true
       else
         errors.add(
           :insurance_type,
-          "Please select a plan",
+          error_message,
         )
         false
+      end
+    end
+
+    private
+
+    def error_message
+      if members.select(&:is_insured).count == 1
+        "Please select a plan"
+      else
+        "Please select a plan for each person"
       end
     end
   end
--- a/app/views/medicaid/insurance_current_member/edit.html.erb
+++ b/app/views/medicaid/insurance_current_member/edit.html.erb
@@ -7,24 +7,24 @@
     </div>
     <% if @step.members_not_needing_insurance.present? %>
       <p class="text--help text--centered">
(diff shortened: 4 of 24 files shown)


db/migrate/20171020215318_move_insurance_info_to_member.rb
```diff
@@ -0,0 +1,65 @@
...
+            is_insured=medicaid_applications.insured,
+            insurance_type=medicaid_applications.insurance_type
+          FROM (
+            SELECT
+              min(id) as id, benefit_application_id
+            FROM 
+              members
+            WHERE 
+              benefit_application_type = 'MedicaidApplication'
+            GROUP BY benefit_application_id
```
jayroh: This internal/sub query - actually probably with the larger query in question too - I think it'd be super beneficial to have a brief comment (above the `SELECT`) describing the what and why for the query. Where am I coming from with that request? I can get by with enough sql knowledge to figure things out but reading someone _else's_ queries can sometimes be a tall order. So, comments in more involved queries are especially helpful. I'll only speak for myself here but I don't think I'm alone(?)

jayroh: From what I can tell from the context around this, though - we're grouping members by the application, and returning only the "primary"/lowest ID'd member of that group. Correct?

db/migrate/20171020215318_move_insurance_info_to_member.rb
```diff
@@ -0,0 +1,65 @@
...
+          AS primary_medicaid_members
+          WHERE medicaid_applications.id = primary_medicaid_members.benefit_application_id;
+      SQL
+
+      remove_column :members, :is_insured
+      remove_column :members, :insurance_type
+      remove_column :medicaid_applications, :anyone_is_insured
+    end
+  end
+end
```
jayroh: As far as these migrations go?

SURVEY SAYS:

app/controllers/medicaid/insurance_current_type_controller.rb
```diff
@@ -9,7 +9,7 @@ def skip?
     end
 
     def not_insured?
-      !current_application&.insured?
+      !current_application&.anyone_is_insured?
```
jayroh: What do you think about offloading a liiiittle bit of cognitive friction here and making a `#no_one_is_insured?` method on current application. Unpacking "not anyone is insured" makes my feeble brain pause for longer than it should.

db/migrate/20171020215318_move_insurance_info_to_member.rb
hartsick: Yeah, I totally agree with the comment ... comment! I still have to step through it when I read, and I'm the one who wrote it. min is particularly tricky. You're right in your understanding, but let's make getting there a bit easier.

db/migrate/20171020215318_move_insurance_info_to_member.rb
hartsick: weeeee

app/controllers/medicaid/insurance_current_type_controller.rb
hartsick: agree! was actually just struggling with this when continuing to develop the feature

spec/features/medicaid_application_with_maximum_info_spec.rb
```diff
@@ -111,7 +111,7 @@
       fill_in "Your Unemployment", with: 100
       click_on "Next"
 
-      find(".icon-arrow_back").click
+      click_back
```
jayroh: 🎉

spec/steps/medicaid/insurance_current_type_spec.rb
```diff
@@ -0,0 +1,101 @@
+require "rails_helper"
+
+RSpec.describe Medicaid::InsuranceCurrentType do
+  describe "Validations" do
+    context "each insured household member has an insurance type" do
+      it "is valid" do
+        member = create(:member,
+                        is_insured: true,
```
jessieay: generally prefer booleans to NOT start with `is`

spec/steps/medicaid/insurance_current_type_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        )
+
+        expect(step).not_to be_valid
+      end
+    end
+  end
+
+  describe "#insured_members_needing_insurance" do
+    context "with insured members needing insurance" do
+      it "returns only insured members needing insurance" do
```
jessieay: this is good -- we should be doing this check on *all* steps, perhaps

spec/features/medicaid_application_with_multiple_members_spec.rb
```diff
@@ -68,22 +80,89 @@
...
+      expect(page).to have_content("Jessie Tester")
+      expect(page).to have_content("Christa Tester")
+      expect(page).to have_content(
+        "You've already indicated that Joel Tester is not in need of" \
+        " additional health coverage.",
+      )
+      click_on "Next"
+    end
+
+    on_page "Health Coverage Needs" do
```
jessieay: these "on_page" blocks can probably be combined

spec/controllers/medicaid/insurance_current_type_controller_spec.rb
```diff
@@ -15,7 +17,9 @@
 
     context "client is not insured" do
       it "redirects to next step" do
-        medicaid_application = create(:medicaid_application, insured: false)
+        medicaid_application = create(:medicaid_application,
+                                      anyone_is_insured: false,
```
jessieay: could also be `anyone_insured`

db/seeds.rb
```diff
@@ -279,3 +279,42 @@
...
+)
+
+complete_medicaid_application.update!(
+  members: [
+    complete_medicaid_primary,
+    complete_medicaid_secondary,
+  ],
+)
+
+puts "A more complete medicaid application created (or found) " \
```
jessieay: yay! 🤚

config/locales/en.yml
```diff
@@ -1,3 +1,24 @@
+# Files in the config/locales directory are used for internationalization
```
jessieay: we can keep this removed of you don't mind

app/views/medicaid/insurance_current_type/edit.html.erb
```diff
@@ -2,21 +2,28 @@
 
 <div class="form-card">
   <header class="form-card__header">
-    <div class="form-card__title">What type of insurance plan are you currently enrolled in?</div>
+    <div class="form-card__title">
+      <%= t("medicaid.insurance_current_type.edit.title",
```
jessieay: this is used for both single-member and multi-member. right now you can get a message that says "Tell us what insurance plan each person is enrolled in." even if only one member is effectively applying.

app/views/medicaid/insurance_current_member/edit.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+<div class="form-card">
+  <header class="form-card__header">
+    <div class="form-card__title">
+      Tell us who is currently enrolled in a health insurance plan.
+    </div>
+    <% if @step.members_not_needing_insurance.present? %>
+      <p class="text--help text--centered">
+        You've already indicated that <%=@step.members_not_needing_insurance.map(&:full_name).to_sentence %>
+        <%= @step.members_not_needing_insurance.count == 1 ? "is" : "are" %>
```
jessieay: could define a custom inflection and use `pluralize` here https://stackoverflow.com/questions/3517989/ruby-on-rails-how-do-you-explicitly-define-plural-names-and-singular-names-in-r

but maybe not worth it

another option:

- throw the whole string into a localization with a `one` and `other` option
- pass in `@step.members_not_needing_insurance.map(&:full_name).to_sentence` as a variable to that localization

app/steps/medicaid/insurance_current_type.rb
```diff
@@ -2,6 +2,41 @@
...
+    private
+
+    def requesting_members_valid?
+      insured_members_requesting_insurance.all? do |m|
+        m.insurance_type.present?
+      end
+    end
+
+    def error_message
+      if insured_members_requesting_insurance.count == 1
```
jessieay: could also use i18n with `count`

app/steps/medicaid/insurance_current_type.rb
```diff
@@ -2,6 +2,41 @@
...
+          error_message,
+        )
+        false
+      end
+    end
+
+    private
+
+    def requesting_members_valid?
+      insured_members_requesting_insurance.all? do |m|
```
jessieay: this seems similar to `ManyMembersStep`, may consider inheriting from that

app/controllers/medicaid/insurance_current_type_controller.rb
```diff
@@ -1,15 +1,19 @@
 # frozen_string_literal: true
 
 module Medicaid
-  class InsuranceCurrentTypeController < MedicaidStepsController
+  class InsuranceCurrentTypeController < Medicaid::ManyMemberStepsController
     private
 
     def skip?
       not_insured?
```
jessieay: could put `current_application.nobody_insured?` here, maybe be easier to read

app/controllers/medicaid/insurance_current_controller.rb
```diff
@@ -2,5 +2,25 @@
 
 module Medicaid
   class InsuranceCurrentController < MedicaidStepsController
+    def update
+      @step = step_class.new(step_params)
+
+      if @step.valid?
+        if single_member_household?
```
jessieay: this whole method is a pattern we've implemented at least 3 times now, perhaps should start to think about extracting

jessieay: and now we did extract! woop woop

hartsick: ✅

app/steps/medicaid/insurance_current_type.rb
hartsick: ✅

app/steps/medicaid/insurance_current_type.rb
hartsick: ✅

app/views/medicaid/insurance_current_type/edit.html.erb
hartsick: (that comment is from me ^, and more a thing to be aware of but not change right now!)

spec/controllers/medicaid/insurance_current_type_controller_spec.rb
hartsick: ✅

spec/steps/medicaid/insurance_current_type_spec.rb
hartsick: ✅

spec/steps/medicaid/insurance_current_type_spec.rb
hartsick: this is a good point

config/locales/en.yml
hartsick: ✅

app/views/medicaid/insurance_current_member/edit.html.erb
hartsick: ✅

PR #382, opened 2017-10-24
Update employment status on member for single member households

jessieay:
**WHY**"
- when we just has single member, we saved all data on the app itself
- recently added multi-member via https://github.com/codeforamerica/michigan-benefits/pull/380
- now we want to save this data on the member record for single member
  households as well (for consistency)


spec/controllers/medicaid/income_expenses_job_number_controller_spec.rb
```diff
@@ -1,33 +0,0 @@
-require "rails_helper"
-
-RSpec.describe Medicaid::IncomeJobNumberController do
```
jessieay: because of weird naming, this was a duplicate spec file (we also have specs for this controller in the correct location)

jessieay:
[[Medicaid MM Flow] Current Income: Does anyone in your household currently have a job? + Nested](https://trello.com/c/Mz9hOKuQ/439-medicaid-mm-flow-current-income-does-anyone-in-your-household-currently-have-a-job-nested)

app/controllers/medicaid/amounts_income_controller.rb
```diff
@@ -18,19 +18,23 @@ def income_attributes
...
+
     def skip?
-      not_employed? && not_self_employed? && not_receiving_unemployment?
+      nobody_employed? && not_self_employed? && not_receiving_unemployment?
     end
 
-    def not_employed?
-      !current_application&.employed?
+    def nobody_employed?
+      !current_application&.anyone_employed?
```
jayroh: Was just mentioning to @hartsick that it might make it easier on us to consider implementing these as first class methods (ie: `#no_one_employed?`) on the current_application instead of negating the predicate (not anyone employed).

On that note - if we continue to do that and see that there are a LOT of predicate methods defined on the application object we should think about extracting them to some sort of policy object.

jessieay: is the main reason for implementing these that it will reduce duplication or is there an additional reason?

jayroh: Adding methods like `no_one_employed`? I suggest that purely for ease of comprehension. It's easier for me to unpack "no one is employed" instead of "not anyone employed".  The "no one" versus "not anyone" namely - just an english / legibility thing.

PR #383, opened 2017-10-24
Ask if, and who, has been pregnant recently

jayroh:
* If everyone in the household is a male, skip the questions entirely.
* If no one has been pregnant, skip the "who has been pregnant" question.
* If someone has been pregnant, and there's only one female, mark as new
  mom and skip the "who has been pregnant" question.
* If someone has been pregnant, and there is more than one female, ask
  the "who has been pregnant" question.


app/controllers/medicaid/health_recent_pregnancy_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    def only_one_female?
+      current_application.members.select(&:female?).count == 1
+    end
+
+    def application_has_a_new_mom?
+      current_application.new_mom?
+    end
+
+    def update_primary_female_member
+      current_application.primary_member.update(new_mom: true)
```
jessieay: it feels a bit unexpected for this to be located in the `skip` action

app/controllers/medicaid/health_recent_pregnancy_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    def single_female_member_is_pregnant?
+      only_one_female? && application_has_a_new_mom?
+    end
+
+    def only_one_female?
+      current_application.members.select(&:female?).count == 1
+    end
+
+    def application_has_a_new_mom?
+      current_application.new_mom?
```
jessieay: thoughts on renaming to`anyone_new_mom` to be consistent with snap app naming?

app/controllers/medicaid/health_recent_pregnancy_controller.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Medicaid
+  class HealthRecentPregnancyController < Medicaid::ManyMemberStepsController
+    private
+
+    def skip?
+      no_pregnancies? ||
+        only_males? ||
+        (single_female_member_is_pregnant? && update_primary_female_member)
```
jessieay: the way we've been handling this in other "multi member" controllers is:

- the single member us updated via the "yes/no" controller
-  we skip the follow on controller for single member households. 

thoughts on the benefits and drawbacks of that approach vs this one?

app/controllers/medicaid/health_recent_pregnancy_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def only_males?
+      current_application.members.select(&:female?).empty?
+    end
+
+    def single_female_member_is_pregnant?
+      only_one_female? && application_has_a_new_mom?
+    end
+
+    def only_one_female?
```
jessieay: oooh interesting edge case--- good catch!

app/controllers/medicaid/many_member_steps_controller.rb
```diff
@@ -32,9 +32,9 @@ def assign_household_member_attributes
     def step
       @step ||= step_class.new(
         current_application.
-        attributes.
-        slice(*step_attrs).
-        merge(members: current_application.members),
+          attributes.
```
jessieay: is this an intentional change?

app/models/member.rb
```diff
@@ -57,6 +57,14 @@ def not_employed?
     employment_status == "not_employed"
   end
 
+  def female?
+    sex == "female"
```
jessieay: should we add a validation on `Member` that `sex` is one of these two values? right now that validation happens at the step level and I'd be worried about it being `Female` instead of `female` or something like that

app/steps/medicaid/step_navigation.rb
```diff
@@ -24,6 +24,7 @@ class StepNavigation
       "Health" => [
         Medicaid::HealthDisabilityController,
         Medicaid::HealthPregnancyController,
+        Medicaid::HealthRecentPregnancyController,
```
jessieay: I think in general we've been following a pattern that would have this controller named `HealthPregnancyMemberController`

app/controllers/medicaid/health_recent_pregnancy_controller.rb
jayroh: After looking how it's implemented in yours and Christa's recent PR's I came in here to say that the approach I took is inconsistent. So I'll update it (and a previous pr as well)

app/controllers/medicaid/health_recent_pregnancy_controller.rb
jayroh: Sure

app/controllers/medicaid/many_member_steps_controller.rb
jayroh: It was intentional, yeah.

app/models/member.rb
jayroh: Yeah, we should. I think I've seen it capitalized in other places.

app/steps/medicaid/step_navigation.rb
jayroh: k

jessieay:
[[Medicaid MM Flow] Quick Health Questions: Has anyone been pregnant recently? + Nested](https://trello.com/c/owfHPXeO/437-medicaid-mm-flow-quick-health-questions-has-anyone-been-pregnant-recently-nested)

PR #387, opened 2017-10-25
Add benefit label label to header

rachelcope:
* Add "Medicaid" or "Food Assistant Application" to header
* Visual reminder for users coming from `/dual-index` on which application they are completing. 

Note to engineers: The label content will need to change based on the application.

image

image


rachelcope:
@jayroh great! yes, it is "Medicaid Application" and "Food Assistance Application"

jayroh:
on it, @rachelcope

jayroh:
@jessieay @hartsick putting this back up for re-review now that I've added a commit.

PR #391, opened 2017-10-25
Refactor update function for single member to use after_successful_update_hook

hartsick:
...to reduce code duplication.


jayroh:
Siiiiiiiiick

PR #393, opened 2017-10-25
Instead of defaulting to a NullStep, raise error

jayroh:
... with a descriptive reason as to WHAT the step should be.

Also update the controllers to override the step_class with NullStep when
necessary.


adarsh:
image

https://twitter.com/adarshp/status/690027192756076544

PR #394, opened 2017-10-25
Use class methods instead of constants for driving code

jessieay:
* These constants were raising an `already initialized constant.`
warning, which was irritating
* Avdi Grimm suggests that we should prefer class methods to constants,
generally: http://www.virtuouscode.com/2011/08/18/do-we-need-constants/

What do you think?

### DIFF
--- a/db/migrate/20171025184755_add_employed_as_boolean_to_medicaid.rb
+++ b/db/migrate/20171025184755_add_employed_as_boolean_to_medicaid.rb
@@ -9,5 +9,4 @@ class AddEmployedAsBooleanToMedicaid < ActiveRecord::Migration[5.1]
       remove_column :members, :employed
     end
   end
-
 end
--- a/lib/mi_bridges/driver.rb
+++ b/lib/mi_bridges/driver.rb
@@ -122,7 +122,7 @@ module MiBridges
     end
 
     def complete_application
-      while page_title != SubmitPage::TITLE
+      while page_title != SubmitPage.title
         klass = find_page_klass
         run_flow([klass])
       end
@@ -192,8 +192,8 @@ module MiBridges
 
     def find_page_klass
       klass = APPLY_FLOW.detect do |page|
-        page_title == page::TITLE ||
-          page_title.match?(page::TITLE)
+        page_title == page.title ||
+          page_title.match?(page.title)
       end
 
       if klass.nil?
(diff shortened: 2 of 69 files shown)


adarsh:
I like what Avdi has to say here - thoughtful, as ever and it makes sense to me.

I don't have a preference myself, maybe slightly for constants out of habit. Some other perspectives:

* The constant is the most performant (benchmarking [here](https://gist.github.com/benoittgt/ee69d941ac2158d47f701656b58ad99c))
* [Class methods can resist refactoring](http://blog.codeclimate.com/blog/2012/11/14/why-ruby-class-methods-resist-refactoring/)

jessieay:
@adarsh ideas on how to get rid of the `already initialized constant` error? I agree that constants are good in many cases but in this app the number of errors that are output due to them is distracting.

to see what I mean, cd into the app and run `rails g migration test_migration` -- all of that output!

jayroh:
I'm 50/50 on this but I think I'd land on using the class methods. The performance aspect is a compelling argument _IF_ the context of these classes wasn't in the realm of selenium driving. The latency there is network so the ruby side of things shaving some speed off isn't what's going to hurt us. The class method + refactoring side of things won't be much a concern because there's no interaction with other methods. These are only acting as a gateway to a string literal.

On the other hand, yeah - just by sheer force of habit, prefer constants.

I think maybe ruby 2.5 is going to address things around this? https://blog.bigbinary.com/2017/10/18/ruby-2.5-has-removed-top-level-constant-lookup.html

Not sure if that has any direct relationship to the warnings we see in our tests.

Also, damn these tests really are noisy.

jessieay:
@jayroh perhaps that will fix. The exact error we see here is `already initialized constant`, which happens every time we reference an existing constant. Basically, I think that each time it looks for the constant it is doing re-assignment, so we get that error.

Found some SO posts on the topic, but not good solutions:
https://stackoverflow.com/questions/7953611/solutions-to-the-annoying-warning-already-initialized-constant-message

this is a non-class method solution: https://stackoverflow.com/a/23552366/1019369

preferences?

jayroh:
> this is a non-class method solution: https://stackoverflow.com/a/23552366/1019369

I don't think this would work as it would pick up the value of previously defined TITLE constants.

adarsh:
I can't find support for the following theory / ghost story but I feel like I've only seen this on projects with inherited classes. My hypothesis has been that the code sharing leads to re-initialization of constants but I have no proof this is the case. 👻👻👻 

IMHO, class methods to get rid of this warning is a totally reasonable step to take and I agree with @jayroh's assessment on performance impact.

jessieay:
@jayroh @adarsh merged! but always happy to revert this commit if it seems like this is hurting more than helping

PR #396, opened 2017-10-26
Question re: disabilities open to multiple members

jayroh:
When asking a household of multiple members the question regarding
disabilities will now ask IF and WHO is disabled in the household.

### DIFF
--- a/app/controllers/medicaid/health_disability_controller.rb
+++ b/app/controllers/medicaid/health_disability_controller.rb
@@ -2,5 +2,16 @@
 
 module Medicaid
   class HealthDisabilityController < MedicaidStepsController
+    private
+
+    def after_successful_update_hook
+      if single_member_household?
+        current_application.primary_member.update!(member_attrs)
+      end
+    end
+
+    def member_attrs
+      { disabled: step_params[:anyone_disabled] }
+    end
   end
 end
--- a/app/controllers/medicaid/health_disability_member_controller.rb
+++ b/app/controllers/medicaid/health_disability_member_controller.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Medicaid
+  class HealthDisabilityMemberController < Medicaid::ManyMemberStepsController
+    private
+
+    def skip?
+      nobody_disabled? || single_member_household?
+    end
+
+    def member_attrs
+      %i[disabled]
+    end
+
+    def nobody_disabled?
+      !current_application.anyone_disabled?
+    end
+  end
+end
--- a/app/steps/medicaid/health_disability.rb
+++ b/app/steps/medicaid/health_disability.rb
@@ -2,6 +2,6 @@
 
 module Medicaid
   class HealthDisability < Step
-    step_attributes(:disabled)
+    step_attributes(:anyone_disabled)
   end
 end
--- a/app/steps/medicaid/health_disability_member.rb
+++ b/app/steps/medicaid/health_disability_member.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Medicaid
+  class HealthDisabilityMember < ManyMembersStep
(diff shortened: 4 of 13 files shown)


jayroh:
[[Medicaid MM Flow] Quick Health Questions: Does anyone have a disability? + Nested](https://trello.com/c/Nd5VajaP/436-medicaid-mm-flow-quick-health-questions-does-anyone-have-a-disability-nested)

app/controllers/medicaid/health_disability_member_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def skip?
+      no_one_disabled? || single_member_household?
+    end
+
+    def member_attrs
+      %i[disabled]
+    end
+
+    def no_one_disabled?
```
jessieay: `nobody_disabled` ?

app/steps/medicaid/health_disability.rb
```diff
@@ -2,6 +2,6 @@
 
 module Medicaid
   class HealthDisability < Step
-    step_attributes(:disabled)
+    step_attributes(:anyone_disabled)
```
jessieay: hooray!

app/steps/medicaid/health_disability_member.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Medicaid
+  class HealthDisabilityMember < ManyMembersStep
+    step_attributes(:members)
+
+    def valid?
+      if members.select(&:disabled).any?
+        true
+      else
+        errors.add(:disabled, "Make sure you select at least one person")
```
jessieay: we have this error string a lot, wonder if we should throw it in i18n for share-ablity at some point

app/steps/medicaid/step_navigation.rb
```diff
@@ -23,6 +23,7 @@ class StepNavigation
       ],
       "Health" => [
         Medicaid::HealthDisabilityController,
+        Medicaid::HealthDisabilityMemberController,
```
jessieay: ❤️  love we that we have this pattern going - feels soooo predictable

db/migrate/20171026150119_rename_disabled_to_anyone_disabled.rb
```diff
@@ -0,0 +1,19 @@
+class RenameDisabledToAnyoneDisabled < ActiveRecord::Migration[5.1]
+  def up
+    add_column :medicaid_applications, :anyone_disabled, :boolean
```
jessieay: should we default to `false` ? can't remember if we do that for other similar fields

app/controllers/medicaid/health_disability_controller.rb
```diff
@@ -2,5 +2,16 @@
 
 module Medicaid
   class HealthDisabilityController < MedicaidStepsController
+    private
+
+    def after_successful_update_hook
+      if single_member_household?
```
jessieay: should we have a  controller spec for this condition?

db/migrate/20171026150119_rename_disabled_to_anyone_disabled.rb
jayroh: I don't think we should and here's why - if we have `nil` for these answers we know they have not been answered. Whereas a false or true will tell us they've been answers. I think it's important to be able to infer if a client has gone through/by that step and if we default to a value, we'd lose that visibility.

app/steps/medicaid/health_disability_member.rb
jayroh: Yeah I think it's a good idea. Worthwhile to go in the backlog:

https://trello.com/c/SvkqRfzJ/500-pull-make-sure-you-select-at-least-one-person-out-from-the-controllers-and-into-locales-enyml

PR #402, opened 2017-10-26
Better prompt when asking about household pregnancy

jayroh:
Because the pregnancy question is specific to how many females are among
the members, the question for this should be catered to several
different contexts, or states:

1. Multiple members. One female. That female is you. - "Are YOU pregnant"
2. Multiple members. One female. That female is someone else - "Is SHE pregnant"
3. Multiple members. Several females. "Is ANYONE pregnant"

When there is a single female in the household that's pregnant we should
not show them the page where we ask who. It's a given that if there's
only one female in the house, and someone is pregnant - then she is that
person.

This commit revises the view and the locale to specify how those
questions start and personalizes them accordingly.

A view spec is added as well so we don't exercise the entirety of the
application in a feature spec.

Is the other member, a female, pregnant?
==============================

quick_health_questions

Are you, the only female, pregnant?
==========================

quick_health_questions

Has anyone, amongst the 2 females in the house, been pregnant?
================================================

quick_health_questions

### DIFF
--- a/spec/views/medicaid/health_pregnancy/edit.html.erb_spec.rb
+++ b/spec/views/medicaid/health_pregnancy/edit.html.erb_spec.rb
@@ -1,0 +1,67 @@
+require "rails_helper"
+
+describe "medicaid/health_pregnancy/edit.html.erb" do
+  before do
+    controller.singleton_class.class_eval do
+      def current_path
+        "/steps/medicaid/health-pregnancy"
+      end
+
+      helper_method :current_path
+    end
+  end
+
+  context "2 members - primary member is male" do
+    it "asks if the only *other* female is pregnant" do
+      app = create(:medicaid_application, anyone_new_mom: true)
+      create(:member, :male, first_name: "Joel", benefit_application: app)
+      create(:member, :female, first_name: "Sara", benefit_application: app)
+      pregnancy_step = Medicaid::HealthPregnancy.new(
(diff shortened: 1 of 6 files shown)


jayroh:
[[Medicaid MM Flow] Quick Health Questions: Has anyone been pregnant recently? + Nested](https://trello.com/c/owfHPXeO/437-medicaid-mm-flow-quick-health-questions-has-anyone-been-pregnant-recently-nested)

spec/views/medicaid/health_pregnancy/edit.html.erb_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      pregnancy_step = Medicaid::HealthPregnancy.new(
+        anyone_new_mom: true,
+        members: app.members,
+      )
+
+      assign(:step, pregnancy_step)
+
+      render
+
+      expect(rendered).to include("Has Sara been pregnant recently?")
```
jessieay: is it problematic that we use the word `recently` rather than `in the last 3 months` ? Seems pretty vague

jayroh: Who's got final word on copy? @rachelcope - what do you think?

adarsh:
I hope this small change has a significant positive effect for our users. 🤰

PR #403, opened 2017-10-26
Part II: "other income" for multi member flow

jessieay:
* Adds page with checkboxes to indicate *who* has other income
* next up is to indicate what type of other income for each member

screen shot 2017-10-26 at 4 23 54 pm


jessieay:
[[Medicaid MM Flow] Current Income: Does anyone in the household get income that's not from a job? + Nested (2)](https://trello.com/c/tHLD1PgG/441-medicaid-mm-flow-current-income-does-anyone-in-the-household-get-income-thats-not-from-a-job-nested-2)

PR #408, opened 2017-10-30
Fix the snap application search

jayroh:
Change some fields to NOT be searchable so that the rest of the fields
will work.

Add some (regression) tests

### DIFF
--- a/spec/controllers/admin/snap_application_controller_spec.rb
+++ b/spec/controllers/admin/snap_application_controller_spec.rb
@@ -1,0 +1,27 @@
+require "rails_helper"
+
+RSpec.describe Admin::SnapApplicationsController do
+  describe "#index" do
+    it "returns error with no proper authentication" do
+      get :index
+
+      expect(response.status).to eq 401
+    end
+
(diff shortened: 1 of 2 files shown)


spec/controllers/admin/snap_application_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      it "returns successfully" do
+        create(:snap_application, signature: "Joel Tester")
+
+        request.env["HTTP_AUTHORIZATION"] =
+          ActionController::HttpAuthentication::Basic.encode_credentials(
+            "admin",
+            "password",
+          )
+
+        get :index, params: { search: "Joel" }
```
hartsick: 👍

jayroh:
@hartsick - for some context here. I think the field types we have in this model are incompatible with the "stock" search capabilities of administrate, which are extremely basic and super-duper rudimentary. The depth of any old search is literally performing `like`s in a query against all available columns.

So, to boil it down - yeah trying a `LIKE` against a boolean or integer column is probably the root cause.

PR #410, opened 2017-10-31
Determine if, and who, pays student loan interest

jayroh:


jayroh:
[[Medicaid MM Flow] Current Expenses: Does anyone in your household pay student loan interest? + Nested](https://trello.com/c/E3L3butG/442-medicaid-mm-flow-current-expenses-does-anyone-in-your-household-pay-student-loan-interest-nested)

app/controllers/medicaid/expenses_student_loan_member_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Medicaid
+  # rubocop:disable LineLength
```
jessieay: in other controllers we've just breaking these up onto 2 lines rather than disabling rubocop

app/controllers/medicaid/expenses_student_loan_controller.rb
```diff
@@ -2,5 +2,19 @@
 
 module Medicaid
   class ExpensesStudentLoanController < MedicaidStepsController
+    private
+
+    def after_successful_update_hook
+      if single_member_household?
```
jessieay: should we have a controller spec that confirms this is working as expected?

app/controllers/medicaid/expenses_student_loan_member_controller.rb
jayroh: @jessieay @hartsick how should I break this up into 2 lines?

```ruby
class ExpensesStudentLoanMemberController < Medicaid::ManyMemberStepsController
```

I don't know if I've seen classes that have had to do that yet so I want to follow y'all's lead.

This ok?

```ruby
class ExpensesStudentLoanMemberController <
  Medicaid::ManyMemberStepsController
```

PR #412, opened 2017-10-31
Add shared example group for Medicaid multi-member controllers

jessieay:
* Extracts most common specs
* Assumes skip step if:
  * Single member household
  OR
  * multi-member with `anyone_*` attribute FALSE


jessieay:
cc @jayroh @hartsick for when we add more vanilla multi-member steps

PR #414, opened 2017-11-01
Add sign and submit step to Medicaid flow

hartsick:
screen shot 2017-10-31 at 6 08 46 pm
screen shot 2017-10-31 at 6 09 01 pm

[Finishes #152423776]

### DIFF
--- a/app/controllers/medicaid/sign_and_submit_controller.rb
+++ b/app/controllers/medicaid/sign_and_submit_controller.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Medicaid
+  class SignAndSubmitController < MedicaidStepsController
+    private
+
+    def step_params
+      super.merge(signed_at: Time.current)
+    end
+  end
+end
--- a/app/views/medicaid/_next_step.html.erb
+++ b/app/views/medicaid/_next_step.html.erb
@@ -1,5 +1,5 @@
 <footer class="form-card__button_right">
   <button class="button button--nav  button--cta button--full-width" type="submit">
-    Next
+    <%= local_assigns[:button_label].presence || "Next" %>
   </button>
 </footer>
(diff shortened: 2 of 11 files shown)


app/controllers/medicaid/sign_and_submit_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Medicaid
+  class SignAndSubmitController < MedicaidStepsController
+    private
+
+    def step_params
```
jessieay: I think we do not need to define this here if the attrs we are defining are just on the medicaid app itself because the default implementation in the parent controller looks at the step attrs to determine what to pull

app/views/medicaid/_next_step.html.erb
```diff
@@ -1,5 +1,5 @@
 <footer class="form-card__button_right">
   <button class="button button--nav  button--cta button--full-width" type="submit">
-    Next
+    <%= local_assigns[:button_label].presence || "Next" %>
```
jessieay: 👍

spec/controllers/medicaid/sign_and_submit_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+        "/steps/medicaid/confirmation",
+      )
+    end
+  end
+
+  describe "#update" do
+    it "updates the signature" do
+      medicaid_application = create(
+        :medicaid_application,
+        signature: "Hans Solo",
```
jessieay: maybe set `signed_at` to `nil` here for extra spec clarity?

app/controllers/medicaid/sign_and_submit_controller.rb
hartsick: I think this is required because we want to update it with the current time it hits the controller, not necessarily what's sent as a param in the request (this is also the pattern for SignAndSubmit)!

PR #417, opened 2017-11-01
Collect insurance type per client, per page

jayroh:
We want to loop through each user and ask them what their current insurance
plan is through. To do so we must break out of our current, codified,
patterns. We have to start with *A* member, and move to the next member
after submitting and saving, until we've reached the last user.

To complicate things, we have to be able to go *back* from the next
page/section and allow the client to re-enter adjust the insurance type if
they decide to go back.

To accomplish this we need to create the concept of a _current_member_ and
determine the next and previous members and the accompanying paths.

### DIFF
--- a/app/controllers/medicaid/insurance_current_type_controller.rb
+++ b/app/controllers/medicaid/insurance_current_type_controller.rb
@@ -1,8 +1,57 @@
 # frozen_string_literal: true
 
 module Medicaid
-  class InsuranceCurrentTypeController < Medicaid::ManyMemberStepsController
+  class InsuranceCurrentTypeController < MedicaidStepsController
+    helper_method :current_member
+
+    def update
+      @step = step_class.new(step_params)
+
+      if @step.valid?
+        current_member.update(step_params)
+        redirect_to(next_path)
+      else
+        render :edit
+      end
+    end
+
+    def current_member
+      @_current_member ||= begin
+                             member_from_form ||
+                               member_from_querystring ||
+                               first_insurance_holder
+                           end
+    end
+
     private
+
+    def next_path
+      next_member_path || super
+    end
+
+    def next_member_path
+      return if next_member.nil?
+
+      decoded_step_path(params: { member: next_member.id })
+    end
+
+    def first_insurance_holder
+      current_application.
+        members.
+        first_insurance_holder
+    end
+
+    def next_member
+      return if current_member.nil?
+
+      current_application.
+        members.
+        where(insured: true).
+        where("created_at > ?", current_member.created_at).
+        order(created_at: :asc).
+        limit(1).
+        first
+    end
 
     def skip?
       current_application.nobody_insured?
@@ -10,6 +59,23 @@ module Medicaid
 
     def member_attrs
       %i[insurance_type]
+    end
+
+    def existing_attributes
+      HashWithIndifferentAccess.new(current_member&.attributes)
+    end
+
+    def member_from_form
+      return if params.dig(:step, :member_id).blank?
+
+      member_id = params[:step].delete(:member_id)
+      current_application.members.find(member_id)
+    end
+
+    def member_from_querystring
+      return if params[:member].blank?
+
+      current_application.members.find(params[:member])
     end
   end
 end
--- a/app/controllers/medicaid/insurance_current_type_controller.rb
+++ b/app/controllers/medicaid/insurance_current_type_controller.rb
@@ -38,7 +38,9 @@ module Medicaid
     def first_insurance_holder
       current_application.
         members.
-        first_insurance_holder
+        insured.
+        limit(1).
+        first
     end
 
     def next_member
@@ -46,9 +48,8 @@ module Medicaid
 
       current_application.
         members.
-        where(insured: true).
-        where("created_at > ?", current_member.created_at).
-        order(created_at: :asc).
+        insured.
+        after(current_member).
         limit(1).
         first
     end
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -60,11 +60,14 @@ class Member < ApplicationRecord
     key: Rails.application.secrets.secret_key_for_ssn_encryption,
   )
 
-  def self.first_insurance_holder
+  def self.insured
     where(insured: true).
-      order(created_at: :asc).
-      limit(1).
-      first
+      where(requesting_health_insurance: true).
+      order(created_at: :asc)
+  end
+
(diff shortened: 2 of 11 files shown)


app/models/member.rb
```diff
@@ -58,6 +58,20 @@ class Member < ApplicationRecord
...
+      where(requesting_health_insurance: true).
+      order(created_at: :asc)
+  end
+
+  def self.first_member
+    limit(1).first
+  end
+
+  def self.after(member)
+    where("created_at > ?", member.created_at)
```
jessieay: thoughts on doing this based on `id` rather than `created_at`? IRL will probably be fine, but I imagine in specs this could sometimes lead to intermittent errors

app/steps/medicaid/insurance_current_type.rb
```diff
@@ -1,27 +1,33 @@
...
-      if member.requesting_health_insurance? && member.insured?
-        unless member.insurance_type.present?
-          member.errors.add(
-            :insurance_type,
-            "Please select a plan",
-          )
-        end
-      end
+    def member_has_insurance?
+      member.requesting_health_insurance? && member.insured?
```
jessieay: if this were not true for a member, would we ever get to this step for them anyway?

app/views/medicaid/insurance_current_type/edit.html.erb
```diff
@@ -4,25 +4,28 @@
   <header class="form-card__header">
     <div class="form-card__title">
       <%= t("medicaid.insurance_current_type.edit.title",
-            count: current_application.members.count) %>
+            name: current_member.full_name) %>
```
jessieay: I am going to move us to using `display_name` here soon (note to self, this is)

spec/features/medicaid_application_with_multiple_members_spec.rb
```diff
@@ -134,16 +134,9 @@
 
     on_page "Health Coverage Needs" do
       expect(page).to have_content(
-        "Tell us what insurance plan each person is enrolled in.",
+        "What type of insurance plan is Jessie Tester currently enrolled in",
```
jessieay: should we update this spec to include multiple `insured` members so that we can confirm the `next` logic is working as-expected?

spec/models/member_spec.rb
```diff
@@ -1,6 +1,39 @@
...
+        joel = create(:member)
+        jessie = create(:member)
+        christa = create(:member)
+
+        expect(Member.after(joel)).to eq [jessie, christa]
+        expect(Member.after(jessie)).to eq [christa]
+      end
+    end
+
+    describe ".first_member" do
```
jessieay: the way this is named, it feels like it is redundant of `primary_member` - what about naming in `next_member`?

app/controllers/medicaid/insurance_current_type_controller.rb
```diff
@@ -1,15 +1,80 @@
...
+
+      if @step.valid?
+        current_member.update(step_params)
+        redirect_to(next_path)
+      else
+        render :edit
+      end
+    end
+
+    def current_member
```
jessieay: feels like there's a lot happening in this method that should have specs -- maybe controller specs?

app/controllers/medicaid/insurance_current_type_controller.rb
```diff
@@ -1,15 +1,80 @@
...
+    end
+
+    def member_from_form
+      return if params.dig(:step, :member_id).blank?
+
+      member_id = params[:step].delete(:member_id)
+      current_application.members.find(member_id)
+    end
+
+    def member_from_querystring
```
jessieay: when would this be used?

app/models/member.rb
jayroh: I thought about ID's too and there's a reason why I did it this way.

Considering the fact we're holding onto personal information it is possible -- maybe likely? -- that the id columns will eventually change to be UUID's instead of integers. Now, that's entirely premature -- I know this. However - created_at and ID are analogous in terms of if/when they change. I can't remember the last time I saw a models' created_at timestamp get changed. If that remained the norm then id and created_at would serve in the same capacity.

app/controllers/medicaid/insurance_current_type_controller.rb
jayroh: When redirecting to the "next" member. see `next_path`

spec/models/member_spec.rb
jayroh: What would you suggest?

app/steps/medicaid/insurance_current_type.rb
jayroh: Don't know. I had just moved the logic from where it had been previously at line 17.

spec/models/member_spec.rb
jessieay: `next_member` would work for me!

app/controllers/medicaid/insurance_current_type_controller.rb
jayroh: There were controller specs: https://github.com/codeforamerica/michigan-benefits/pull/417/files#diff-01f7d54e987c44ab7c743f3b92c48b1bR4

app/controllers/medicaid/insurance_current_type_controller.rb
jessieay: gotcha! so we are using the querystring rather than nested routes. what is the thinking behind that?

jayroh: I don't think we could do nested routes considering we're automatically generating the routes via `Medicaid::StepNavigation`. We'd have to re-think the entire method to generate the routes if some of them turned from one controller and resource (the step) to two controllers: the parent question (in this case `type`) and the child (the member).

app/controllers/medicaid/insurance_current_type_controller.rb
```diff
@@ -1,15 +1,80 @@
 # frozen_string_literal: true
```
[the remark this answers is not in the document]
jessieay: agreed!

spec/models/member_spec.rb
jessieay: what do you think?

jayroh: I think I'm going to remove that scope entirely actually. The ROI on its existence isn't clear or beneficial enough to warrant it needing to be there.  `limit(1).first` is just fine in the context of the area of the code that uses it.

PR #418, opened 2017-11-01
Medicaid MM: Last 4 SSN and DOB for all household members

jessieay:
* As a multi-member Medicaid applicant, I want to be able in input
  Social Security and DOB for each household member.
* Finishes #152457770

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -48,6 +48,10 @@ class Member < ApplicationRecord
     inclusion: { in: SEXES },
     allow_nil: true
 
+  scope :requesting_health_insurance, -> do
+    where(requesting_health_insurance: true)
+  end
+
   attribute :ssn
   attr_encrypted(
     :ssn,
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -48,10 +48,6 @@ class Member < ApplicationRecord
     inclusion: { in: SEXES },
     allow_nil: true
 
-  scope :requesting_health_insurance, -> do
-    where(requesting_health_insurance: true)
-  end
-
   attribute :ssn
   attr_encrypted(
     :ssn,
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -1,4 +1,6 @@
 class Member < ApplicationRecord
+  include SocialSecurityNumber
+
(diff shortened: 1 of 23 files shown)


app/models/member.rb
```diff
@@ -1,4 +1,6 @@
 class Member < ApplicationRecord
+  include SocialSecurityNumber
```
jessieay: @hartsick including this module is what wait causing crazy failures before. The SocialSecurityNumber concern previously had some overrides to make activemodel act more like activerecord, so when we included it in an activerecord class it blew up.

Solution here was to create a `StepSocialSecurityNumber` concern that includes that stuff and then to have some duplicate logic in a regular `SocialSecurityNumber` concern

hartsick: Ahhh interesting!

app/views/medicaid/contact_social_security/edit.html.erb
```diff
@@ -7,14 +7,38 @@
...
+            type: :tel,
+            options: { maxlength: 4 },
+            notes: ["If you don’t know it you can skip this"],
+            append_html: "<p class='text--secure'><i class='illustration illustration--safety'></i>Social security numbers help ensure you receive the correct benefits. MDHHS maintains strict security guidelines to protect the identities of our residents.</p>" %>
 
-      <%= f.mb_date_select :birthday, "What is your birthday?", { options: { start_year: 1900, end_year: Time.now.year, default: Date.new(1990,1,15), order: [:month, :day, :year] } } %>
+          <%= ff.mb_date_select :birthday, "What is your birthday?", { options: { start_year: 1900, end_year: Time.now.year, default: Date.new(1990,1,15), order: [:month, :day, :year] } } %>
+        <% end %>
+      <% else %>
+        <% @step.members.select{|m| m.requesting_health_insurance? }.each do |member| %>
```
hartsick: small thing! we had talked about moving this filtering out of the view (maybe it belongs in the step or a shared step, since that's where it currently lives for providing info about your current health insurance)?

jessieay: good call! added a `requesting_health_insurance` scope on the member class at 98d764c0, which cleans this up a bit.

PR #422, opened 2017-11-02
Update Contributing with Tracker info

hartsick:
[Finishes #152517617]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,16 +39,14 @@ Ask the team for access to the following services to begin contributing:
 
 * [GitHub](https://github.com/codeforamerica/michigan-benefits)
 * Slack
+* [Pivotal Tracker](https://www.pivotaltracker.com/n/projects/2123705)
 * Twilio
 * Mailgun
-* [Trello - Michigan Prototype](https://trello.com/b/aN8DB2pN/medicaid-application-arrives-via-online-search) 
-* [Trello - Design Backlog](https://trello.com/b/aBqTrqaJ/the-digital-assister-michiganbenefitsorg)
-* [Pivotal Tracker - Engineering Backlog](https://www.pivotaltracker.com/n/projects/2123705)
+* Trello - Benefits Partnership org
 * Heroku
 * Password Manager
 * [multi-benefits-product Google Calendar](codeforamerica.org_jkbrndbh2ubvh6frucq6dc2chk@group.calendar.google.com)
 * Get calendar invites to: daily standup, weekly planning, weekly retro
-
 
 ## Day-to-day Development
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -126,10 +129,9 @@ resources and workflow.
 ### Planning
 
 * A change is initiated and discussed via a
-  [Trello](https://trello.com/b/aBqTrqaJ/michiganbridgesorg) card.
-* All new Trello cards are added to the "Inbox" Trello list.
-* During a weekly planning meeting, prioritized Trello cards are added to the
-  "This week" Trello list. Cards in this list should have enough detail to be
+  [Pivotal Tracker](https://www.pivotaltracker.com/n/projects/2123705) story.
+* All new Tracker stories are added to the top of the Icebox
+* During a weekly planning meeting, prioritized Tracker stories are moved over to the Backlog. Cards in this list should have enough detail to be
```
jessieay: maybe add a note about doing estimation during that mtg as well?

CONTRIBUTING.md
```diff
@@ -143,18 +145,18 @@ Person who noticed the problem, writes a ticket with a "Bug Triage" label and pu
...
+#### Apply the correct Tracker label
 
 | Label      | Severity | Action |
 | ------------- | -----------| ----- |
-| P1 | Sever bug. (e.g. the site is down, apps are at risk of not being delivered) | Should be fixed immediately |
-| P2 | Some negative impact on users | Should be placed on top of Ready column (PM may increase priority) |
-| P3 | No big impact, low priority | Place in backlog |
+| P1 | Severe bug. (e.g. the site is down, apps are at risk of not being delivered) | Should be fixed immediately |
+| P2 | Some negative impact on users | Should be placed on top of Backlog (PM may increase priority) |
+| P3 | No big impact, low priority | Place in Icebox for prioritization |
```
jessieay: any other labels in tracker that we should be using?

CONTRIBUTING.md
```diff
@@ -143,18 +145,18 @@ Person who noticed the problem, writes a ticket with a "Bug Triage" label and pu
...
 
-### Work In Process
+### Work In Progress
 
-* When someone starts working on a task, they move the Trello card into "In
-  Progress" and assign themselves so that others know they are working on it.
+* When someone starts working on a task, they click start on the Tracker story. This will assign the
```
jessieay: `click "Start"` might be a tad clearer 🍇

CONTRIBUTING.md
```diff
@@ -169,9 +171,9 @@ development environment visit your local [mailcatcher server](http://localhost:1
 ### Code Review
 
 * A Pull Request (PR) is created on GitHub. All PRs should have [a good commit
-  message](https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message).
-[The PR is attached to the related Trello
-card](https://blog.trello.com/github-and-trello-integrate-your-commits).
+  message](https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message). Optionally, include a 
+  link to the Tracker story in the PR summary. When a PR completes a Tracker story, mark that story as 'Finished'
```
jessieay: should we mention the magical [Fixes #1234] thing here or nah?

jessieay: also, I wonder if we should make this a requirement rather than optionally. Much easier to code review when you can see the source ticket!

CONTRIBUTING.md
```diff
@@ -190,6 +192,8 @@ feature. In general, the team prioritizes code reviews over other work.
 * Merged code is automatically deployed to the [staging
   environment](https://michigan-benefits-staging.herokuapp.com/) after a green
 build on Circle CI.
+* Once code completing a Tracker story is deployed to staging, mark the story as 'Delivered'. This
```
jessieay: this is true for features, but not for chores. Not sure about bugs. Should we explain the difference?

CONTRIBUTING.md
```diff
@@ -213,6 +218,34 @@ and will not work on this URL:
 
 [https://michigan-benefits-staging.herokuapp.com](https://michigan-benefits-staging.herokuapp.com)
 
+### Tracker flow and integrations
```
jessieay: this section may be more useful above, since I didn't read it until afterr I'd read the stuff about accept/reject

PR #423, opened 2017-11-02
Add Devise auth for admin dashboard

jessieay:
[Finishes #152367741]

**NOTES**:
- chose Devise bc it will be easy to add 2FA later
- did not include "registerable" functionality so all new accounts will
need to be created via Rails console
- attempted to throw a "sign out" link in there but that was proving to
be difficult CSS-wise and was not a requirement in the ticket so I left
it out.
- confirmed that reset password email does work
- we still need the ADMIN_PASSWORD and ADMIN_USER env vars for delayed
job.

screen shot 2017-11-02 at 11 58 30 am

### DIFF
--- a/app/models/admin_user.rb
+++ b/app/models/admin_user.rb
@@ -1,0 +1,9 @@
+class AdminUser < ApplicationRecord
+  devise(
+    :database_authenticatable,
+    :recoverable,
+    :rememberable,
+    :trackable,
+    :validatable,
+  )
+end
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ gem "bourbon", "~> 4.2.0" # to keep in sync with getcalfresh
 gem "capybara"
 gem "delayed_job_active_record"
 gem "delayed_job_web"
+gem "devise"
 gem "faraday"
 gem "good_migrations"
 gem "high_voltage"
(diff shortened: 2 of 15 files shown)


app/models/admin_user.rb
```diff
@@ -0,0 +1,9 @@
+class AdminUser < ApplicationRecord
+  devise(
+    :database_authenticatable,
+    :recoverable,
+    :rememberable,
+    :trackable,
+    :validatable,
```
hartsick: Probably want to add 'timeoutable' and possibly 'lockable' in future? Maybe we can put in same story as logout functionality

jessieay:
@hartsick we use `delayed_job_web` to view DJ jobs : https://github.com/codeforamerica/michigan-benefits/blob/6faba438eecd7dfb5179aad479895b029fd83ab6/config/routes.rb#L4

and here we protext the view via basic auth: https://github.com/codeforamerica/michigan-benefits/blob/2237791245581679bf39092d8e716deafe87e179/config.ru#L7

but it looks like changing to devise would be easy: https://github.com/ejschmitt/delayed_job_web#authenticating-with-devise-and-warden

making a ticket for icebox now!

jessieay:
@hartsick added ticket at https://www.pivotaltracker.com/story/show/152528278

PR #425, opened 2017-11-03
Select other income types per-member for Medicaid

jessieay:
**WHY**:
- before, this was saved on the medicaid app as a series of booleans
- now, we are saving each type as as string in an array field called
"other_income_types" that is on the member.
- extracted some common controller code into an inheritance structure
- part III of https://trello.com/c/tHLD1PgG/441-medicaid-mm-flow-current-income-does-anyone-in-the-household-get-income-thats-not-from-a-job-nested-2

screen shot 2017-11-03 at 9 46 06 am

TODO:
- remove / migrate the old boolean fields on medicaid
- refine the "back" logic. Right now, going back from any
member page goes to '/medicaid/income-other-income-member' and going
back from "/medicaid/expenses-alimony" goes to
"/medicaid/income-other-income-type" for the primary member (rather than
going back through the pages in the member order)

### DIFF
--- a/app/controllers/medicaid/member_steps_controller.rb
+++ b/app/controllers/medicaid/member_steps_controller.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module Medicaid
+  class MemberStepsController < MedicaidStepsController
+    helper_method :current_member
+
+    def current_member
+      @_current_member ||= member_from_form || member_from_querystring
+    end
+
+    private
+
+    def next_path
+      next_member_path || super
+    end
+
+    def next_member_path
+      return if next_member.nil?
+
+      decoded_step_path(params: { member: next_member.id })
+    end
+
+    def existing_attributes
+      HashWithIndifferentAccess.new(current_member&.attributes)
+    end
+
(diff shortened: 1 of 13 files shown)


jessieay:
[[Medicaid MM Flow] Current Income: Does anyone in the household get income that's not from a job? + Nested (2)](https://trello.com/c/tHLD1PgG/441-medicaid-mm-flow-current-income-does-anyone-in-the-household-get-income-thats-not-from-a-job-nested-2)

app/controllers/medicaid/member_steps_controller.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Medicaid
+  class MemberStepsController < MedicaidStepsController
```
jayroh: Note to us/team: if we're going to stick with the convention of namespacing the classes we inherit from, we should also adjust and move `medicaid_member_steps_controller.rb` and `medicaid_steps_controller.rb` into the medicaid controller namespace too.  Right now they're on the same top level as the SNAP controllers.

PR #428, opened 2017-11-03
Remove all frozen string magic comments

jayroh:
And add frozen string literal ENV variable


jessieay:
see https://freelancing-gods.com/2017/07/27/friendly-frozen-string-literals.html for background

PR #431, opened 2017-11-03
Add docs for admin dashboard / admin accounts

jessieay:
**WHY**:
So people know how to create admin accounts.

Finishes #152446453


jessieay:
@hartsick the ticket for this included

> How and when to delete a user account

but I feel like maybe that should be a part of offboarding docs instead?

PR #432, opened 2017-11-03
Use `display_name` instead of `full_name` to show member names

jessieay:
**WHY**: in the future, we may want to change how we do this - maybe we
show first name and DOB, for example, or first name and last initial.
This way of naming the method makes it more flexible.

I also took this opp to move to use of `display_name` in places where we
were just concatenating first name and last name.

I also also took this opp to titleize first name and last name. :)

Finishes https://www.pivotaltracker.com/story/show/152565843

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -68,8 +68,8 @@ class Member < ApplicationRecord
     where("created_at > ?", member.created_at)
   end
 
-  def full_name
-    "#{first_name} #{last_name}"
+  def display_name
+    "#{first_name} #{last_name}".titleize
   end
 
   def primary_member?
--- a/app/models/snap_application_member_attributes.rb
+++ b/app/models/snap_application_member_attributes.rb
@@ -6,7 +6,7 @@ class SnapApplicationMemberAttributes
 
   def to_h
     {
-      "#{position}_member_full_name" => member.full_name,
+      "#{position}_member_full_name" => member.display_name,
       "#{position}_member_sex_male" => bool_to_checkbox(member.sex == "male"),
       "#{position}_member_sex_female" =>
         bool_to_checkbox(member.sex == "female"),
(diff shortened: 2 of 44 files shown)


app/models/snap_application_member_attributes.rb
```diff
@@ -6,7 +6,7 @@ def initialize(member:, position:)
 
   def to_h
     {
-      "#{position}_member_full_name" => member.full_name,
+      "#{position}_member_full_name" => member.display_name,
```
jessieay: note that I kept these hash keys as `full_name` because that is how the fields are named in the PDF and that felt like an accurate name for the info the PDF is looking for. This does bring up the question of whether we should still use `full_name` for the PDF data (the key, in this case) to protect us in the future if we change `display name` to not be the full name

app/models/member.rb
```diff
@@ -68,8 +68,8 @@ def self.after(member)
     where("created_at > ?", member.created_at)
   end
 
-  def full_name
-    "#{first_name} #{last_name}"
+  def display_name
+    "#{first_name} #{last_name}".titleize
```
jessieay: this is where the magic is happening

jayroh: If the member is the _primary_ member should we use "You" instead of their name? I've seen in a story or two where it's been split in that way - "You" versus "Jane/John Doe".

jessieay: good idea / point!

jessieay: looking through the diff, changing `Joe Shmoe` to `you` would end up creating a product change rather than be a refactor of how we are displaying names, so for now I think it probably makes sense to keep as-is but note that we may change this in the future (and when that happens, having this method will hopfefully make it an easy change!)

jessieay:
@hartsick created a trello card on the research board for this https://trello.com/c/aRipYPYQ

PR #433, opened 2017-11-06
Add sections headers to `/medicaid/amounts-income`

jessieay:
https://www.pivotaltracker.com/story/show/152450581
[Finishes #152450581]

screen shot 2017-11-06 at 12 10 22 pm

### DIFF
--- a/spec/features/medicaid_application_with_maximum_info_spec.rb
+++ b/spec/features/medicaid_application_with_maximum_info_spec.rb
@@ -104,11 +104,11 @@ RSpec.feature "Medicaid app" do
     on_pages "Income & Expense Amounts" do
       click_on "Next"
 
-      fill_in "Your Current Job (#1)", with: 100
-      fill_in "Your Current Job (#2)", with: 50
-      fill_in "Your Current Job (#3)", with: 25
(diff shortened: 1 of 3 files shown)


spec/features/medicaid_application_with_maximum_info_spec.rb
```diff
@@ -104,11 +104,11 @@
     on_pages "Income & Expense Amounts" do
       click_on "Next"
 
-      fill_in "Your Current Job (#1)", with: 100
-      fill_in "Your Current Job (#2)", with: 50
-      fill_in "Your Current Job (#3)", with: 25
-      fill_in "Your Self-Employment", with: 100
-      fill_in "Your Unemployment", with: 100
+      fill_in "step_employed_monthly_income_0", with: 100
```
hartsick: Are we keying off of ID to reduce reliance of our tests on copy? Or is there another reason?

jessieay: the reason is that the input labels are now non-unique so the input ids were better for targeting specific inputs. I liked the readability of what we had before but this does have the added benefit of being *not* tied to copy!

PR #434, opened 2017-11-06
Extract step attrib reconciliation to own object

jayroh:
... and make the form builder behave accordingly when an attribute is of
type "array".

The majority of the attributes on our steps are mapped plain vanilla
literals - strings, integers, bools, etc. On the VERY rare occasion there
are fields that map to arrays of those literals. For example - documents
and paperwork.

Documents and paperwork share one partial that has explicitly overwritten
the value of the "name" option in the form input field.

For example. A normal form field comes out like so:

`<input type="text" name="step[first_name]">`

But when we would like N number of values for a field that maps to a
*array* column in the database, we'd like the html instead to be as
follows:

`<input type="text" name="step[monthly_income][]">`

Before this commit, the only way to do that would be to override the name
option with something akin to:

```
<%= f.mb_money_field :monthly_income,
 "Job N",
 name: "step[monthly_income][]" %>
```

After this commit, the form builder will be smart enough to know that that
attribute is hash key for an array/collection.

***

Also!

In addition to the above, the work that's been done inside the base step
class to munge/extract information about the attribute definitions that are
provided by the `step_attributes` class method has been extracted to its
own class to do the work of extracting the string values that are passed to
`attr_accessor`.

Lastly - a method to ask "is this attribute a symbol or hash key" is in
this attributes class to allow the form builder to decide whether the form
field needs to allow for multiple values (for arrays) or not (for object
literals)

***

Also: Only ask abt `hash_key_attribute?` on Step objects

In some cases the form builder is being used with a straight model object,
therefore the `hash_key_attribute?` method will most certainly not work.

### DIFF
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -18,9 +18,10 @@ class StepsController < ApplicationController
   end
 
   def edit
-    @step = step_class.new(existing_attributes.slice(*step_attrs))
+    attribute_keys = Step::Attributes.new(step_attrs).to_sym
+    @step = step_class.new(existing_attributes.slice(*attribute_keys))
 
-    if step_class.attribute_names&.include?(:members)
+    if attribute_keys.include?(:members)
       @step.members = current_application&.members
     end
 
--- a/app/helpers/mb_form_builder.rb
+++ b/app/helpers/mb_form_builder.rb
@@ -12,6 +12,7 @@ class MbFormBuilder < ActionView::Helpers::FormBuilder
     optional: false
   )
     classes = classes.append(%w[text-input])
+
     text_field_options = {
       autofocus: autofocus,
       type: type,
@@ -21,11 +22,25 @@ class MbFormBuilder < ActionView::Helpers::FormBuilder
       autocapitalize: "off",
       spellcheck: "false",
     }.merge(options)
+
+    if object.hash_key_attribute?(method)
+      text_field_options[:name] = "#{object_name}[#{method}][]"
+    end
+
     text_field_html = text_field(method, text_field_options)
+    label_and_field_html = label_and_field(
+      method,
+      label_text,
+      text_field_html,
+      notes: notes,
+      prefix: prefix,
+      optional: optional,
+      options: options,
+    )
 
     <<-HTML.html_safe
       <fieldset class="form-group#{error_state(object, method)}">
-        #{label_and_field(method, label_text, text_field_html, notes: notes, prefix: prefix, optional: optional, options: options)}
+        #{label_and_field_html}
         #{errors_for(object, method)}
         #{append_html}
       </fieldset>
--- a/app/helpers/mb_form_builder.rb
+++ b/app/helpers/mb_form_builder.rb
@@ -23,30 +23,51 @@ class MbFormBuilder < ActionView::Helpers::FormBuilder
       spellcheck: "false",
     }.merge(options)
 
-    if object.respond_to?(:hash_key_attribute?) &&
-        object.hash_key_attribute?(method)
+    field_values(text_field_options, method).map.with_index do |value, i|
+      if array_of_values?(method)
+        text_field_options[:name] = "#{object_name}[#{method}][]"
+        text_field_options[:id] = "#{object_name}_#{method}_#{i}"
+        options[:id] = "#{object_name}_#{method}_#{i}"
+      end
 
-      text_field_options[:name] = "#{object_name}[#{method}][]"
-    end
+      text_field_options[:value] = value
 
-    text_field_html = text_field(method, text_field_options)
-    label_and_field_html = label_and_field(
-      method,
-      label_text,
-      text_field_html,
-      notes: notes,
-      prefix: prefix,
-      optional: optional,
-      options: options,
-    )
+      text_field_html = text_field(method, text_field_options)
 
-    <<-HTML.html_safe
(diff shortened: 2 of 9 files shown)


app/helpers/mb_form_builder.rb
```diff
@@ -21,11 +22,27 @@ def mb_input_field(
       autocapitalize: "off",
       spellcheck: "false",
     }.merge(options)
+
+    if object.respond_to?(:hash_key_attribute?) &&
```
hartsick: Ben and I found this line particularly confusing! The tests help provide some clarity, but we still had a problem parsing and fully understanding what is going on. Can you think of any way of making the intentions more obvious?

spec/steps/step/attributes_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe Step do
+  describe "#hash_key?" do
+    it "returns true if the attribute is a hash key" do
+      names = [
+        { foo: [] },
```
hartsick: These tests are great! I don't think I understand why we would expect a Hash in one case and a Symbol in another. Could you help explain that? And is there a way we could make those expectations or reasoning clearer in the code?

app/helpers/mb_form_builder.rb
```diff
@@ -23,30 +23,51 @@ def mb_input_field(
...
+  def field_values(options, method)
+    current_value = object.send(method)
+    return Array.wrap(current_value) if current_value.present?
+
+    if options[:count] && current_value.blank?
+      Array.new(options[:count], "")
+    else
+      [""]
+    end
+  end
```
jayroh: The corresponding test for this gives a little more context, but the gist is -

* If there's a value, return it wrapped in an array
* If there's no value, and we are told _how many_ elements we have by default (like, how many jobs does a person have?), we should create an array of empty values that can be filled in
* If there's no value, and no count provided - create an array of one empty string

app/helpers/mb_form_builder.rb
```diff
@@ -23,30 +23,51 @@ def mb_input_field(
...
     }.merge(options)
 
-    if object.respond_to?(:hash_key_attribute?) &&
-        object.hash_key_attribute?(method)
+    field_values(text_field_options, method).map.with_index do |value, i|
+      if array_of_values?(method)
+        text_field_options[:name] = "#{object_name}[#{method}][]"
+        text_field_options[:id] = "#{object_name}_#{method}_#{i}"
+        options[:id] = "#{object_name}_#{method}_#{i}"
+      end
```
jayroh: if there is an array of values, the input name, id, and label id, should all be changed so that the html is correct.

eg:

```
name = "step[income][]"
id = step_income_1
label for = step_income_1
```

app/controllers/steps_controller.rb
```diff
@@ -18,9 +18,10 @@ def current_application
   end
 
   def edit
-    @step = step_class.new(existing_attributes.slice(*step_attrs))
+    attribute_keys = Step::Attributes.new(step_attrs).to_sym
+    @step = step_class.new(existing_attributes.slice(*attribute_keys))
```
jayroh: extract just the keys (as symbols) for the attributes from the attribute names

app/helpers/mb_form_builder.rb
```diff
@@ -23,30 +23,51 @@ def mb_input_field(
       spellcheck: "false",
     }.merge(options)
 
-    if object.respond_to?(:hash_key_attribute?) &&
-        object.hash_key_attribute?(method)
+    field_values(text_field_options, method).map.with_index do |value, i|
```
jayroh: loop through all our field values (1...N) times and build the appropriate HTML

spec/steps/step_spec.rb
```diff
@@ -0,0 +1,31 @@
+require "rails_helper"
+
+RSpec.describe Step, type: :model do
+  describe ".step_attributes" do
+    it "creates setters and getters" do
+      class TestStep < Step
```
jessieay: 😍

app/helpers/mb_form_builder.rb
jessieay: oooh I think Christa and I dealt with a similar issue where we had to set name and id explicitly for a field and it was totally bananas.

see `app/views/medicaid/amounts_income/edit.html.erb`

spec/helpers/mb_form_builder_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "rails_helper"
+
+RSpec.describe MbFormBuilder do
```
jessieay: feels really really nice that we have unit tests for this now -- should ping our friends at GCF @bensheldon

app/helpers/mb_form_builder.rb
jayroh: @jessieay I'm 99% sure that I would consider _*that*_ a bug. To work around the issue by setting attributes to get the expected behavior correct? That's definitely a bug.

jayroh:
Jessie a lot of this, I would consider, is bug _like_  ... by that I mean that expected behavior is not what happens. It doesn't act how we expect it to. Most of this is around the edge case where we have columns using arrays - to get around it is a troublesome series of hoops to jump thought.

spec/helpers/mb_form_builder_spec.rb
bensheldon: That is lovely 🌷

PR #435, opened 2017-11-07
Indicate if mailing address same as residential

disambiguator:
[Finishes #152531799]

screen shot 2017-11-06 at 3 55 29 pm

Signed-off-by: Paras Sanghavi <paras@codeforamerica.org>


jessieay:
We did it! merge away

PR #436, opened 2017-11-07
Add and use SSN rather than last four SSN

hartsick:
Keeps last four SSN around in case we want to use again in near future. Prevents accessing this field on model to prevent confusion.

[Finishes #152614357]
https://www.pivotaltracker.com/story/show/152614357

Notes:
* We made a column on medicaid_applications for ssn (to just revert the conversion to last_four_ssn), but think that medicaid applications are actually referring to members' SSN. This means we could remove SSN column (and potentially last_four_ssn) from the medicaid_applications table. Because we have time pressure on this ticket, we're not including in here.


db/migrate/20171107000347_add_ssn_to_member_and_medicaid_application.rb
```diff
@@ -0,0 +1,12 @@
+class AddSsnToMemberAndMedicaidApplication < ActiveRecord::Migration[5.1]
+  def up
+    add_column :members, :encrypted_ssn, :string
+    add_column :members, :encrypted_ssn_iv, :string
+    add_column :medicaid_applications, :encrypted_ssn, :string
```
jessieay: I don't think we need to add this at this point -- we are only saving SSN on `members` for medicaid apps.

if my memory is correct, we were keeping `last_four_ssn` around on medicaid apps because we'd collected that data on the medicaid app directly for the single member flow but never got around to transferring that data onto the primary member for legacy medicaid_applications.

db/seeds.rb
```diff
@@ -58,8 +58,8 @@
 minimal_primary.update!(
   marital_status: "Never married",
   sex: "male",
-  encrypted_last_four_ssn: "aohUJzeKx7bLqVQ1SfI4tl+OeIg=\n",
-  encrypted_last_four_ssn_iv: "Lyh6r4q4SVQLKyDr\n",
+  encrypted_ssn: "UdGpZ8L9KTaTjsjOtT9F3He/AtUaxIz2EQ==\n",
```
jessieay: good catch getting the seeds this time!

db/migrate/20171107000347_add_ssn_to_member_and_medicaid_application.rb
hartsick: Ahhh that was it! Thanks! Just added migration to remove.

PR #439, opened 2017-11-07
Small copy update

rachelcope:
* https://trello.com/c/nL9ANXqf/119-employment-as-a-client-i-need-to-know-that-temporary-jobs-count-as-employment
* add "temporary" to employment description

image


hartsick:
@rachelcope We just re-ran the Circle build and it cleaned up the tests!

PR #441, opened 2017-11-08
In Medicaid flow, ask if anyone in household is married.

hartsick:
Specifying which members of the family are married and to whom for multi-member household come in the next step.

We've kept married status independent from SNAP's marital status, because they are used and defined differently across the two applications.

[Finishes #152548726]
https://www.pivotaltracker.com/story/show/152548726

screen shot 2017-11-07 at 5 26 03 pm
screen shot 2017-11-07 at 5 26 46 pm

### DIFF
--- a/app/controllers/medicaid/intro_marital_status_controller.rb
+++ b/app/controllers/medicaid/intro_marital_status_controller.rb
@@ -1,0 +1,15 @@
+module Medicaid
+  class IntroMaritalStatusController < MedicaidStepsController
+    private
+
+    def after_successful_update_hook
+      if single_member_household?
+        current_application.primary_member.update!(member_attrs)
+      end
+    end
+
+    def member_attrs
+      { married: step_params[:anyone_married] }
+    end
+  end
+end
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -30,6 +30,14 @@ class Member < ApplicationRecord
     "Other",
   ].freeze
 
+  MARITAL_STATUSES = [
+    "Married",
+    "Never married",
+    "Divorced",
+    "Widowed",
+    "Separated",
+  ].freeze
+
   belongs_to :benefit_application, polymorphic: true
 
   validates :employed_pay_interval,
--- a/app/views/medicaid/intro_household/edit.html.erb
+++ b/app/views/medicaid/intro_household/edit.html.erb
@@ -40,8 +40,10 @@
   </div>
 
   <footer class="form-card__button_right">
-    <a href="/steps/medicaid/intro-college" class="button button--nav  button--cta button--full-width">
-      Next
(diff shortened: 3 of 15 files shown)


app/controllers/medicaid/intro_marital_status_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Medicaid
+  class IntroMaritalStatusController < MedicaidStepsController
+    private
+
+    def after_successful_update_hook
+      if single_member_household?
```
jessieay: ❤️  😻

app/models/member.rb
```diff
@@ -30,6 +30,14 @@ class Member < ApplicationRecord
     "Other",
   ].freeze
 
+  MARITAL_STATUSES = [
```
jessieay: 💪

app/models/member.rb
```diff
@@ -30,6 +30,14 @@ class Member < ApplicationRecord
...
+    "Married",
+    "Never married",
+    "Divorced",
+    "Widowed",
+    "Separated",
```
jessieay: maybe alpha order these ?

app/views/medicaid/intro_household/edit.html.erb
```diff
@@ -40,8 +40,10 @@
   </div>
 
   <footer class="form-card__button_right">
-    <a href="/steps/medicaid/intro-college" class="button button--nav  button--cta button--full-width">
-      Next
-    </a>
+    <%= link_to(
```
jessieay: hah nice

app/models/member.rb
jessieay: should we validate that marital statuses are included in this list at the model level? I know we are doing so at the step level but might be good to have it here too so we have "normalized data" -- your call!

hartsick: yeah, I think that's a great idea! we decided not to for now because 1) we ended up not using marital statuses in our feature, and 2) it would likely introduce some validation problems with existing data when we deployed it. thought we could add validations for inclusion in a follow-up!

app/views/medicaid/intro_household/edit.html.erb
hartsick: this was the ultimate culprit >.<

db/migrate/20171108010437_add_married_to_member.rb
```diff
@@ -0,0 +1,10 @@
+class AddMarriedToMember < ActiveRecord::Migration[5.1]
+  def up
+    add_column :members, :married, :boolean
```
jessieay: random thought I had (in my sleep!) about this: why not use the `marital_status` column that we already have on `members` and just updated to `Married` ?

PR #442, opened 2017-11-08
Collect income amounts per member on series of pages

jayroh:
If a medicaid application household has anyone who is receiving some sort of income, we ask WHO and HOW MUCH each of those members is making. In addition - if there are multiple jobs for a member, we ask how much they make per each job.

Also:

* Update `Member#employed` to true if the job number is positive when that job number changes
* Create a `MemberStep` class that can be subclassed. This allows for an inherited `member` method/object for the step.


config/locales/en.yml
```diff
@@ -81,5 +81,14 @@ en:
...
       edit:
         title:
-          one: Provide your Social Security Number and Date of Birth if you’re ready
-          other: Provide the Social Security Numbers and Dates of Birth for your household members if you're ready
+          one:
+            Provide your Social Security Number and Date of Birth if
+            you’re ready
+          other:
+            Provide the Social Security Numbers and Dates of Birth for your
+            household members if you're ready
```
jessieay: curly quote on this `you're` while we're here?

spec/controllers/medicaid/amounts_income_controller_spec.rb
```diff
@@ -2,13 +2,13 @@
 
 RSpec.describe Medicaid::AmountsIncomeController do
   describe "#edit" do
-    context "client is has no income" do
```
jessieay: 😆

spec/controllers/medicaid/amounts_income_controller_spec.rb
```diff
@@ -18,14 +18,33 @@
       end
     end
 
-    context "client has income" do
+    context "current member does not have income" do
```
jessieay: should this be wrapped in a `single member household` context?

app/views/medicaid/amounts_income/edit.html.erb
```diff
@@ -1,8 +1,12 @@
-<% content_for :header_title, "Income & Expense Amounts" %>
+<% content_for :header_title, "Income &amp; Expense Amounts".html_safe %>
```
jessieay: I think that this file will have merge conflicts with `master` because we recently updated it to have a slightly different style, which affected the input field names

spec/features/medicaid_application_with_maximum_info_spec.rb
```diff
@@ -104,11 +104,11 @@
     on_pages "Income & Expense Amounts" do
       click_on "Next"
 
-      fill_in "Your Current Job (#1)", with: 100
-      fill_in "Your Current Job (#2)", with: 50
-      fill_in "Your Current Job (#3)", with: 25
-      fill_in "Your Self-Employment", with: 100
-      fill_in "Your Unemployment", with: 100
+      fill_in "Job 1", with: 100
```
jessieay: these will change with the new page styles that are on `master` (we are using `id`s now since all income inputs technically have the same label and are only differentiated by a header element)

app/steps/medicaid/amounts_income.rb
```diff
@@ -1,19 +1,49 @@
...
+        false
+      else
+        true
+      end
+    end
+
+    private
+
+    def jobs_and_income_dont_match?
+      employed_number_of_jobs != employed_monthly_income&.count
```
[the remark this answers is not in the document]
jayroh: I think it's a toss-up. I never remember either. It's kind of a gut feel. This seems like we're asking for a count of the number of salaries. 

:man_shrugging:

PR #445, opened 2017-11-09
Select spouses for all married people

jessieay:
* [Finishes #152550364]
GIVEN I am multi-member Medicaid applicant
WHEN I select which members are currently married
THEN I should go to a NEW page (for each member), where I can select which person they are married to

* Only show members of household who have marked that they were married on previous step.

NOTE: the following was included as part of the feature request:
* If someone has indicated they are married to someone in the same
household, we should assume that person is also married to the original
person and not ask them.

If it is indicated that Person 1 is married to Person 2, I am marking
Person 1 as the spouse of Person 2 as well. *BUT* I am not skipping over
Person 2. The reason for this is that if we go back and update the
spouse of Person 1 to Person 3, we would need to clear out the spouse of
Person 2 in order to get everyone's spouse recorded properly. It felt
more straightforward to just show each person with info pre-populated
for spouse than to do this complicated clearing stuff. But, we totally
can do that if we need to.

screen shot 2017-11-08 at 3 42 47 pm

screen shot 2017-11-08 at 3 42 52 pm

screen shot 2017-11-08 at 3 43 00 pm

screen shot 2017-11-08 at 3 47 00 pm

### DIFF
--- a/app/controllers/medicaid/intro_marital_status_indicate_spouse_controller.rb
+++ b/app/controllers/medicaid/intro_marital_status_indicate_spouse_controller.rb
@@ -3,18 +3,6 @@ module Medicaid
     Medicaid::MemberStepsController
     helper_method :spouse_options
 
-    def update
-      @step = step_class.new(step_params)
-
-      if @step.valid?
-        current_member.update(step_params)
-        update_spouse_of_other_member_if_present
-        redirect_to(next_path)
-      else
-        render :edit
-      end
-    end
-
     def current_member
       @_current_member ||= super || first_married_member
     end
@@ -26,6 +14,15 @@ module Medicaid
     end
 
     private
+
+    def after_successful_update_hook
+      current_member.update(step_params)
+      update_spouse_of_other_member_if_present
+    end
+
+    def application_params
+      {}
+    end
 
     def skip?
       single_member_household? ||
--- a/app/controllers/medicaid/intro_marital_status_indicate_spouse_controller.rb
+++ b/app/controllers/medicaid/intro_marital_status_indicate_spouse_controller.rb
@@ -1,0 +1,61 @@
+module Medicaid
+  class IntroMaritalStatusIndicateSpouseController <
+    Medicaid::MemberStepsController
+    helper_method :spouse_options
+
+    def update
+      @step = step_class.new(step_params)
+
+      if @step.valid?
+        current_member.update(step_params)
+        update_spouse_of_other_member_if_present
+        redirect_to(next_path)
+      else
+        render :edit
+      end
+    end
+
+    def current_member
+      @_current_member ||= super || first_married_member
+    end
+
+    def spouse_options
+      current_application.members -
+        [current_member] +
+        [OtherSpouse.new]
+    end
+
+    private
+
+    def skip?
+      single_member_household? ||
+        current_application.nobody_married?
+    end
+
+    def first_married_member
+      current_application.
+        members.
+        married.
+        limit(1).
+        first
+    end
+
+    def next_member
+      return if current_member.nil?
+
+      current_application.
+        members.
+        married.
+        after(current_member).
+        limit(1).
+        first
+    end
+
+    def update_spouse_of_other_member_if_present
+      spouse = Member.where(id: step_params[:spouse_id]).first
+      if spouse.present?
+        spouse.update!(spouse_id: current_member.id)
+      end
+    end
+  end
+end
--- a/app/controllers/medicaid/member_steps_controller.rb
+++ b/app/controllers/medicaid/member_steps_controller.rb
@@ -8,11 +8,11 @@ module Medicaid
       @_current_member ||= member_from_form || member_from_querystring
     end
 
-    private
-
     def next_path
       next_member_path || super
     end
+
+    private
 
     def next_member_path
       return if next_member.nil?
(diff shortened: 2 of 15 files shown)


app/controllers/medicaid/intro_marital_status_indicate_spouse_controller.rb
```diff
@@ -0,0 +1,61 @@
+module Medicaid
+  class IntroMaritalStatusIndicateSpouseController <
+    Medicaid::MemberStepsController
+    helper_method :spouse_options
+
+    def update
+      @step = step_class.new(step_params)
+
+      if @step.valid?
```
disambiguator: I would replace:
```
def update
 ...
end
```

with:

```
    def after_successful_update_hook
      update_spouse_of_other_member_if_present
    end
```

especially since the controller_spec is not testing the non-valid code path here

app/controllers/medicaid/member_steps_controller.rb
```diff
@@ -8,12 +8,12 @@ def current_member
       @_current_member ||= member_from_form || member_from_querystring
     end
 
-    private
-
     def next_path
       next_member_path || super
     end
 
+    private
```
disambiguator: next_path could probably stay private if we use the inheritance structure in IntroMaritalStatusIndicateSpouseController

app/models/medicaid_application.rb
```diff
@@ -28,6 +28,10 @@ def anyone_with_income?
       anyone_other_income?
   end
 
+  def nobody_married?
```
disambiguator: Does this need a test? Genuine question - idk testing philosophy for units like this.

app/steps/medicaid/intro_marital_status_indicate_spouse.rb
```diff
@@ -0,0 +1,22 @@
+module Medicaid
+  class IntroMaritalStatusIndicateSpouse < Step
+    step_attributes(
+      :member_id,
+      :spouse_id,
+      :spouse,
+    )
+
+    def valid?
```
disambiguator: this should probably be added as a validator instead of overriding activemodel validations

i.e.

```
validate_presence_of :spouse, message: "Make sure you select a person"
```

app/controllers/medicaid/intro_marital_status_indicate_spouse_controller.rb
disambiguator: also this would fix the codeclimate code duplication

jessieay: we need to override `update` here because we are updating `current_member`on L10 rather than `current_application`. Or at least, I think it is clearer to override it completely since we are customizing it in multiple ways rather than just in a single way, which is how we usually use ` after_successful_update_hook`. Thoughts?

app/controllers/medicaid/member_steps_controller.rb
jessieay: yeah for sure! the reason I moved it out was so I could test it like we do for most of our controllers. What are your thoughts on those tests, are they providing value or should we keep `next_path` private?

app/models/medicaid_application.rb
jessieay: in my mind it's small enough that it is not required and I dont think we have tests for similar methods in this class, but I am always happy to write more unit tests. do you think we should?

app/steps/medicaid/intro_marital_status_indicate_spouse.rb
jessieay: The way we have this working, we show the error on `spouse` but the validation is on `spouse_id` -- the reason for this is that each checkbox represents a `spouse_id` field, so if we returned an error on that, we'd have a list of errors.

This way, the error just shows up once. A little weird, but it's the way we've validated all of these lists of checkboxes for other steps so it seems to work ok. Does that make sense?

app/controllers/medicaid/intro_marital_status_indicate_spouse_controller.rb
disambiguator: I messed up. Here's the actual code I'm proposing.

```
    def after_successful_update_hook
      current_member.update(step_params)
      update_spouse_of_other_member_if_present
    end

    def application_params
      {}
    end
```

app/models/medicaid_application.rb
disambiguator: Cool, sounds good to me. Just wanted to hear your thoughts. Nbd, maybe in the future if we modify this.

app/steps/medicaid/intro_marital_status_indicate_spouse.rb
disambiguator: Oh, I see, the error shows up on spouse. Ok, I'm mostly just worried about overriding the default valid? method. How about this?

```
module Medicaid
  class IntroMaritalStatusIndicateSpouse < Step
    step_attributes(
      :member_id,
      :spouse_id,
      :spouse,
    )
    
    validate :spouse_present

    def spouse_present
      if spouse_id.blank?
        errors.add(:spouse, "Make sure you select a person")
      end
    end

    def member
      @_member ||= Member.find(member_id)
    end
  end
end
```

app/controllers/medicaid/member_steps_controller.rb
disambiguator: Got it - I was confused before and was conflating a few things. looks good

PR #450, opened 2017-11-09
Ask for employer name + pay frequency for each job

jayroh:
For however many jobs a person has we must ask who the employer is, and
how often they are paid by that company.

* Change the select field type in the form builder to allow for multiple
  values in the database

### DIFF
--- a/app/steps/medicaid/amounts_income.rb
+++ b/app/steps/medicaid/amounts_income.rb
@@ -18,53 +18,71 @@ module Medicaid
       to: :member,
     )
 
-    validate :incomes
-    validate :employer_names
-    validate :payment_frequencies
-    validate :unemployment_income_provided
-    validate :self_employment_income
+    with_options if: :has_jobs? do |employed|
+      employed.validate :incomes
+      employed.validate :employer_names
+      employed.validate :payment_frequencies
+    end
+
+    validate :unemployment_income_provided, if: :receives_unemployment_income?
+    validate :self_employment_income, if: :self_employed?
 
     private
 
+    def has_jobs?
+      employed_number_of_jobs&.positive?
+    end
+
     def incomes
-      return true if employed_number_of_jobs == employed_monthly_income&.count
+      return true if job_and_income_counts_match? && present_incomes?
       errors.add(:incomes_missing, "Make sure all jobs have an income amount")
     end
 
+    def job_and_income_counts_match?
+      employed_number_of_jobs == employed_monthly_income&.count
+    end
+
+    def present_incomes?
+      employed_monthly_income&.all?(&:present?)
+    end
+
     def employer_names
-      return true if employed_number_of_jobs == employed_monthly_employer&.count
+      return true if job_and_employer_counts_match? && present_employer_names?
       errors.add(:incomes_missing, "Make sure all jobs have an employer name")
     end
 
-    def payment_frequencies
-      if employed_number_of_jobs == employed_payment_frequency&.count
-        return true
-      end
+    def job_and_employer_counts_match?
+      employed_number_of_jobs == employed_monthly_employer&.count
+    end
 
+    def present_employer_names?
+      employed_monthly_employer&.all?(&:present?)
+    end
+
+    def payment_frequencies
+      return true if job_and_frequency_counts_match? && present_frequencies?
       errors.add(
         :incomes_missing,
         "Make sure all jobs have a payment frequency",
       )
     end
 
-    def unemployment_income_provided
-      return true if unemployed_and_unemployment_income_provided?
-      errors.add(:incomes_missing, "Make sure to provide unemployment income")
+    def job_and_frequency_counts_match?
+      employed_number_of_jobs == employed_payment_frequency&.count
     end
 
-    def unemployed_and_unemployment_income_provided?
-      !receives_unemployment_income? ||
-        (receives_unemployment_income? && unemployment_income.present?)
+    def present_frequencies?
+      employed_payment_frequency&.all?(&:present?)
+    end
+
+    def unemployment_income_provided
+      return true if unemployment_income.present?
+      errors.add(:incomes_missing, "Make sure to provide unemployment income")
     end
 
     def self_employment_income
-      return true if self_employment_income_provided?
+      return true if self_employed_monthly_income.present?
       errors.add(:incomes_missing, "Make sure to provide unemployment income")
-    end
-
-    def self_employment_income_provided?
-      !self_employed? ||
-        (self_employed? && self_employed_monthly_income.present?)
     end
   end
 end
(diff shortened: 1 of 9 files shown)


app/steps/medicaid/amounts_income.rb
```diff
@@ -14,36 +18,53 @@ class AmountsIncome < MemberStep
...
-    def unemployment_income_is_not_provided?
-      receives_unemployment_income? && unemployment_income.blank?
+    def self_employment_income
+      return true if self_employment_income_provided?
+      errors.add(:incomes_missing, "Make sure to provide unemployment income")
     end
 
-    def self_employment_income_not_provided?
-      self_employed? && self_employed_monthly_income.blank?
+    def self_employment_income_provided?
```
jessieay: same comment as above re: method will be simplified if we check for invalid states and then return an error rather than checking for a valid state and returning true

app/steps/medicaid/amounts_income.rb
```diff
@@ -14,36 +18,53 @@ class AmountsIncome < MemberStep
...
+    validate :incomes
+    validate :employer_names
+    validate :payment_frequencies
+    validate :unemployment_income_provided
+    validate :self_employment_income
+
+    private
+
+    def incomes
+      return true if employed_number_of_jobs == employed_monthly_income&.count
```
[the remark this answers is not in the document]
jessieay: if we do this, we can also change the names of some of these methods to be the opposite, which might clarify things! eg:

```
def unemployment_income_provided
  if unemployed_and_unemployment_income_not_provided?
    errors.add(:incomes_missing, "Make sure to provide unemployment income")
  end  
end

def unemployed_and_unemployment_income_not_provided?
  receives_unemployment_income? && unemployment_income.blank?
end
```

app/views/medicaid/amounts_income/edit.html.erb
```diff
@@ -26,11 +26,30 @@
...
-         :employed_monthly_income,
-         "Monthly income (before taxes)",
-         options: { count: current_member.employed_number_of_jobs },
-         prepend_html: "<p class=\"text--section-header\">Current Job ({index})</p>"
-       ) %>
+          :employed_monthly_income,
+          "Monthly income (before taxes)",
+          options: { count: current_member.employed_number_of_jobs },
+          prepend_html:
+            "<p class=\"text--section-header\">Income - Job ({index})</p>"
```
jessieay: I think we want a `#` sign before the index here

app/views/medicaid/amounts_income/edit.html.erb
```diff
@@ -26,11 +26,30 @@
...
+          :employed_monthly_employer,
+          "Employer name",
+          options: { count: current_member.employed_number_of_jobs },
+          prepend_html:
+          "<p class=\"text--section-header\">Employer Name - Job ({index})"
+        ) %>
+
+        <%= f.mb_select(
+          :employed_payment_frequency,
+          "How often are you paid that amount? - Job ({index})",
```
jessieay: should this also be prepended with a `p` tag like the other fields? (for the blue header thingy). May want to ask design

db/migrate/20171109155919_add_employer_information_to_members.rb
```diff
@@ -0,0 +1,14 @@
+class AddEmployerInformationToMembers < ActiveRecord::Migration[5.1]
+  def up
+    add_column :members, :employed_monthly_employer, :string, array: true
```
jessieay: monthly employer?

jessieay: maybe it should be `employed_employer_name` ?

spec/features/medicaid_application_with_maximum_info_spec.rb
```diff
@@ -110,6 +110,15 @@
       fill_in "step_employed_monthly_income_0", with: 100
       fill_in "step_employed_monthly_income_1", with: 50
       fill_in "step_employed_monthly_income_2", with: 25
+
+      fill_in "step_employed_monthly_employer_0", with: "Acme Co."
+      fill_in "step_employed_monthly_employer_1", with: "CfA"
+      fill_in "step_employed_monthly_employer_2", with: "Cylinder"
+
+      select "Monthly", from: "step_employed_payment_frequency_0"
```
jessieay: should we test with a value other than `Monthly`?

spec/features/medicaid_application_with_maximum_info_spec.rb
```diff
@@ -119,6 +128,11 @@
       expect(find("#step_employed_monthly_income_0").value).to eq "100"
       expect(find("#step_employed_monthly_income_1").value).to eq "50"
       expect(find("#step_employed_monthly_income_2").value).to eq "25"
+
+      expect(find("#step_employed_monthly_employer_0").value).to eq "Acme Co."
+      expect(find("#step_employed_monthly_employer_1").value).to eq "CfA"
+      expect(find("#step_employed_monthly_employer_2").value).to eq "Cylinder"
```
jessieay: should we add expectations for the payment interval as well?

app/steps/medicaid/amounts_income.rb
```diff
@@ -14,36 +18,53 @@ class AmountsIncome < MemberStep
...
-        false
-      elsif self_employment_income_not_provided?
-        errors.add(
-          :incomes_missing,
-          "Make sure to provide self-employment income",
-        )
-        false
-      else
-        true
+    validate :incomes
```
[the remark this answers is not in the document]
hartsick: ooo

jayroh:
The subtitle and labels here seem a little redundant, no?

 income___expense_amounts

db/migrate/20171109155919_add_employer_information_to_members.rb
jayroh: There's already a field called `employed_employer_name` unfortunately (that's for SNAP purposes) and it's a single string type, not an array of strings. So it muddies this up a little bit.

All this smells sufficiently like something that should end up being extracted to its own model. `Job`? `MemberJob`?

Also - I called it `employed_monthly_employer` so that it stays as close to the income amount column as possible. It's not the perfect name but it's similar enough to the income column where you can see they're related.

app/views/medicaid/amounts_income/edit.html.erb
jayroh: Yeah I put a screenshot down below. We can get the design team to chime in on it.

db/migrate/20171109155919_add_employer_information_to_members.rb
disambiguator: @jessieay Made a ticket! https://www.pivotaltracker.com/story/show/152765054

PR #451, opened 2017-11-10
Update `amounts-expenses` page for multi-member flow

jessieay:
* [Finishes #152510521]
* We did not add any validations to these inputs, just changed the
wording and design and in some cases moved the field to the member class

screen shot 2017-11-09 at 4 05 00 pm

screen shot 2017-11-09 at 4 04 32 pm

### DIFF
--- a/app/controllers/medicaid/amounts_expenses_controller.rb
+++ b/app/controllers/medicaid/amounts_expenses_controller.rb
@@ -1,9 +1,45 @@
 module Medicaid
-  class AmountsExpensesController < MedicaidStepsController
+  class AmountsExpensesController < Medicaid::MemberStepsController
+    def current_member
+      @_current_member ||= super || first_member_with_expenses
+    end
+
+    def update
+      @step = step_class.new(
+        step_params.merge(member_id: current_member.id),
+      )
+
+      if @step.valid?
+        current_member.update!(step_params)
+        redirect_to(next_path)
+      else
+        render :edit
+      end
+    end
+
     private
 
     def step_class
       Medicaid::AmountsExpenses
+    end
+
+    def first_member_with_expenses
+      current_application.
+        members.
+        with_expenses.
+        limit(1).
+        first
+    end
+
+    def next_member
+      return if current_member.nil?
+
+      current_application.
+        members.
+        with_expenses.
+        after(current_member).
+        limit(1).
+        first
     end
 
     def skip?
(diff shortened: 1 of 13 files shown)


app/controllers/medicaid/amounts_expenses_controller.rb
```diff
@@ -1,11 +1,47 @@
 module Medicaid
-  class AmountsExpensesController < MedicaidStepsController
+  class AmountsExpensesController < Medicaid::MemberStepsController
+    def current_member
+      @_current_member ||= super || first_member_with_expenses
+    end
+
+    def update
```
hartsick: one thing @disambiguator and I had been starting to try was using the `after_successful_update_hook` and defining model-specific params from the step_params (e.g. `member_params` that could just be `step_params.merge(member_id: current_member.id)`). the idea behind that is that it exercises our existing interface, surfaces pain points, and allows us to have a contained site for change in future. may be worth trying here?

app/controllers/medicaid/expenses_alimony_controller.rb
```diff
@@ -1,4 +1,18 @@
 module Medicaid
   class ExpensesAlimonyController < MedicaidStepsController
+    private
+
+    def after_successful_update_hook
```
hartsick: woo

db/migrate/20171109193950_add_expense_amounts_to_members.rb
```diff
@@ -0,0 +1,55 @@
...
+           child_support_alimony_arrears_expenses=primary_medicaid_members.child_support_alimony_arrears_expenses,
+           self_employment_expenses=primary_medicaid_members.self_employed_monthly_expenses
+        FROM (
+          SELECT min(id) as id, benefit_application_id, unemployment_income
+          FROM members
+          WHERE benefit_application_type = 'MedicaidApplication'
+          GROUP BY benefit_application_id, unemployment_income
+        )
+        AS primary_medicaid_members
+        WHERE medicaid_applications.id = primary_medicaid_members.benefit_application_id
```
hartsick: these are so wild 🎈

jessieay:
specs should be green now. Feel free to merge tomorrow if it gets approved @disambiguator !

app/controllers/medicaid/amounts_expenses_controller.rb
jayroh: I like the idea of `member_params`. One slight suggestion I have is to give it one nudge towards more specificity and call it `member_step_params`. That'd keep the `step` context in line.

disambiguator: @jayroh Dig it.

We should move in that direction, although on second look this one deviates just ever so slightly from that pattern and we can't just do the quick refactor right now. Gonna save this one for later.

PR #454, opened 2017-11-10
Ch admin office location not shown

hartsick:
* Fixes broken seeds and creates a local Admin User
* Adds office_location to admin view for SnapApplication!

Fixes this bug: https://www.pivotaltracker.com/story/show/152566845

screen shot 2017-11-09 at 5 09 17 pm

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -37,7 +37,7 @@ minimal_application.update!(
 )
 
 minimal_mailing = Address.find_or_initialize_by(
-  snap_application: minimal_application,
+  benefit_application: minimal_application,
 )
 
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -1,3 +1,9 @@
+puts "Creating admin user"
```
jessieay: nice!

thoughts on putting together a quick smoke test for the seeds just to make sure they run OK? I feel like we've broken them a few times recently and not caught it until later on

hartsick: yeah, I was thinking about that, too! I think we should. I'll make a chore

PR #457, opened 2017-11-10
New pattern for our update inheritance structure

disambiguator:
* Still can update the current application with step params with no
 boilerplate
 * Add additional behavior to the happy path update with a call to super
 * Replace the call entirely without needing a noop update

Signed-off-by: Joel Oliveira <joel.oliveira@gmail.com>

I don't particularly care if we merge this, but I wanted to show an example for a new style of inheritance, one where we can explicitly show which fields go from the step to a specific db model.
This is going to be more important as we move fields out of the application model.

`app/controllers/residential_address_controller.rb` shows a good example of this transformation.

Let's discuss!

@jayroh @jessieay @hartsick @bengolder

### DIFF
--- a/app/controllers/medicaid/contact_home_address_controller.rb
+++ b/app/controllers/medicaid/contact_home_address_controller.rb
@@ -26,7 +26,8 @@ module Medicaid
       !current_application&.stable_housing?
     end
 
-    def after_successful_update_hook
+    def update_application
+      current_application.update!(application_params)
       address.update!(address_params)
     end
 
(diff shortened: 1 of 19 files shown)


jayroh:
From a comprehension perspective, I like this _a lot_. It's terse, tells you what it's doing, very straightforward.

hartsick:
Yeah, I dig this! Think it's very understandable, and provides flexibility as we figure out more about our use cases. I'd be pro merging.

One thought that this diff brings up: we have a lot of cases in the code where we use:
```
if single_member_household?
   current_application.primary_member.update!(member_attrs)		          
end
```
Would love to think about a way to cleanly consolidate this.

PR #458, opened 2017-11-13
Devise route constraints for admin-only DJ access

jayroh:
Using devise's built in route helpers we can restrict access to both
delayed job AND the admin. This allows users who visit the admin or DJ
to be forced to log in first, and then be redirected back to their
intended destination.

More information can be found here:

  https://github.com/plataformatec/devise/wiki/How-To:-Define-resource-actions-that-require-authentication-using-routes.rb

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -2,8 +2,6 @@
 # We use a different name because RAILS_ENV is overloaded and is technically the
 # same across each release stage.
 
-ADMIN_PASSWORD="password"
-ADMIN_USER="admin"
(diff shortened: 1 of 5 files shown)


hartsick:
Could you add a simple feature test that asserts we need to have an admin user authenticated to access delayed job, like we currently have for the admin path?

hartsick:
Also, if we aren't using ADMIN_USER and ADMIN_PASSWORD anymore (I think we aren't with this PR!), could we remove from the `app.json`?

PR #459, opened 2017-11-13
Log DelayedJob access

jayroh:
Using some rack middleware we can log the actions of whoever is
accessing the DelayedJob web admin.

[Finishes #152762754]


hartsick:
Looks good! Could we add a simple feature test for this to verify that delayed job is protected by admin user auth?

PR #463, opened 2017-11-13
Refactor validations to never override valid?

disambiguator:
* Use custom validation methods instead

[#152736999]

One big thing I want to flag here is the inheritance structure for validations in the ManyMembersStep. There are two types of validations here:
  * Checking if every member has a certain condition
  * Checking if at least one member has a certain condition

As a result, half of the child controllers of ManyMembersStepController use the inheritance structure, the other half just override the parent validator entirely. It feels like inheritance isn't particularly helping us here.

Also, I want to flag that we have certain questions where we do attribute nesting which seems a bit convoluted. For example: HealthDisabilityMemberStep.
It asks one question with member ids as choices. We have a data model of:
step.members[].disabled, where disabled is a boolean
instead of:
step.disabled, where disabled is an id.

If we had the latter, we could simply do a validates_presence_of :disabled, instead of needing a custom validation method. It would also let us do a single update, as so:
```
Member.find(step[:disabled]).update!(disabled: true)
```

Instead of what we have now, where we do an update on _every_ member of the application, where all but one is a noop:
```
def assign_household_member_attributes
    step.members.each do |member|
      attrs = params.dig(:step, :members, member.to_param)

      if attrs.present?
        member.assign_attributes(attrs.permit(member_attrs))
      end
    end
  end

assign_household_member_attributes
ActiveRecord::Base.transaction { step.members.each(&:save!) }
```


app/steps/many_members_step.rb
```diff
@@ -1,13 +1,16 @@
 class ManyMembersStep < Step
   step_attributes :members
 
-  def valid?
+  validate :members_valid
+
+  def members_valid
```
jessieay: instead of having default behavior, should we just `raise NotImplementedError` here? as far as I can tell, we are not using the default behavior defined in this parent step in any of its children

app/steps/medicaid/income_other_income_type.rb
```diff
@@ -11,15 +11,14 @@ class IncomeOtherIncomeType < Step
       :member_id,
     )
 
-    def valid?
+    validate :income_type_selected
+
+    def income_type_selected
```
jessieay: would it be more consistent to name this `members_valid` like we have in many of the other steps?

app/steps/medicaid/intro_college_member.rb
```diff
@@ -5,13 +5,11 @@ class IntroCollegeMember < Step
...
-    def valid?
-      if members.select(&:in_college?).any?
-        true
-      else
-        errors.add(:in_college, "Make sure you select at least one person")
-        false
-      end
+    validate :in_college_selected
+
+    def in_college_selected
```
jessieay: `members_valid` instead?

app/steps/medicaid/intro_marital_status_member.rb
```diff
@@ -5,13 +5,11 @@ class IntroMaritalStatusMember < Step
...
-    def valid?
-      if members.select(&:married?).any?
-        true
-      else
-        errors.add(:married, "Make sure you select at least one person")
-        false
-      end
+    validate :married_selected
+
+    def married_selected
```
jessieay: `members_valid` instead?

app/steps/medicaid/insurance_needed.rb
```diff
@@ -5,16 +5,14 @@ class InsuranceNeeded < Step
...
-      else
-        errors.add(
-          :requesting_health_insurance,
-          "Make sure you select at least one person",
-        )
-        false
-      end
+    validate :members_requesting_health_insurance_selected
+
+    def members_requesting_health_insurance_selected
```
jessieay: `members_valid` instead?

app/steps/medicaid/insurance_current_type.rb
```diff
@@ -5,12 +5,11 @@ class InsuranceCurrentType < MemberStep
       :member_id,
     )
 
-    def valid?
+    validate :insurance_plan_selected
+
+    def insurance_plan_selected
```
jessieay: `members_valid` instead?

app/steps/many_members_step.rb
```diff
@@ -1,13 +1,16 @@
...
     members.each do |member|
       member.errors.clear
       validate_household_member(member)
```
jessieay: are we even using `validate_household_member` anymore?

disambiguator:
@jessieay I went ahead and removed the inheritance from any place that was overriding the members_valid method. Now only 3 controllers, inherit, all of whom use the pattern in exactly the same way. Let me know what you think!

app/steps/medicaid/citizen_member.rb
```diff
@@ -1,14 +0,0 @@
-module Medicaid
```
jessieay: did you mean to delete this step?

disambiguator: Yup!

Looks like this is left over from some previous refactor and is duplicate code.
https://github.com/codeforamerica/michigan-benefits/blob/master/app/steps/medicaid/citizen_member.rb
https://github.com/codeforamerica/michigan-benefits/blob/master/app/steps/medicaid/intro_citizen_member.rb

PR #464, opened 2017-11-13
Add 'Sign Out' link for Admin dashboard

hartsick:
Accepting some stories about Admin User authentication was difficult, because we don't have an easy way to sign out. To address this, I added a "Sign Out" link to the admin dashboard that destroys the user session upon clicking.

https://www.pivotaltracker.com/story/show/152826538
[Finishes #152826538]

screen shot 2017-11-13 at 1 41 12 pm

screen shot 2017-11-13 at 1 41 17 pm

### DIFF
--- a/app/views/admin/application/_navigation.html.erb
+++ b/app/views/admin/application/_navigation.html.erb
@@ -1,0 +1,24 @@
+<%#
+# Navigation
+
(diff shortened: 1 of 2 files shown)


hartsick:
@jessieay I think Devise defaults to sending them back to the root path, so that's where they land now

PR #466, opened 2017-11-14
Fill out Step 1 of 1426 PDF

jessieay:
* Mailing address, residential address, phone email
* Note that I left "language" blank since we do not collect that info
* Note that I renamed a few fields in the 1426 PDF so that it is more consistent with the 1171
* [Finishes #152564632]
* https://www.pivotaltracker.com/story/show/152564632

screen shot 2017-11-13 at 5 10 11 pm


spec/models/medicaid_application_attributes_spec.rb
```diff
@@ -14,7 +26,48 @@
...
+        mailing_address_county: "Genesee",
+        mailing_address_state: "MI",
+        mailing_address_street_address: "123 Main St.",
+        mailing_address_zip: "12345",
+        residential_address_city: "Flint",
+        residential_address_county: "Genesee",
+        residential_address_state: "MI",
+        residential_address_street_address: "456 I Live Here Ave",
+        residential_address_zip: "12345",
+        phone_number_0: "0",
```
jessieay: note that I am not testing that these `phone_` fields are filled out properly in the actual PDF...but I could. Not sure how much to test there. Here we are confirming that the hash is formed properly but we will not know that the PDF fields are named properly for sure unless we test everything in the `Dhs1426Pdf` class...

hartsick: I feel like we should move toward testing everything in the PDF classes, since that's the only time we pull it out of the PDF and that's the important part. I feel like this is more important to test the translation of application attributes to fields (if there's any logic), but realize that leaves the tests pretty thin.

basically, I'm not a big fan of duplicating these attribute tests, but also feel like right now we're not thoroughly testing the most important part.

app/models/medicaid_application_attributes.rb
```diff
@@ -6,10 +6,57 @@ def initialize(medicaid_application:)
...
+
+  def residential_or_homeless
+    if medicaid_application.stable_housing?
+      medicaid_application.residential_address.street_address
+    else
+      "Homeless"
+    end
+  end
+
+  def phone_attributes
```
hartsick: does it make sense to move this/bool_to_checkbox/other shared methods into a shared helper across attribute classes?

jessieay: done for at least Medicaid/Snap at 8c38fa05

PR #471, opened 2017-11-14
Update release script to show diff of env var keys

bengolder:
https://www.pivotaltracker.com/story/show/152831563

- Edits `bin/release` to show a diff of keys between stage and production.
- Does _not_ show values of env vars
- Uses `jq` CLI to separate out the keys, then bash to diff arrays of keys
- Adds `brew install jq` install to `bin/setup`
- Edits last step of `bin/release` which falsely said it was deploying if you answerered `"N"` to the final question.

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -26,7 +26,7 @@ fi
 
 echo "Checking for buildpack changes between staging and production."
 diff <(heroku buildpacks -r staging) <(heroku buildpacks -r production) && echo ""
-read -p "Are these the same (ignoring order)? " -n 1 -r
+read -p "Are these the same (ignoring order)? (y/N)" -n 1 -r
 echo    # (optional) move to a new line
 if [[ $REPLY =~ ^[Yy]$ ]];then
   echo "Great!"
@@ -36,8 +36,11 @@ else
 fi
 
 echo "Diffing ENV variables between staging and production."
-diff <(heroku config -r staging) <(heroku config -r production) && echo ""
-read -p "Are these config variables the same (ignoring their value) (y/N) " -n 1 -r
+staging_env=$(heroku config -r staging --json | jq . -S | jq 'keys')
+production_env=$(heroku config -r production --json | jq . -S | jq 'keys')
+echo ${staging_env[@]} ${production_env[@]} | tr ' ' '\n' | sort | uniq -u
+echo "Any keys listed above exist on one env but not the other."
+read -p "Are you okay with these differences? (y/N)" -n 1 -r
 echo    # (optional) move to a new line
 if [[ $REPLY =~ ^[Yy]$ ]]; then
   echo "Great!"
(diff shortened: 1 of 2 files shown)


bin/setup
```diff
@@ -54,6 +54,17 @@ chdir APP_ROOT do
     end
   end
 
+  step "Installing jq, a JSON CLI used by our release script" do
+    # use --version flag because running `chromedriver` will start
+    # chromedriver, which stops the script indefinitely if already installed.
+    if cli_installed?("jq")
```
hartsick: woo

bin/release
```diff
@@ -36,8 +36,11 @@ else
 fi
 
 echo "Diffing ENV variables between staging and production."
-diff <(heroku config -r staging) <(heroku config -r production) && echo ""
-read -p "Are these config variables the same (ignoring their value) (y/N) " -n 1 -r
+staging_env=$(heroku config -r staging --json | jq . -S | jq 'keys')
```
hartsick: looks like we don't need the jq . portion!

bin/release
```diff
@@ -36,8 +36,11 @@ else
...
+production_env=$(heroku config -r production --json | jq . -S | jq 'keys')
+echo ${staging_env[@]} ${production_env[@]} | tr ' ' '\n' | sort | uniq -u
+echo "Any keys listed above exist on one env but not the other."
```
hartsick: may want to be explicit about exist on staging but not on prod

hartsick: and maybe have a note saying "if this is unexpected, make sure to add the environment variable to production"

bin/release
```diff
@@ -52,7 +55,10 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
   heroku pipelines:promote -a michigan-benefits-staging
   heroku run bin/rails one_offs:run_all -a michigan-benefits-production
   heroku ps:restart -a michigan-benefits-production
+  echo "Success! Application deployed to Production!"
+  echo "Please do some testing on production! (use zip 12345 to send a fake fax)"
+else
+  echo "Okay, I won't deploy. Thank you for being thoughtful and careful!"
```
hartsick: <3

jayroh:
Possible to past a before and after PR to compare how we were diffing this?  I remember how it's _been_ but I'm curious how it measures up to the new approach.

Awesome to see this being smoothed out, by the way. It was virtually impossible to tell what the differences were before.

hartsick:
@jayroh Yeah, good call! 

Here's the censored version of before (values cropped out)
screen shot 2017-11-14 at 11 52 35 am

And here's the after!
screen shot 2017-11-14 at 11 53 45 am

bin/setup
disambiguator: tempted to use a Brewfile for dependencies like this

PR #473, opened 2017-11-14
Add income & expenses intro page

jessieay:
* [Finishes #152656208]
* `/steps/medicaid/income-introduction`

screen shot 2017-11-14 at 11 59 15 am

### DIFF
--- a/app/controllers/medicaid/income_introduction_controller.rb
+++ b/app/controllers/medicaid/income_introduction_controller.rb
@@ -1,0 +1,11 @@
+module Medicaid
+  class IncomeIntroductionController < MedicaidStepsController
+    def step_class
+      NullStep
+    end
+
+    def update
+      redirect_to(next_path)
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


app/controllers/medicaid/income_introduction_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Medicaid
+  class IncomeIntroductionController < MedicaidStepsController
+    def step_class
+      NullStep
+    end
+
+    def update
```
disambiguator: This looks like it inherits a ton of functionality from MedicaidStepsController just so that it can get the `next_path` method, and then overrides the rest of it. Makes me a little sad.

jessieay: also uses `edit` and `current_application` from there 😄

disambiguator: yeah, but all of edit is unused, right? Would it break if you had a `def edit; end` there? It's a static page with no logic except for next_path.

jessieay: the main benefit we are getting from inheritance here is `current_application`. This is the way we are doing all of the "breather" pages so far -- do you think that re-doing the way do structure breather pages should block the merging of this PR?

PR #475, opened 2017-11-14
Allow 2-factor authentication for admins

jayroh:
[Finishes #152368004]

Using a fork of the `devise-otp` gem here:

  https://github.com/pynixwang/devise-otp

... we are able to integrate with Google Authenticator to add an added
layer of security for admins logging into the app.

After signing in, as usual, and heading to `/admin` you can see `2FA`
link in the left navigation where you can enable it and scan a QR code
with the Google Authenticator app:

* https://itunes.apple.com/us/app/google-authenticator/id388497605?mt=8
* https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en

Once enabled, every subsequent sign-in will require that your 2FA code
is provided after the usual email and password step.

### DIFF
--- a/app/views/admin/application/_navigation.html.erb
+++ b/app/views/admin/application/_navigation.html.erb
@@ -21,4 +21,9 @@ as defined by the routes in the `admin/` namespace
     method: :delete,
     class: "navigation__link navigation__link--sign-out",
   ) %>
+  <%= link_to(
+    "2FA Setup",
+    admin_user_otp_token_path,
+    class: "navigation__link navigation__link--sign-out",
+  ) %>
 </nav>
--- a/app/views/devise_otp/tokens/show.html.erb
+++ b/app/views/devise_otp/tokens/show.html.erb
@@ -1,0 +1,24 @@
+<h2><%= t "title", { scope: "devise.otp.tokens" } %></h2>
+<p><%= t "explain", { scope: "devise.otp.tokens" } %></p>
+
+<%= form_for resource, as: resource_name, url: [resource_name, :otp_token], html: { method: :put } do |f| %>
+  <%= devise_error_messages! %>
+
+  <h3><%= t "enable_request", { scope: "devise.otp.tokens" } %></h3>
+
+  <p>
+    <%= f.label :otp_enabled, t("status", { scope: "devise.otp.tokens" }) %>
+    <br />
+    <%= f.check_box :otp_enabled %>
+  </p>
+
+  <p>
+    <%= f.submit I18n.t("submit", {:scope => "devise.otp.tokens"}) %>
+    <%= link_to "Back to Admin", admin_root_path %>
+  </p>
+<% end %>
+
+<%- if resource.otp_enabled? %>
+  <%= render partial: "token_secret" if resource.otp_enabled? %>
+  <%= render partial: "trusted_devices" if trusted_devices_enabled? %>
+<% end %>
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -12,4 +12,34 @@ Devise.setup do |config|
   config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
   config.reset_password_within = 6.hours
   config.sign_out_via = :delete
+
+  # ==> Devise OTP Extension
+  # Configure OTP extension for devise
+
+  # OTP is mandatory, users are going to be asked to
+  # enroll OTP the next time they sign in, before they can successfully complete the session establishment.
+  # This is the global value, can also be set on each user.
+  config.otp_mandatory = !Rails.env.test?
+
+  # Drift: a window which provides allowance for drift between a user's token device clock
+  # (and therefore their OTP tokens) and the authentication server's clock.
+  # Expressed in minutes centered at the current time. (Note: it's a number, *NOT* 3.minutes )
+  # config.otp_drift_window = 3
+
+  # Users that have logged in longer than this time ago, are going to be asked their password
+  # (and an OTP challenge, if enabled) before they can see or change their otp informations.
+  # config.otp_credentials_refresh = 15.minutes
+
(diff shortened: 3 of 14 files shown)


spec/features/admin_user_delayed_job_spec.rb
```diff
@@ -12,16 +12,15 @@
...
       create(:admin_user, email: "test@example.com", password: "password")
       visit "/delayed_job"
 
       fill_in "Email", with: "test@example.com"
       fill_in "Password", with: "password"
       click_on "Log in"
 
-      expect(page).to have_content "The list below shows an overview "\
-        "of the jobs in the delayed_job queue"
+      expect(page).to have_content("Signed in successfully")
```
jayroh: Note that this behavior changed because of what I would consider a bug in the `devise-otp` gem. The create action does not redirect to the saved "return to" path that is saved in session. For now I've adjusted the test instead of monkey-patching, forking, or creating a PR to the original project.

spec/features/admin_user_login_spec.rb
```diff
@@ -5,12 +5,11 @@
...
     visit admin_root_path
-    expect(current_path).to eq "/admin_users/sign_in"
-
     fill_in "Email", with: admin_user.email
     fill_in "Password", with: admin_user.password
     click_on "Log in"
 
-    expect(current_path).to eq admin_root_path
+    expect(current_path).to eq "/"
+    expect(page).to have_content("Signed in successfully")
```
jayroh: See above:

> Note that this behavior changed because of what I would consider a bug in the `devise-otp` gem. The create action does not redirect to the saved "return to" path that is saved in session. For now I've adjusted the test instead of monkey-patching, forking, or creating a PR to the original project.

app/views/devise_otp/tokens/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+
+  <p>
+    <%= f.label :otp_enabled, t("status", { scope: "devise.otp.tokens" }) %>
+    <br />
+    <%= f.check_box :otp_enabled %>
+  </p>
+
+  <p>
+    <%= f.submit I18n.t("submit", {:scope => "devise.otp.tokens"}) %>
+    <%= link_to "Back to Admin", admin_root_path %>
```
jayroh: ☝️ This is the only line I changed (other than formatting).

The behavior of updating the setting and reloading this page was a little confusing, so I added the option to click back and return to the admin.

app/views/admin/application/_navigation.html.erb
```diff
@@ -21,4 +21,9 @@ as defined by the routes in the `admin/` namespace
     method: :delete,
     class: "navigation__link navigation__link--sign-out",
   ) %>
+  <%= link_to(
```
jessieay: in the tracker story the feature reads as 

```
WHEN I visit /admin and log in
THEN I should be prompted to complete two-factor authentication
```

my guess here is that someone can opt into 2FA but is not required to use it -- is there a way to make it required per the language used in the story?

config/initializers/devise.rb
```diff
@@ -12,4 +12,34 @@
...
   config.reset_password_within = 6.hours
   config.sign_out_via = :delete
+
+  # ==> Devise OTP Extension
+  # Configure OTP extension for devise
+
+  # OTP is mandatory, users are going to be asked to
+  # enroll OTP the next time they sign in, before they can successfully complete the session establishment.
+  # This is the global value, can also be set on each user.
+  # config.otp_mandatory = false
```
jessieay: set to `true` ?

db/migrate/20171114160332_devise_otp_add_to_admin_users.rb
```diff
@@ -0,0 +1,32 @@
+class DeviseOtpAddToAdminUsers < ActiveRecord::Migration[5.1]
+  def up
```
jessieay: seems like this could have been a `def change` ?

jayroh: Tried that, no dice. So I'll keep the up/down methods here.

See below:

```
== 20171114160332 DeviseOtpAddToAdminUsers: reverting =========================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:



This migration uses execute, which is not automatically reversible.
To make the migration reversible you can either:
1. Define #up and #down methods in place of the #change method.
2. Use the #reversible method to define reversible behavior.


/home/joel/Projects/michigan-benefits/db/migrate/20171114160332_devise_otp_add_to_admin_users.rb:15:in `block in change'
/home/joel/Projects/michigan-benefits/db/migrate/20171114160332_devise_otp_add_to_admin_users.rb:3:in `change'
/home/joel/.rbenv/versions/2.4.2/bin/bundle:23:in `load'
/home/joel/.rbenv/versions/2.4.2/bin/bundle:23:in `<main>'
ActiveRecord::IrreversibleMigration:

This migration uses execute, which is not automatically reversible.
To make the migration reversible you can either:
1. Define #up and #down methods in place of the #change method.
2. Use the #reversible method to define reversible behavior.


/home/joel/Projects/michigan-benefits/db/migrate/20171114160332_devise_otp_add_to_admin_users.rb:15:in `block in change'
/home/joel/Projects/michigan-benefits/db/migrate/20171114160332_devise_otp_add_to_admin_users.rb:3:in `change'
/home/joel/.rbenv/versions/2.4.2/bin/bundle:23:in `load'
/home/joel/.rbenv/versions/2.4.2/bin/bundle:23:in `<main>'
Tasks: TOP => db:rollback
(See full trace by running task with --trace)
```

app/views/admin/application/_navigation.html.erb
jayroh: I enabled the mandatory option in the Devise initializer (enabled only for environments outside of "test")

PR #483, opened 2017-11-15
Add "Welcome" breather/intro page

jayroh:
[Finishes #152654927]

Before a client starts filling out the application we provide them with a quick "Hi!".

This commit adds that page _and_:

    * Changes the hard-coded link path in dual-index homepage to use the
      step_path helper (which is fixed to not URLencode `/`s).
    * Multiple members feature spec updated to match 2 other features. IE:
      one `on_pages` assertion instead of many `on_page`.

### DIFF
--- a/app/controllers/static_pages_controller.rb
+++ b/app/controllers/static_pages_controller.rb
@@ -1,6 +1,8 @@
 class StaticPagesController < ApplicationController
   before_action :clear_session
 
+  helper_method :step_path
+
   def index; end
 
   def union; end
@@ -8,6 +10,10 @@ class StaticPagesController < ApplicationController
   def clio; end
 
   def dual_index; end
+
+  def step_path(*args)
+    super.gsub("%2F", "/")
+  end
 
   private
 
--- a/app/steps/medicaid/step_navigation.rb
+++ b/app/steps/medicaid/step_navigation.rb
@@ -2,6 +2,7 @@ module Medicaid
   class StepNavigation
     ALL = {
       "Introduction" => [
+        Medicaid::WelcomeController,
         Medicaid::IntroLocationController,
         Medicaid::IntroLocationHelpController,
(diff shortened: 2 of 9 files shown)


app/controllers/static_pages_controller.rb
```diff
@@ -9,6 +11,10 @@ def clio; end
 
   def dual_index; end
 
+  def step_path(*args)
+    super.gsub("%2F", "/")
```
jessieay: darrrrrn that path!

app/steps/medicaid/step_navigation.rb
```diff
@@ -2,6 +2,7 @@ module Medicaid
   class StepNavigation
     ALL = {
       "Introduction" => [
+        Medicaid::WelcomeController,
         Medicaid::IntroLocationController,
```
jessieay: do we need to add something to the view template for this controller that allows the person to go "Back"? I can't remember how we do the "clear session" on back for the first step.

app/views/static_pages/dual_index.html.erb
```diff
@@ -18,7 +18,7 @@ end %>
       Apply for FAP
     <% end %>
 
-    <%= link_to "steps/medicaid/intro-location", class: "button button--cta button--fully-width" do %>
+    <%= link_to step_path(Medicaid::StepNavigation.first), class: "button button--cta button--fully-width" do %>
```
jessieay: 🌈

hartsick: love it

app/steps/medicaid/step_navigation.rb
jayroh: It automatically clears the session upon leaving. The static pages controller clears out the session info. And this first page has the "back" link enabled. Or are you saying this page should *not* have it enabled?

PR #484, opened 2017-11-15
Fix migration

jessieay:
* This migration was accidentally updated incorrectly after it was
migrated but before it was merged. This migration is currently blocking
deploys to staging because it is an invalid migration.


jessieay:
error from circle:

```
remote: == 20171115014535 AddFilingTaxesWithPrimaryMemberToMembers: migrating =========
remote: -- add_column(:members, :filing_taxes_with_primary_member)
remote: rake aborted!
remote: StandardError: An error has occurred, this and all later migrations canceled:
remote: 
remote: wrong number of arguments (given 2, expected 3..4)
remote: /app/vendor/bundle/ruby/2.4.0/gems/activerecord-5.1.4/lib/active_record/connection_adapters/postgresql/schema_statements.rb:407:in `add_column'
```

PR #487, opened 2017-11-15
Up to question 30 for 1426 PDF

jessieay:
Part of [#152727221]


db/migrate/20171115191926_rename_employed_monthly_employer_to_employer_employer_names.rb
```diff
@@ -0,0 +1,41 @@
+class RenameEmployedMonthlyEmployerToEmployerEmployerNames < ActiveRecord::Migration[5.1]
+  def up
+    add_column :members, :employed_employer_names, :string, array: true
```
jessieay: renamed these fields to make their meaning clearer.

Just had a call with @rachelcope where I confirmed that in the next iteration of the "amounts - income" page, we are going to ask for employment pay quantities and intervals, so nothing related to monthly income

db/schema.rb
```diff
@@ -161,11 +161,11 @@
     t.boolean "disabled"
     t.boolean "employed", default: false
     t.string "employed_employer_name"
+    t.string "employed_employer_names", default: [], array: true
```
bengolder: Do you think these should eventually be pulled out into a separate table?

db/migrate/20171115191926_rename_employed_monthly_employer_to_employer_employer_names.rb
bengolder: This feels like an improvement, but I still find it a little confusing to read `employed_employer`. would it be clearer if it were `job_employer_names` or `employment_employer_names`? 
It seems like a noun in place of the `employed` adjective would make it clear that this is a property of a thing, which I think is implied.

db/schema.rb
jessieay: yes perhaps! now that we have employer data that is different for SNAP and medicaid this is all getting pretty confusing.

db/migrate/20171115191926_rename_employed_monthly_employer_to_employer_employer_names.rb
jessieay: yeahhhh -- the reason I named it this way was that we prepend ever `job` field with `employed` and every self employment field with `self_employed` -- I am not sure this pattern is even that helpful at this point, but it is an existing pattern that provides some means of remembering which fields are named what. Do you think we should move away from that pattern, or perhaps do that when moving this data to a separate table?

PR #488, opened 2017-11-15
Account for `nil` birthday possibilities for 1426

jessieay:
* We do not require birthdays in the 1426 flow, so we need to write PDF
filling logic that works if there is no birthday present
* Part of https://www.pivotaltracker.com/story/show/152727221

### DIFF
--- a/app/models/medicaid_application_member_attributes.rb
+++ b/app/models/medicaid_application_member_attributes.rb
@@ -16,7 +16,8 @@ class MedicaidApplicationMemberAttributes
       :"#{position}_member_spouse_name" => member.spouse&.display_name,
       member_caretaker_key => "Yes",
       :"#{position}_member_in_college_#{yes_no(member.in_college)}" => "Yes",
-      :"#{position}_member_under_21_#{yes_no(under_21?)}" => "Yes",
+      :"#{position}_member_under_21_yes" => bool_to_checkbox(under_21?),
+      :"#{position}_member_under_21_no" => bool_to_checkbox(over_21_or_21?),
       :"#{position}_member_new_mom_#{yes_no(member.new_mom?)}" => "Yes",
       requesting_health_insurance_key => "Yes",
       :"#{position}_member_citizen_#{yes_no(member.citizen?)}" => "Yes",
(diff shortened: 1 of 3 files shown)


app/models/medicaid_application_member_attributes.rb
```diff
@@ -45,6 +46,14 @@ def requesting_health_insurance_key
   end
 
   def under_21?
-    member.age < 21
+    if member.birthday.present?
+      member.age < 21
+    end
+  end
+
+  def over_21_or_21?
```
disambiguator: Why do we need this? Since under_21? will return `nil` if the birthday isn't present, `yes_or_no` would give the correct output, ya?

app/models/member.rb
```diff
@@ -138,7 +138,7 @@ def male?
   end
 
   def formatted_birthday
-    birthday.strftime("%m/%d/%Y")
+    birthday&.strftime("%m/%d/%Y")
```
disambiguator: I'd argue you shouldn't even be calling `formatted_birthday` if `birthday` is nil. Also, if we're going to add additional logic to this method, I think we should unit test it.

disambiguator: Since we're testing this at the `medicaid_application_member_attributes_spec` level, would it make more sense to have this as:
```
if member.birthday.present?
  member_attributes[:"#{position}_member_birthday"] => member.formatted_birthday,
end
```

up higher?

app/models/medicaid_application_member_attributes.rb
jessieay: I was having a hard time sticking with `yes_or_no` here because if we have no birthday we want to return `nil` for both, and `yes_or_no` would always have us returning "Yes" for one of the options.

disambiguator: Oh RIGHT got it

Hmm, thoughts:
```
if member.birthday.present?
  member_attributes[:"#{position}_member_birthday"] => member.formatted_birthday,
  member_attributes[:"#{position}_member_under_21_#{yes_no(under_21?)}"] => "Yes",
end
```

and then nuke the presence conditional from `under_21?`

bengolder:
@disambiguator we added fixes based on your recs: cda4e76e

PR #490, opened 2017-11-15
Add health introduction breather page

jessieay:
* [Finishes #152655651]
* This is using less inheritance than other breather pages and can serve
as a model for the refactor described in
https://www.pivotaltracker.com/story/show/152867437 once this is
finalized (to follow)

### DIFF
--- a/app/controllers/medicaid/health_introduction_controller.rb
+++ b/app/controllers/medicaid/health_introduction_controller.rb
@@ -1,50 +1,7 @@
 module Medicaid
-  class HealthIntroductionController < ApplicationController
-    layout "step"
-    before_action :ensure_application_present
-    helper_method :current_application
-    helper_method :application_title
-
-    def self.to_param
-      controller_path.dasherize
-    end
-
-    private
-
-    def application_title
-      "Medicaid Application"
-    end
-
-    def ensure_application_present
-      return if current_application
-
-      redirect_to root_path
-    end
-
-    def current_application
-      MedicaidApplication.find_by(id: session[:medicaid_application_id])
-    end
-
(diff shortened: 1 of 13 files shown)


app/controllers/medicaid/health_introduction_controller.rb
```diff
@@ -0,0 +1,50 @@
+module Medicaid
+  class HealthIntroductionController < ApplicationController
```
jessieay: @disambiguator I inherited from `ApplicationController` and only included methods required for a step to work. If we are happy with this, I can update the other breather controllers

jayroh: If I'm reading this right - would the other breather controllers need everything below this comment?

jessieay: yep! every controller for a breather page would need all of these methods

jessieay: where's the :shrug: emoji when you need it!

jessieay: may be helpful to compare with the controller changes required to add a breather page *with* inheritance https://github.com/codeforamerica/michigan-benefits/pull/481/files#diff-a1c0b16dd0b1db3ce5d021c461d4edba

jayroh: > yep! every controller for a breather page would need all of these methods

That much duplication, in my opinion, is not ideal. But I'm not entirely sure what the best / right answer is to combat this outside of inheritance or another more composition-based approach. The latter of which I think would need some further discussion with the whole group

PR #493, opened 2017-11-15
Make route paths case-INsensitive

jayroh:
[Finishes #152771807]


jayroh:
All of the feature specs exercise the normal paths, I clicked around a bunch to various URL's with differing cases and it worked ok. Specs for this would be low on the ROI side, I think.

PR #497, opened 2017-11-16
Validate other_income_types

bengolder:
- Checks to ensure that any set of choices for the other_income_types
  array are a subset of predefined options
- Adds an error if not


app/models/member.rb
```diff
@@ -49,12 +58,20 @@ class Member < ApplicationRecord
     inclusion: { in: SEXES },
     allow_nil: true
 
+  validate :correct_other_income_types
```
jessieay: this is a "custom validator", which is a totally cool thing to do and in many cases necessary. In this case, rails gives us [a validator we can use](http://guides.rubyonrails.org/active_record_validations.html#inclusion):

```
 validates :other_income_types,
    inclusion: { in: OTHER_INCOME_TYPES },
    allow_nil: true

```
that "inclusion" validator will do the same thing that your custom validator is doing.

spec/models/dhs1426_pdf_spec.rb
```diff
@@ -152,40 +163,54 @@
         primary_member_self_employed: "Yes",
         primary_member_first_employed_employer_name: "AA Accounting",
         primary_member_first_employed_pay_interval_hourly: "Yes",
-        # primary_member_first_employed_hours_interval_weekly: nil,
```
jessieay: seems like these un-commenting changes are perhaps not related to this specific PR?

spec/models/member_spec.rb
```diff
@@ -164,4 +164,36 @@
       end
     end
   end
+
+  describe "#other_income_types" do
```
jessieay: to test this, we can use a cool library called `shoulda` (which we are already using) https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb

so we could replace all of these tests with:

```
it { should validate_inclusion_of(:other_income_tpes).in_array(Member::OTHER_INCOME_TYPES) }
it { should allow_value(nil).for(:other_income_types) }
```

spec/models/medicaid_application_member_attributes_spec.rb
```diff
@@ -60,6 +69,13 @@
         primary_member_second_employed_employer_name: "BB Burgers",
         primary_member_second_employed_pay_interval_weekly: "Yes",
         primary_member_second_employed_pay_quantity: "222",
+        primary_member_additional_income_none: nil,
```
jessieay: this seems unrelated to this PR?

app/models/member.rb
disambiguator: The problem with the Rails validator here is that `other_income_types` is not a normal string but a magic array type.
`t.string "other_income_types", default: [], array: true`

I don't know if you can use the inclusion validator in this case sadly.

spec/models/medicaid_application_member_attributes_spec.rb
disambiguator: Agree, there are a lot of different things happening. Hard to sift through what's what.

PR #500, opened 2017-11-16
Validate other_income_types

bengolder:
- checks to ensure that any set of choices for the other_income_types
  array are a subset of predefined options
- adds an error if not
- I don't know ruby and patched together advice from the internet.
  So please let me know how I can do better

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -64,13 +64,11 @@ class Member < ApplicationRecord
     inclusion: { in: SEXES },
     allow_nil: true
 
-<<<<<<< HEAD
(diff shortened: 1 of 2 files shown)


app/models/member.rb
```diff
@@ -49,12 +58,20 @@ class Member < ApplicationRecord
     inclusion: { in: SEXES },
     allow_nil: true
 
+  validate :correct_other_income_types
```
jessieay: could name it something like `other_income_types`

jessieay:
merge it! (but first, merge conflict)

PR #510, opened 2017-11-17
Updates to PDF to fix yes/no bug

jessieay:
* Some fields on the PDF are radios, where the "Yes" option requires
input of "Yes" and the "No" option requires an input of "No".
* We were previously treating these all like normal checkboxes, so we
were setting a value of "Yes" wherever we wanted to check a box.
* To make the "No" values check properly, I updated the PDF to look for
an answer of "Yes" for both the yes/no radios.
* Also removed all fields that we are not filling out, to reduce
confusion.
* Fixed for https://www.pivotaltracker.com/story/show/152727254

To test:
* log in as an admin on staging
* visit https://staging.michiganbenefits.org/admin/medicaid_applications/144/pdf
* confirm that on page 3, question 9, the "No" box is checked (it should
be but was not checked before this PR)


jayroh:
As a PR I'm not sure how to review this ... so I'll accept and leave this up for the acceptance part.

jessieay:
@jayroh yeah, it is super hard to review :(

PR #512, opened 2017-11-17
Always show household members in same order on checkbox pages

jessieay:
* Previously, we were using Rails default ordering, which will put the
most recently updated record at the end.
* This resulting in a different order of household members depending on
who had been "updated" recently.
* Example: I am applying as "First Person" and "Second Person" is in my
household.
* If I am married and Second Person is not, I will show up second in the
list of checkboxes on questions after the marriage question, which is
unexpected.
* There may be better ways to do this than what I have in this PR --
suggestions?

screen shot 2017-11-16 at 6 06 37 pm

screen shot 2017-11-16 at 6 07 24 pm

screen shot 2017-11-16 at 6 07 00 pm

### DIFF
--- a/app/models/medicaid_application.rb
+++ b/app/models/medicaid_application.rb
@@ -1,7 +1,13 @@
 class MedicaidApplication < ApplicationRecord
   include Submittable
 
-  has_many :members, as: :benefit_application, dependent: :destroy
+  has_many(
+    :members,
+    -> { order(created_at: :asc) },
+    as: :benefit_application,
+    dependent: :destroy,
+  )
+
   has_many :addresses, as: :benefit_application, dependent: :destroy
 
   attribute :ssn
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -82,6 +82,10 @@ class Member < ApplicationRecord
     end
   end
 
+  def self.ordered
+    order(created_at: :asc)
+  end
+
   def self.insured
     where(insured: true).
       where(requesting_health_insurance: true).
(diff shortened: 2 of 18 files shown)


jayroh:
Instead of `@step.members.ordered` in the _view_, could we push the `ordered` scope down into the step(s) themselves?

jessieay:
@hartsick @disambiguator @bengolder 

what do you think about adding a [`default scope`](http://api.rubyonrails.org/classes/ActiveRecord/Scoping/Default/ClassMethods.html#method-i-default_scope) to set the order for `members` by `created_at` ? 

In general, I am not a fan of default scopes, but it seems cleaner than adding this `ordered` in so many views and potentially forgetting to add it some places.

disambiguator:
@jessieay 

Yeah, I agree about not liking default scopes, although I think this is a good use case for it. I would prefer that to calling ordered everywhere. We'll definitely forget in the future. I also thought it would always order by id otherwise, which would be consistent. Curious what is causing it to come out random.

jessieay:
@disambiguator it's not *exactly* random, it's based on when they were each last updated. Which, to client, will definitely be experienced as random.

I will updated this PR to use a default scope. :)

jessieay:
@disambiguator updated at 7bbaaaad

(BTW -- how cool is that commit SHA?)

Note that I put the order on the relation on the `MedicaidApplication` instead, since I realized after starting this that it was the *association* order that was based on `updated_at`. `Member.all` was always in ascending created at order, but `medicaid_application.members` was ordering based on `updated_at.

jayroh:
7bbaaaa ?

Well ok then.

This gets a 



7 happy sheep + 3 for good measure

disambiguator:
@jessieay Cool tyty. lgtm!

PR #513, opened 2017-11-17
Updates w/o active application will be redirected

jayroh:
[Finishes #152965607]

If someone trys to update an application that is no longer active --
example: they finish the application, hit the back button a few times,
and submit an update to a page -- they will be redirected back to the
homepage.


spec/controllers/steps_controller_spec.rb
```diff
@@ -2,11 +2,17 @@
...
       get :index
 
       expect(response).to redirect_to(root_path)
     end
+
+    it "redirects to the homepage on a PUT" do
+      put :update, params: { id: "" }
+
+      expect(response).to redirect_to(root_path)
+    end
```
jayroh: One question on this - should we include something in a flash message that tells then _why_ they were redirected back to the homepage? Right now we just send them back without any notice.

config/routes.rb
```diff
@@ -44,7 +44,7 @@
 
   resource :skip_send_application, only: [:create]
 
-  resources :steps, only: %i[index show] do
+  resources :steps, only: %i[index show update] do
```
jessieay: do we call `update ` on `steps` ? I thought we only ever used the `update` action for the nested resources inside

spec/controllers/steps_controller_spec.rb
```diff
@@ -2,11 +2,17 @@
...
   context "when no applications have been started" do
-    it "redirects to the homepage" do
+    it "redirects to the homepage on a GET" do
       get :index
 
       expect(response).to redirect_to(root_path)
     end
+
+    it "redirects to the homepage on a PUT" do
+      put :update, params: { id: "" }
```
jessieay: when would this request to the `StepsController` happen IRL?

spec/controllers/steps_controller_spec.rb
jessieay: flash message sounds nice, although the flash messages styles on our homepage are kinda wonky at the moment

config/routes.rb
jayroh: We call update on the classes that inherit from this. Never directly on this controller.

spec/controllers/steps_controller_spec.rb
jayroh: Never would.

disambiguator:
Feels like all this additional test coverage doesn't get us that much, since it would be really hard to break one without breaking them all. I'd be okay with not adding that shared example everywhere - curious about others' thoughts.

jessieay:
@disambiguator I agree it feels like a lot to have it everywhere, but if we are not including it everywhere how do we decide where to include it? Originally Joel had it tested on `StepsController`, but that required defining an `update` route for `StepsController` which was unused except for in the test.

jessieay:
@disambiguator you want to take this over to get it closed out today, or should I?

disambiguator:
@jessieay yes i can!

spec/controllers/steps_controller_inheritance_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  describe "#edit" do
+    before do
+      routes.draw { get "edit" => "steps#edit" }
+    end
+
+    context "when there is no application under way" do
+      it "redirects to the specified path" do
+        session[:snap_application_id] = nil
```
jessieay: should we have tests for `medicaid_application_id` as well?

disambiguator: @jessieay good calllllll

PR #514, opened 2017-11-17
Validate that at least one other income type given

jayroh:
[Finishes #152702632]

The `Other Income Type` page was not displaying the validation error. This
commit fixes that.

### DIFF
--- a/spec/views/medicaid/income_other_income_type/edit.html.erb_spec.rb
+++ b/spec/views/medicaid/income_other_income_type/edit.html.erb_spec.rb
@@ -1,0 +1,39 @@
+require "rails_helper"
+
+describe "medicaid/income_other_income_type/edit.html.erb" do
+  before do
+    controller.singleton_class.class_eval do
+      def current_path
+        "/steps/medicaid/income-other-income-type"
+      end
+
+      def current_member
+        Member.last
+      end
+
+      def current_application
+        current_member.benefit_application
+      end
+
+      helper_method :current_path, :current_application, :current_member
+    end
+  end
+
+  context "has not selected another income type" do
+    it "shows an error message" do
+      member = create(:member, other_income: true)
+      step = Medicaid::IncomeOtherIncomeType.new(
+        other_income_types: [],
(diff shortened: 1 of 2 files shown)


spec/views/medicaid/income_other_income_type/edit.html.erb_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "rails_helper"
+
+describe "medicaid/income_other_income_type/edit.html.erb" do
+  before do
+    controller.singleton_class.class_eval do
+      def current_path
+        "/steps/medicaid/income-other-income-type"
+      end
+
+      def current_member
```
jessieay: defining `current_member` up here but `member` in the test itself feels kind of confusing to me.

spec/views/medicaid/income_other_income_type/edit.html.erb_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        Member.last
+      end
+
+      def current_application
+        current_member.benefit_application
+      end
+
+      helper_method :current_path, :current_application, :current_member
```
jessieay: why not make these private methods instead?

spec/views/medicaid/income_other_income_type/edit.html.erb_spec.rb
jayroh: `current_member` is up here so that the view has access to it. It's a gateway to what we create in the setup phase.

It's weird, I agree, but it's a necessity to get this to function.  Now, if we had *one* object passed through from the controller to the view then this wouldn't be necessary. Like `@step.current_member`. It's an interesting smell that we provide access to this many objects in the view. View test is a nice way to suss those out.

spec/views/medicaid/income_other_income_type/edit.html.erb_spec.rb
jayroh: I'm not sure I understand. As opposed to / instead of ...?

PR #515, opened 2017-11-17
Fix issue with `under 21` checkboxes for 1426

jessieay:
* "No" checkbox for under 21 was not working properly
* This was because the export value for this checkbox was defaulting to
"No" but we were setting it to "Yes" when we wanted to check it.
* This PR updates the PDF so that ALL checkboxes have a value of "Yes"
when they are checked.

* To test:
- visit https://staging.michiganbenefits.org/admin/medicaid_applications/144/pdf
- the "no" checkbox should be checked for both household members under
question 9 ("Under 21?"), because they are both over 21.


jessieay:
@disambiguator once this is on staging, I will follow the steps listed in the PR commit msg to confirm that all is working as expected!

jessieay:
working!

screen shot 2017-11-17 at 12 44 11 pm

PR #516, opened 2017-11-17
More explicitly setting members in controllers

disambiguator:
* Controllers that directly inherit from StepsController now have to
 explicitly call super and set it, instead of having it done for you
 based on if your attributes include members
 * Remove the edit hook in favor of calling `super`
 * ManyMembersStepController's children override edit anyways so they
 did not need to be changed as part of this commit.

 Feedback please!

[#152975776]

### DIFF
--- a/app/controllers/medicaid/contact_social_security_controller.rb
+++ b/app/controllers/medicaid/contact_social_security_controller.rb
@@ -1,5 +1,11 @@
 module Medicaid
   class ContactSocialSecurityController < Medicaid::ManyMemberStepsController
+    def step
+      @step ||= step_class.new(
+        members: members_requesting_health_insurance,
+      )
+    end
+
     private
 
     def skip?
@@ -12,6 +18,12 @@ module Medicaid
 
     def member_attrs
       %i[ssn birthday]
+    end
+
+    def members_requesting_health_insurance
+      current_application.
+        members.
+        where(requesting_health_insurance: true)
     end
   end
 end
--- a/app/controllers/medicaid/health_pregnancy_controller.rb
+++ b/app/controllers/medicaid/health_pregnancy_controller.rb
@@ -1,5 +1,11 @@
 module Medicaid
   class HealthPregnancyController < MedicaidStepsController
+    def edit
+      super
+
+      @step.members = current_application.members
+    end
+
     private
 
     def update_application
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -20,12 +20,6 @@ class StepsController < ApplicationController
   def edit
     attribute_keys = Step::Attributes.new(step_attrs).to_sym
     @step = step_class.new(existing_attributes.slice(*attribute_keys))
-
-    if attribute_keys.include?(:members)
-      @step.members = current_application&.members
-    end
-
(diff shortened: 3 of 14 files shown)


app/controllers/medicaid/contact_social_security_controller.rb
```diff
@@ -1,5 +1,11 @@
 module Medicaid
   class ContactSocialSecurityController < Medicaid::ManyMemberStepsController
+    def step
+      @step ||= step_class.new(
+        members: members_requesting_health_insurance,
```
disambiguator: Now we can only add members to the step that actually need updating! Prevents no-ops for irrelevant members in the update step.

app/controllers/medicaid/health_pregnancy_controller.rb
```diff
@@ -1,12 +1,20 @@
 module Medicaid
   class HealthPregnancyController < MedicaidStepsController
+    def edit
+      super
+
+      @step.members = current_application.members
```
jessieay: why add to `update` rather than define `step` here like you did in `ContactSocialSecurityController`? Feels like they are similar scenarios?

app/controllers/medicaid/health_pregnancy_controller.rb
```diff
@@ -15,13 +23,7 @@ def skip?
...
 
     def all_males?
-      current_application.members.all? do |member|
-        member.sex == "male"
-      end
-    end
-
-    def member_attrs
-      { new_mom: step_params[:anyone_new_mom] }
+      current_application.members.all?(&:male?)
```
jessieay: 🆗

app/controllers/steps_controller.rb
```diff
@@ -20,12 +20,6 @@ def current_application
   def edit
     attribute_keys = Step::Attributes.new(step_attrs).to_sym
     @step = step_class.new(existing_attributes.slice(*attribute_keys))
-
-    if attribute_keys.include?(:members)
```
jessieay: ✅

spec/models/medicaid_application_spec.rb
```diff
@@ -90,4 +90,15 @@
       expect(app.no_expenses?).to eq(true)
     end
   end
+
+  describe "#members_count" do
+    it "returns the count of associated members" do
+      app = create(
+        :medicaid_application,
+        members: build_list(:member, 5),
```
jessieay: woooo `build` !

app/steps/medicaid/expenses_alimony.rb
```diff
@@ -2,7 +2,7 @@ module Medicaid
   class ExpensesAlimony < Step
     step_attributes(
       :anyone_pay_child_support_alimony_arrears,
-      :members,
+      :members_count,
```
jessieay: nice!

app/controllers/medicaid/health_pregnancy_controller.rb
disambiguator: The flow is _ever so slightly_ different for steps that inherit from the ManyMembersStepsController and the MedicaidStepsController. Because of that, the behavior is a lil bit different.

But you know me, I'd love to tackle making that the same as well. :)

PR #520, opened 2017-11-20
Append pages to 1426 PDF for 3+ member households

jessieay:
* The 1426 PDF advises to copy pages 5 and 6 when there are  more than 2
members in the household.
* We are now doing this programmatically.
* [Finishes #152727264]


jessieay:
@bengolder want to do the merging honors?

bengolder:
@jessieay thank you! done.

PR #521, opened 2017-11-20
Fix bug with steps using `members_count` attribute

jessieay:
* [Fixes #153018504]
* The i18n copy was not showing up properly on these pages because
`members_count` was an attribute on steps but by default steps are
initialized with attributes of the `current_application` and
`members_count` was a method rather than an attribute.
* To fix this, I made `members_count` an attribute on the
`medicaid_applications` table. It is a counter cache column so we no
longer need to do a SQL join query each time we ask for `members.count` in
the view, which does a join on `medicaid_application.members` and
returns the count. Now, it just gets the data from a column in the
database.
* Note that we will need to run rails console and `reset_counters` for
all Medicaid and Snap apps once this is deployed. https://apidock.com/rails/ActiveRecord/CounterCache/reset_counters

### DIFF
--- a/app/views/medicaid/amounts_expenses/edit.html.erb
+++ b/app/views/medicaid/amounts_expenses/edit.html.erb
@@ -4,7 +4,7 @@
   <header class="form-card__header">
     <div class="form-card__title">
       <%= t("medicaid.amounts_expenses.edit.title",
-            count: @step.members_count,
+            count: current_application.members_count,
             name: current_member.display_name) %>
     </div>
   </header>
--- a/app/views/medicaid/amounts_expenses/edit.html.erb
+++ b/app/views/medicaid/amounts_expenses/edit.html.erb
@@ -4,7 +4,7 @@
   <header class="form-card__header">
     <div class="form-card__title">
       <%= t("medicaid.amounts_expenses.edit.title",
-            count: current_application.members.count,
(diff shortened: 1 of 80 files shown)


app/views/medicaid/amounts_expenses/edit.html.erb
```diff
@@ -4,7 +4,7 @@
   <header class="form-card__header">
     <div class="form-card__title">
       <%= t("medicaid.amounts_expenses.edit.title",
-            count: current_application.members.count,
+            count: @step.members_count,
```
jessieay: note to self: move this to be `current_application.members_count`

spec/models/dhs1426_pdf_spec.rb
```diff
@@ -40,12 +39,13 @@
...
         student_loan_interest_expenses: 70,
       )
-      secondary_member = create(
+      secondary_member = build(
         :member,
         first_name: "Roger",
         last_name: "Rabbit",
         birthday: Date.new(1980, 1, 20),
         sex: "male",
+        spouse: primary_member,
```
disambiguator: oh right duh :) @jessieay

jessieay: yeah took me a min :)

PR #522, opened 2017-11-20
Prepend cover sheet, append paperwork for 1426

jessieay:
* Following same pattern as 1171
* [Finishes #152564700]

Tested locally:

screen shot 2017-11-20 at 3 30 07 pm


app/models/verification_document.rb
```diff
@@ -1,7 +1,7 @@
 class VerificationDocument
-  def initialize(url:, snap_application:)
+  def initialize(url:, benefit_application:)
     @url = url
-    @snap_application = snap_application
+    @benefit_application = benefit_application
```
disambiguator: yay shared code

app/models/dhs1426_pdf.rb
```diff
@@ -108,4 +111,17 @@ def first_two_members
       { member: application_members[1], position: "second" },
     ]
   end
+
+  def verification_paperwork_paths
+    verification_paperwork.map do |document|
+      VerificationDocument.new(
+        url: document,
+        benefit_application: medicaid_application,
+      ).file&.path
```
disambiguator: When is `file` nil? If it actually is possible, should there be a test for that scenario?

PR #523, opened 2017-11-21
WIP add questions as `legend ` elements in fieldsets

jessieay:
* Showing how I will tackle the rest of
https://www.pivotaltracker.com/story/show/152761627 before going through
each page, since it will require a lot of manual work.

* Any objections to this approach?

### DIFF
--- a/app/assets/stylesheets/michigan-benefits/atoms/_typography.scss
+++ b/app/assets/stylesheets/michigan-benefits/atoms/_typography.scss
@@ -1,6 +1,7 @@
 .text--section-header {
   color: $color-sky;
   font-weight: 600;
+  margin-bottom: 1em;
 }
 
 .text--little-help {
--- a/app/views/contact_confirm_phone_number/edit.html.erb
+++ b/app/views/contact_confirm_phone_number/edit.html.erb
@@ -20,6 +20,7 @@
         autofocus: true,
         options: { maxlength: 10 } %>
 
-    <%= render "shared/next_step" %>
-  <% end %>
+      <%= render "shared/next_step" %>
+    <% end %>
+  </div>
 </div>
--- a/app/views/income_details_per_member/_household_member_group.html.erb
+++ b/app/views/income_details_per_member/_household_member_group.html.erb
@@ -1,0 +1,6 @@
+<fieldset class="form-group household-member-group" data-member-name="<%= member.display_name %>">
+  <legend class="text--section-header">
+    <%= member.display_name %> (<%= member.employment_status.titleize %>)
+  </legend>
+  <%= yield %>
+</fieldset>
--- a/spec/features/medicaid_application_with_multiple_members_spec.rb
+++ b/spec/features/medicaid_application_with_multiple_members_spec.rb
@@ -155,7 +155,7 @@ RSpec.feature "Medicaid app" do
 
     on_page "Health Coverage Needs" do
       expect(page).to have_content(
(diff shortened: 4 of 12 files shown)


spec/features/medicaid_application_with_multiple_members_spec.rb
```diff
@@ -155,7 +155,7 @@
 
     on_page "Health Coverage Needs" do
       expect(page).to have_content(
-        "Tell us who is currently enrolled in a health insurance plan.",
+        I18n.t("medicaid.insurance_current_member.edit.title"),
```
disambiguator: Using I18n.t in tests is more typing than you really need. Here’s how you can use t() instead of I18n.t() in all of your tests:
```
RSpec.configure do |config|
  config.include AbstractController::Translation
end
```
[From this article](https://robots.thoughtbot.com/foolproof-i18n-setup-in-rails)

disambiguator:
@jessieay Love it.

hartsick:
dig it. thank you so much for doing this (esp pulling things into translations file!). know this is teeeeedious

spec/features/medicaid_application_with_multiple_members_spec.rb
bengolder: Do feature specs need to reference the translations file? It seems like it might be useful to make assertions about the final string output here.

jessieay: @bengolder it's nice to have them reference the translation because that means copy changes will not affect specs. I see what you're saying with it being nice to see the strings in test, but as long as we stick to a consistent naming scheme, finding the correct translations should be pretty easy.

app/views/contact_confirm_phone_number/edit.html.erb
```diff
@@ -20,6 +20,7 @@
         autofocus: true,
         options: { maxlength: 10 } %>
 
-    <%= render "shared/next_step" %>
-  <% end %>
+      <%= render "shared/next_step" %>
+    <% end %>
+  </div>
```
bengolder: this was just a missing closing tag I encountered.

app/views/income_details_per_member/_household_member_group.html.erb
```diff
@@ -0,0 +1,6 @@
+<fieldset class="form-group household-member-group" data-member-name="<%= member.display_name %>">
+  <legend class="text--section-header">
+    <%= member.display_name %> (<%= member.employment_status.titleize %>)
+  </legend>
+  <%= yield %>
+</fieldset>
```
bengolder: This is a layout that can be used to wrap a set of household member fields in a fieldset

bengolder:
### Wave analysis of a page with missing legend: 
screen shot 2017-11-22 at 11 26 04 am

### Wave analysis of modified income-per-member page with fieldset and legend layout
screen shot 2017-11-22 at 11 25 40 am

app/assets/stylesheets/michigan-benefits/atoms/_typography.scss
```diff
@@ -1,6 +1,7 @@
 .text--section-header {
   color: $color-sky;
   font-weight: 600;
+  margin-bottom: 1em;
```
bengolder: this is necessary as we replace `<p>` tags with `<legend>` tags. `<p>` tags have a `1em` margin bottom, which should presumably be the same for any section header.

bengolder:
merging after in-person review from Paras

PR #524, opened 2017-11-21
Fill out 1426 Step 4

bengolder:
- Adds an insurance type spec shared example
- Refactoring for some tests and attributes classes
- Includes an edit to the 1426 PDF

[Finishes #152727191]

Signed-off-by: Christa Hartsock <christa@codeforamerica.org>


hartsick:
One sec — got this into a bad spot with outdated master. Correcting now!

hartsick:
Merging since we paired!

PR #526, opened 2017-11-21
Bugfix: PDF with documents

jessieay:
* We were getting an error on staging related to a `nil` birthday.
Medicaid applicants will not always have a birthday, so we have to check
for birthday before calling `strftime` on it.
* [Finishes #152564700] (again)
* Also removes unused `birthday` field on `medicaid_applications` table

### DIFF
--- a/app/models/verification_document.rb
+++ b/app/models/verification_document.rb
@@ -31,7 +31,15 @@ class VerificationDocument
 
   def header
     output = "Name:                  #{benefit_application.display_name}\n"
-    output += "Date of Birth:        #{benefit_application.birthday}\n"
+    output += "Date of Birth:        #{birthday}\n"
     output
+  end
+
+  def birthday
+    if benefit_application.primary_member.birthday.present?
+      benefit_application.primary_member.formatted_birthday
+    else
+      ""
+    end
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171120220922) do
+ActiveRecord::Schema.define(version: 20171121231237) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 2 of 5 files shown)


jessieay:
this is causing all 1426's with docs to break right now so let me know when you can review @luigi @hartsick @disambiguator @bengolder

app/models/verification_document.rb
```diff
@@ -31,7 +31,15 @@ def add_header(pdf)
...
   def header
     output = "Name:                  #{benefit_application.display_name}\n"
-    output += "Date of Birth:        #{benefit_application.birthday}\n"
+    output += "Date of Birth:        #{birthday}\n"
     output
   end
+
+  def birthday
+    if benefit_application.primary_member.birthday.present?
+      benefit_application.primary_member.formatted_birthday
```
hartsick: I'm uncomfortable with the formatted_birthday as an attribute on model, but don't think this should block this PR! would like to talk about how our team wants to handle model vs view concerns going forward

db/schema.rb
```diff
@@ -125,7 +125,6 @@
     t.boolean "anyone_pay_child_support_alimony_arrears"
     t.boolean "anyone_pay_student_loan_interest"
     t.boolean "anyone_self_employed", default: false
-    t.datetime "birthday"
```
bengolder: does this mean we aren't storing date of birth anymore?

jessieay: we are storing in on each member. we used to store it here but no longer do so!

app/models/verification_document.rb
```diff
@@ -31,7 +31,15 @@ def add_header(pdf)
...
+    else
+      ""
+    end
```
luigi: I see `formatted_birthday` is a method on the `member` model. It would be cool if that were modified to account for nil values of `birthday` so that this above could be a one-liner:

`benefit_application.primary_member.formatted_birthday.presence`

app/models/verification_document.rb
jessieay: yeah agreed! we have to deal with a `nil` case on birthday somewhere else in our PDF parsing code as well and it's icky

db/schema.rb
bengolder: answered my own question. Looks like it's stored on member.

app/models/verification_document.rb
luigi: I think it's a nice-to-have, not essential.

jessieay: yeah, we could define `formatted_birthday` to allow for nil values. Per @hartsick's comment, maybe we don't want that method in the model at all since it's more of a view concern

PR #527, opened 2017-11-22
Refactor: move `formatted_birthday` to decorator

jessieay:
* Perhaps we should store view-specific methods like this in decorators?

cc @hartsick  @luigi

### DIFF
--- a/app/decorators/member_decorator.rb
+++ b/app/decorators/member_decorator.rb
@@ -1,0 +1,19 @@
+class MemberDecorator
+  def initialize(member)
+    @member = member
+  end
+
+  delegate :birthday, to: :member
+
+  def formatted_birthday
+    if birthday.present?
+      birthday.strftime("%m/%d/%Y")
+    else
+      ""
+    end
+  end
+
+  private
+
+  attr_reader :member
+end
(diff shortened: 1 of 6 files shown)


app/decorators/member_decorator.rb
```diff
@@ -0,0 +1,19 @@
+class MemberDecorator
+  def initialize(member)
+    @member = member
+  end
+
+  delefate :birthday, to: member
```
luigi: Typo. :)

app/decorators/member_decorator.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def formatted_birthday
+    if birthday.present?
+      birthday.strftime("%m/%d/%Y")
+    else
+      ""
+    end
```
luigi: I like `birthday.present? : birthday.strftime("%m/%d/%Y") : ""` but I don't know how the team/rubocop feel about it.

db/migrate/20171121231237_remove_birthday_from_medicaid_applications.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveBirthdayFromMedicaidApplications < ActiveRecord::Migration[5.1]
```
jessieay: ack! somehow forgot to add this file to my last PR -- sorry! moving too fast over here

app/decorators/member_decorator.rb
jessieay: I am fine with that, not my usual style but nothing wrong with it and rubocop does not complain. anyone else have an opinion?

PR #529, opened 2017-11-22
Remove validations on phone number for SNAP/Medicaid

jessieay:
* Makes phone number a non-required field.
* [Finishes #152981586]
* Also added 'optional' text to inputs to be consistent with other optional inputs

screen shot 2017-11-22 at 10 00 55 am

screen shot 2017-11-22 at 9 59 24 am

### DIFF
--- a/spec/steps/medicaid/contact_phone_spec.rb
+++ b/spec/steps/medicaid/contact_phone_spec.rb
@@ -2,12 +2,6 @@ require "rails_helper"
 
 RSpec.describe Medicaid::ContactPhone do
   describe "Validations" do
-    it "does not validates presence" do
-      step = Medicaid::ContactPhone.new(phone_number: nil)
-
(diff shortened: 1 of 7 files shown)


spec/steps/medicaid/contact_phone_spec.rb
```diff
@@ -2,13 +2,10 @@
 
 RSpec.describe Medicaid::ContactPhone do
   describe "Validations" do
-    it "validates presence" do
+    it "does not validates presence" do
```
hartsick: could we remove this test entirely? think this is useful to drive out the change, but like only keeping around positive assertions for posterity!

jessieay: yeah good call,I was 50/50 on this. removed at 8bac807a

PR #531, opened 2017-11-22
Add street address 2 to SNAP and Medicaid applications

luigi:
Adds Street Address 2 field to `address` model, steps, and PDFs.

Notable refactors:
* Param handling in `medicaid/contact_home_address_controller.rb`
* Pulls out `residential_or_homeless` method to each attributes class since the residential street address is handled differently for each.

Bonus:
* Drops in a migration which was accidentally left off a previous merge: https://github.com/codeforamerica/michigan-benefits/commit/a95d171fba1c88902a424f4cb8e12202d88a8814

Tracker: https://www.pivotaltracker.com/n/projects/2123705/stories/152945205
[Finishes #152945205]


app/controllers/medicaid/contact_home_address_controller.rb
```diff
@@ -3,13 +3,8 @@ class ContactHomeAddressController < MedicaidStepsController
...
 
     def existing_attributes
-      attributes = super
-      residential_address = current_application.residential_address
-      attributes.merge!(
-        street_address: residential_address.street_address,
-        city: residential_address.city,
-        zip: residential_address.zip,
-      )
+      addr_attributes = address.attributes.merge(current_application.attributes)
```
jessieay: how about naming this `address_attributes` to be clearer?

app/steps/medicaid/contact_home_address.rb
```diff
@@ -2,7 +2,10 @@ module Medicaid
   class ContactHomeAddress < Step
     step_attributes(
       :street_address,
+      :street_address_2,
       :city,
+      :county,
```
jessieay: now that county and state are attributes on the step, what about adding validations for them at the step level so we can be sure they aren't left blank?

app/controllers/medicaid/contact_home_address_controller.rb
```diff
@@ -27,27 +22,23 @@ def unstable_housing?
...
 
     def update_application
-      current_application.update!(application_params)
-      address.update!(address_params)
-    end
-
-    def application_params
-      step_params.except(:street_address, :city, :zip)
+      current_application.update!(
+        same_address_key => step_params[same_address_key],
```
jessieay: nice!

app/steps/medicaid/contact_home_address.rb
luigi: Thanks for pointing this out. @hartsick and I will address this in the work on non-Genesee counties:

https://www.pivotaltracker.com/story/show/152981944

app/controllers/medicaid/contact_home_address_controller.rb
jessieay: you can always break it up!

```
address_attributes = address.
  attributes.
  merge(current_application.attributes)
```

PR #532, opened 2017-11-22
Full width buttons

hartsick:
### Full-width buttons on mobile, and center text
Repurposes the existing '.button--next' class as '.button--full-mobile', and removes the now unused .button--next class.

Previously we were not using the class "button--full-width", but were using "button--next" in a similar fashion. This makes that a bit more clear.

screen shot 2017-11-22 at 11 43 14 am

screen shot 2017-11-22 at 11 44 00 am

### Reorder paperwork guide buttons
Our pattern is to always put the call-to-action on righthand side / bottom. This reorders the paperwork-related buttons accordingly.

screen shot 2017-11-22 at 11 42 11 am

Also deletes unused templates.


app/views/pages/medicaid/amounts_income_all_members.erb
```diff
@@ -1,52 +0,0 @@
...
-
-  <!-- Replace with partial -->
-  <footer class="form-card__button_right">
-    <a href="/pages/medicaid/amounts_loan_interest" class="button button--nav  button--cta button--full-width">
-      Next
-    </a>
-  </footer>
-</div>
-
-<!-- https://trello.com/c/Q79fWjze/74-tell-us-the-specific-income-received-from-members-current-job-employment-other-income-all-asked-on-same-page -->
```
bengolder: was this an unused view?

bengolder: @hartsick said yes. it was unused.

PR #538, opened 2017-11-23
update copy

disambiguator:
[#152981692]

### DIFF
--- a/app/views/medicaid/intro_household/edit.html.erb
+++ b/app/views/medicaid/intro_household/edit.html.erb
@@ -7,8 +7,9 @@
       in your household.
     </div>
     <p class="text--help text--centered">
-      This includes everyone who lives in your home (including you), even if
-      they are not seeking benefits today. Include room-mates that are not family.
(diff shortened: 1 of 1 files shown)


app/views/medicaid/intro_household/edit.html.erb
```diff
@@ -7,8 +7,9 @@
       in your household.
     </div>
     <p class="text--help text--centered">
-      This includes everyone who lives in your home (including you), even if
-      they are not seeking benefits today. Include room-mates that are not family.
+      This includes everyone who lives in your home (including you), and anyone
```
bengolder: I know this is written as intended, and still not sure where to put the concern, but in the case of someone who lives with their landlord, they might be inclined to include their landlord, even if the landlord is not a part of their tax household.

bengolder: I'll put a comment on the design Trello card

disambiguator:
@jessieay Truth. Done deal for next time ty :) :) :)

PR #539, opened 2017-11-23
Edit copy

disambiguator:
* Extract description to i18n

[#152981733]

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -118,3 +118,5 @@ en:
           typically file taxes."
           other: "In the next section, we will ask you for information how you
           and your household typically file taxes."
+  ssn:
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -118,3 +118,5 @@ en:
           typically file taxes."
           other: "In the next section, we will ask you for information how you
           and your household typically file taxes."
+  ssn:
+    description: "Providing Social Security numbers ensures you receive the correct benefits. Even if you've applied before, this helps connect this application to the right file. MDHHS maintains strict security guidelines to protect the identities of our residents."
```
jessieay: may want to consider using a key other than `description` since this is more of a security message / warning that a description

disambiguator: word - no attachment to it if it changes. i found myself thinking too hard about a name and then gave up

PR #541, opened 2017-11-27
Checkbox for medicare premium help

disambiguator:
* if person has Medicare, then we should automatically check "Yes" on the
   1426 field, "Do you need help paying Medicare premiums?"

[Finishes #152727191]

https://www.pivotaltracker.com/story/show/152727191

### DIFF
--- a/spec/models/medicaid_application_attributes_spec.rb
+++ b/spec/models/medicaid_application_attributes_spec.rb
@@ -21,7 +21,29 @@ RSpec.describe MedicaidApplicationAttributes do
 
     describe "insurance types" do
       it_should_behave_like "insurance type", "medicaid", "Medicaid"
-      it_should_behave_like "insurance type", "medicare", "Medicare"
+
+      context "applicant is on medicare" do
+        it_should_behave_like "insurance type", "medicare", "Medicare"
+        it "should fill out yes for do you need help" do
+          member_with_medicare = build(
+            :member,
+            insurance_type: "Medicare",
+          )
+          medicaid_application = create(
+            :medicaid_application,
+            members: [member_with_medicare],
+          )
+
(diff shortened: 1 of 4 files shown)


spec/models/medicaid_application_attributes_spec.rb
```diff
@@ -21,7 +21,29 @@
 
     describe "insurance types" do
       it_should_behave_like "insurance type", "medicaid", "Medicaid"
-      it_should_behave_like "insurance type", "medicare", "Medicare"
+
+      context "applicant is on medicare" do
```
jessieay: shoud we add "and needs help paying medical bills from the past 3 months" to this  context?

spec/support/shared_examples/insurance_type.rb
```diff
@@ -3,20 +3,20 @@
 RSpec.shared_examples_for "insurance type" do |insurance_slug, insurance_name|
   context "one household member with #{insurance_name}" do
     it "sets insured_#{insurance_slug}_yes to 'Yes' and lists their name" do
-      member_with_medicaid = build(
+      member_with_insurance = build(
```
jessieay: 💯

spec/models/medicaid_application_attributes_spec.rb
disambiguator: That feels unrelated. This is the box we're checking.

pasted image at 2017_11_21_06_03 pm

jessieay: gotcha, I interpreted this feature to be related to that question but it sounds like it isn't

disambiguator: Yee

PR #542, opened 2017-11-27
Remove presence validations on `amounts-income` page

jessieay:
* These fields are all optional.
* [Finishes #153069326]
https://www.pivotaltracker.com/story/show/153069326


disambiguator:
woke

PR #543, opened 2017-11-28
1426: Add tax info for primary member

jessieay:
* We need some clarification before adding this information for
additional household members.
* [Part of #153062895]

https://www.pivotaltracker.com/story/show/153062895

Screenshot of this section of the 1426 from a test Medicaid app locally:

screen shot 2017-11-27 at 4 12 08 pm


app/models/medicaid_application_attributes.rb
```diff
@@ -102,4 +111,12 @@ def residential_or_homeless
       "Homeless"
     end
   end
+
+  def dependents
```
hartsick: Would it make sense to have tests around this logic (filtering dependents & the first name display) at this object level?

app/models/medicaid_application_member_attributes.rb
```diff
@@ -40,6 +40,10 @@ def to_h
         bool_to_checkbox(member.pay_student_loan_interest?),
       :"#{position}_member_deduction_alimony_yes" =>
         bool_to_checkbox(member.pay_child_support_alimony_arrears?),
+      joint_filer_key => "Yes",
```
hartsick: could we add object-level tests for logic about other joint filing member?

app/models/medicaid_application_attributes.rb
jessieay: done at 42daaa993bb04faaa6c973436262c8bee6aa135d

app/models/medicaid_application_member_attributes.rb
jessieay: done at 42daaa993bb04faaa6c973436262c8bee6aa135d

jessieay:
Paired with Christa on this so mergin'

PR #546, opened 2017-11-28
Email 1426 to MDHHS office when submitted

jessieay:
* This email is triggered when someone submits their medicaid app (is on the `/steps/medicaid/success` page)
* We already had this behavior for SNAP, so used the existing `Export`
model to track emails to offices (so we do not re-email if we do not intend to)
* `Export` now has a polymorphic relationship to SNAP and Medicaid apps.
* Main difference between the 1171 and 1426 emails to offices is that we
have a different template for the 1171 if the applicant is currently in
the office lobby (aka - applying from an office-specific landing page).
For medicaid, you cannot currently apply from an office-specific landing
page, so we are just using one email template. The email template
assumes that the person is applying *not* from the office.
* This is potentially blocked by
https://www.pivotaltracker.com/story/show/153062895, which has some open
questions.
* [Finishes #152564773]
* https://www.pivotaltracker.com/story/show/152564773

screenshot of email from local test:

screen shot 2017-11-28 at 11 43 55 am


spec/controllers/medicaid/success_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    it "emails the medicaid application to the office" do
+      allow(ExportFactory).to receive(:create)
+
+      get :edit
+
+      expect(ExportFactory).to have_received(:create).
+        with(benefit_application: current_app, destination: :office_email)
+    end
+
+    context "application is not exportable" do
```
disambiguator: Why would the application not be exportable?

app/jobs/office_email_application_job.rb
```diff
@@ -1,14 +1,24 @@
 class OfficeEmailApplicationJob < ApplicationJob
   def perform(export:)
-    export.execute do |snap_application, logger|
-      ApplicationMailer.office_application_notification(
-        file_name: snap_application.pdf.path,
-        recipient_email: snap_application.receiving_office_email,
-        office_location: snap_application.office_location,
-      ).deliver
+    export.execute do |benefit_application, logger|
+      if benefit_application.class == SnapApplication
```
disambiguator: Feels a bit odd that perform is one big conditional. My hunch is that this should be two different job classes.

spec/controllers/medicaid/success_controller_spec.rb
jessieay: if it is not signed it is not exportable.

app/jobs/office_email_application_job.rb
jessieay: ok we can do that! I was hoping that the job could run the same email with different view templates but this somehow felt clearer to me. what do you think?

disambiguator: Would a shared partial solve that?

spec/controllers/medicaid/success_controller_spec.rb
disambiguator: How could you get to this step if that was the case? Especially if this is called the "success" controller.

app/jobs/office_email_application_job.rb
jessieay: yeah we already have different partials used for the mailer depending on whether the SNAP app has on `office_location` so we could have another branch of view partial logic for Medicaid apps....let me put something together once Christa and I are done with our current work and circle back with a solution that uses a different partial in the mailer rather than logic here

spec/controllers/medicaid/success_controller_spec.rb
jessieay: technically you shouldnt be able to...I am not sure what the historical reason for having this as a method was, but it is the pattern we had in SNAP so I figured we should replicate here. Perhaps we should page product/design on this?

jessieay: we removed this functionality!

disambiguator: YAY LESS LOGIC :)

app/jobs/office_email_application_job.rb
jessieay: moved to its own class /job rather than a conditional at cc1a53e9

app/models/concerns/submittable.rb
```diff
@@ -16,4 +20,11 @@ def receiving_office_phone_number
   def receiving_office
     @receiving_office ||= OfficeRecipient.new(benefit_application: self)
   end
+
+  def close_pdf
+    if @pdf.present? && @pdf.respond_to?(:close)
```
disambiguator: when would a pdf not respond to `close`?
when would a pdf not be present? Why would we even be calling close_pdf if that was true?

app/models/medicaid/export_factory.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    def enqueue(export)
+      export.transaction do
+        export.save!
+        case export.destination
+        when :office_email
+          Medicaid::OfficeEmailApplicationJob.perform_later(export: export)
+        else
+          raise UnknownExportTypeError, export.destination
```
disambiguator: When would this happen?

jessieay: when someone writes invalid code but the error gets lost because we just don't catch it in a case statement

app/models/concerns/submittable.rb
jessieay: if the PDF is not generated properly and we try to close it, this will throw an exception

app/jobs/fax_application_job.rb
```diff
@@ -1,11 +1,11 @@
 class FaxApplicationJob < ApplicationJob
   def perform(export:)
-    export.execute do |snap_application, logger|
-      receiving_office = snap_application.receiving_office
+    export.execute do |benefit_application, logger|
```
jessieay: to be consistent with your other changes, should this be `snap_application` as well @hartsick ?

app/models/medicaid/export_factory.rb
disambiguator: seems like a testing concern rather than a app code concern but nbd. lgtm

disambiguator:
Looks good to me but deferring final approval to @jessieay.

Yay to a few less conditionals! Slowly but sure removing code paths :)

jessieay:
Woah it looks like I cannot officially approve because I am one of the PR authors but unofficially this looks good to me. Do we need to wait before merging @hartsick or are we good to go? (wondering if we need to let MDHHS know that they will start receiving emails for 1426?)

hartsick:
Oh weird! I can approve—I think we're good to go, but I'll give Alan a heads up. We had talked about being fine to merge this in ASAP, since we don't have a public Medicaid application anyway.

PR #549, opened 2017-11-29
Add staging.rb

luigi:
This new file is a verbatim copy of production.rb.

This should make it possible to set `RAILS_ENV` and `RACK_ENV` to `staging` for our staging instance. Currently, the staging instance uses `production` for those environment variables.

This will then set the one-time password issuer to "MichiganBenefits.org - staging" for the staging instance.

https://www.pivotaltracker.com/story/show/152983494

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,37 +1,5 @@
-require Rails.root.join("config/smtp")
+require File.expand_path('../shared_production', __FILE__)
 
 Rails.application.configure do
-  config.force_ssl = true
-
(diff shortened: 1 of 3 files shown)


jessieay:
thoughts on inheriting from the production.rb file so we don't have to have duplicate configuration? https://stackoverflow.com/a/18278650/1019369

luigi:
Thanks for the suggestion @jessieay. Updated the PR with a `shared_production.rb`.

hartsick:
looks goooooood

PR #550, opened 2017-11-29
Remove leftover hardcoded county

hartsick:
Also fixes a bug where people with residential addresses were being marked as Homeless in the SNAP flow.

[Finishes #152981944]


hartsick:
Paired on this, so merging!

PR #552, opened 2017-11-29
Pre-production instance set up

disambiguator:
* CI auto deploys production branch to pre-production
 * Release script promotes pre-production instance instead of staging

[#152985452]

### DIFF
--- a/DEPLOYING.md
+++ b/DEPLOYING.md
@@ -7,6 +7,7 @@
 ### Timing for Deploys
 
 * Staging: Deployed automatically on successful builds from `master`.
+* Pre-production: Deployed automatically on successful builds from `production`.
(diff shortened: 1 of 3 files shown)


jessieay:
nice! could we add some updated docs to https://github.com/codeforamerica/michigan-benefits/blob/master/DEPLOYING.md about how we deploy in this new setup? My main question is about when we merge `master` into `production` and how (with merge commits? if so, do we merge those back into master later?)

disambiguator:
@jessieay I added a little bit of documentation but don't know how clear it is.

To sum it up:
 - the production branch would either be at or in front of master, and you should always be able to  do `git merge master --ff-only` without it failing. You never merge back to master since you'd never have commits that only exist on `production`. It just lags a bit behind.

PR #563, opened 2017-12-04
Fix last name bug

jessieay:
* Previously, a name with a capital letter in the middle would be split
into 2
* Now, we only upcase the first letter of the first and last names and
leave the rest alone.
* [Finishes #153189719]
https://www.pivotaltracker.com/story/show/153189719

Before:

screen shot 2017-12-04 at 9 04 34 am

After:

screen shot 2017-12-04 at 9 04 13 am

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -107,7 +107,7 @@ class Member < ApplicationRecord
   end
 
   def display_name
-    "#{first_name} #{last_name}".titleize
+    "#{first_name.upcase_first} #{last_name.upcase_first}"
   end
 
   def self.filing_taxes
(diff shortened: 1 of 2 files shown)


jessieay:
cc @hartsick and @disambiguator -- y'all did the work on this in a WIP branch but when I checked it out, it seemed done?

disambiguator:
@jessieay I don't know if there are other places this needs to get done as well. We use `titleize` in a bunch of places in the code. @hartsick and I were gonna go through and evaluate them all. Also, I have a vague suspicion that `StepsHelper.header_name` is not used anywhere in our code.

jessieay:
@disambiguator should we hold off on merging this, then?

disambiguator:
i think we should hold off on finishing the story but not on merging

bengolder:
Another example last name: `deYoung`

PR #568, opened 2017-12-08
[PAIR] Extract Medicaid employment array fields into Employment model

disambiguator:
Updated AmountsIncome Form

ALSO yak shave:
200w_d

For Medicaid:
- Added employment model with fields based on the existing employment-related array fields.
- Create or destroy employment records when a new job count is entered for a member
- Added idempotent single-transaction one-off rake task to copy data from array fields into new Employment records
- Some of these commits are red so we'll use a merge commit

[#153069646]


spec/features/medicaid_application_with_maximum_info_spec.rb
```diff
@@ -126,38 +126,28 @@
...
+        fill_in "Employer name", with: "Acme Co."
+        fill_in "Average amount / paycheck", with: 50
+        select "Monthly", from: "How often are you paid that amount?"
+      end
 
       fill_in "step_self_employed_monthly_income", with: 100
       fill_in "step_unemployment_income", with: 100
       click_on "Next"
 
-      click_back
```
disambiguator: We want to note that in this PR we removed this part of the feature test that clicks the back button and ensured that data is pre-filled correctly. Our instinct was that this doesn't make sense in this feature test. Does that sound right? @luigi @jessieay @hartsick

spec/features/medicaid_application_with_maximum_info_spec.rb
```diff
@@ -126,38 +126,28 @@
...
+        fill_in "Employer name", with: "Cylinder"
+        fill_in "Average amount / paycheck", with: 25
+        select "Hourly", from: "How often are you paid that amount?"
+      end
+
+      within(find("fieldset", text: "Job #3")) do
+        fill_in "Employer name", with: "Acme Co."
+        fill_in "Average amount / paycheck", with: 50
+        select "Monthly", from: "How often are you paid that amount?"
+      end
```
disambiguator: We felt that asserting the fieldset structure here obviated the need for the view spec

spec/models/medicaid_application_member_attributes_spec.rb
```diff
@@ -110,11 +121,43 @@
         ).to_h
 
         expect(result.keys).not_to include(
-          %i[
-            primary_member_birthday
-            primary_member_under_21_no
-            primary_member_under_21_yes
-          ],
```
disambiguator: This was asserting that the array of keys was not one item contained in `.keys`. In other words it would never have caught the keys.

PR #572, opened 2017-12-08
[PAIR] Specify members to populate many members steps

luigi:
* Previously, `ManyMembersStepsController` had to check if attributes
 were present, because there were times where only specific members
 populate a form, but when assigning attributes, it uses all members in
 application
 * No steps using `ManyMembersStepsController` have fields in the
 application, just on members, so no need to merge attributes
 * Remove some unused attributes from steps

 [#153526809]

Signed-off-by: Paras Sanghavi <paras@codeforamerica.org>


disambiguator:
@hartsick @jessieay @bengolder 

Check out this PR! It's a small but significant change to the way we're handling many member steps. Less conditionals in shared code!

hartsick:
this is rad—

PR #573, opened 2017-12-08
Add missing MiBridges pages

jessieay:
Add MIBridges page for "More About Child Support..."  …
* We do not have any information for this page, so we just need to click
"Next"
* [Finishes #153389765]
* Confirmed this works by following the directions in the story
https://www.pivotaltracker.com/story/show/153389765
3495d00

Add MiBridges page for "More About Savings..."  …
* This page was missing from our driving code
* We do not have any info to fill out so we just click "Next"
* [Finishes #153381448]
* Confirmed that this works using the Snap app ID listed in the story
https://www.pivotaltracker.com/story/show/153381448

MiBridges page for "More about Disability or Blindness..."  …
* We did not have a page for this before
* All we need to do is click next.
* [Finishes #153374449]
* Confirmed that this works by running Snap app indicated in the story
* https://www.pivotaltracker.com/story/show/153374449

MiBridges "More About Your Pregnancy" page  …
* If someone is pregnant or recently pregnancy, MiBridges asks about
"Your" pregnancy (even if the pregnant person is not the applicant...)
* Through this story I also found that the way we were calling
`check_in_section` was invalid. (needed named args).
* [Finishes #153383655]
* Confirmed that this works with the Snap App id in the story
* https://www.pivotaltracker.com/story/show/153383655


lib/mi_bridges/driver.rb
```diff
@@ -56,15 +56,20 @@ class Driver
       MoneyOtherSourcesPage,
       MoreAboutCashOnHandPage,
       MoreAboutCheckingAccountPage,
+      MoreAboutChildSupportPaymentPage,
       MoreAboutDisabilityBenefitsPage,
+      MoreAboutDisabilityOrBlindnessPage,
       MoreAboutJobIncomePage,
       MoreAboutOtherIncomePage,
       MoreAboutPensionOrRetirementPage,
+      # MoreAboutPregnancyPage,
```
hartsick: Is this intentionally commented out?

jessieay: just removed :) Originally thought I needed 2 new pregnancy pages then realized there was just one (seemed weird that the page would be title "More About Your Pregnancy" for a household where the non-primary member is pregnant but....that was the case)

PR #576, opened 2017-12-09
Strip first name and last name leading/trailing whitespace

jessieay:
* This is important for driving code because in MiBridges we depend on
first name to know which household member we are answering questions
about
* Could have used a gem instead, but this seemed like the simplest solution. See: https://stackoverflow.com/questions/3268053/rails-before-validation-strip-whitespace-best-practices
* [Finishes #153377419]
https://www.pivotaltracker.com/story/show/153377419

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -77,6 +77,14 @@ class Member < ApplicationRecord
     key: Rails.application.secrets.secret_key_for_ssn_encryption,
   )
 
+  def first_name=(value)
+    super(value.try(:strip))
+  end
+
+  def last_name=(value)
+    super(value.try(:strip))
+  end
+
   def other_income_types_inclusion
     if (other_income_types - OTHER_INCOME_TYPES).any?
       errors.add(:other_income_types, "Not a valid income type")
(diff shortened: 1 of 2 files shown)


app/models/member.rb
```diff
@@ -77,6 +77,14 @@ class Member < ApplicationRecord
     key: Rails.application.secrets.secret_key_for_ssn_encryption,
   )
 
+  def first_name=(value)
+    super(value.try(:strip))
+  end
+
+  def last_name=(value)
+    super(value.try(:strip))
```
bengolder: I learned a lot from reading these methods as a beginner to ruby!

spec/controllers/introduce_yourself_controller_spec.rb
```diff
@@ -38,6 +38,26 @@
...
+          },
+        }
+
+        put :update, params: params
+        member.reload
+
+        expect(member.first_name).to eq "Space"
+        expect(member.last_name).to eq "Case"
+      end
+    end
```
bengolder: for using `.try` above, would it be worthwhile to add assertions about nil inputs?

PR #582, opened 2017-12-11
MiBridges: Add Dependent care expenses and details

jessieay:
* We were previously saying that "No one" had dependent care expenses;
now we are adding them when present and indicating the amount on the
follow up page.
* [Finishes #153570213]
* Confirmed that this works by adding `dependent_care: true` to the Snap
App in the MIBridges spec and running it - it passes.
https://www.pivotaltracker.com/story/show/153570213

### DIFF
--- a/spec/features/mi_bridges_driving_spec.rb
+++ b/spec/features/mi_bridges_driving_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 
 RSpec.feature "MI Bridges Driving" do
   # to run this in a non-headless way, remove the `:js` flag below
-  scenario "successfully drives, then re-drives a SNAP App", :driving do
+  scenario "successfully drives, then re-drives a SNAP App", :js, :driving do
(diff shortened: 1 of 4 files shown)


spec/features/mi_bridges_driving_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.feature "MI Bridges Driving" do
   # to run this in a non-headless way, remove the `:js` flag below
-  scenario "successfully drives, then re-drives a SNAP App", :driving do
+  scenario "successfully drives, then re-drives a SNAP App", :js, :driving do
```
luigi: Why do we need `:js` here?

jessieay: so it runs on a headless browser by default

luigi: 👍

PR #584, opened 2017-12-11
Link to SnapApplication from DriverError in admin

bengolder:
Administrate does not seem to make it particularly easy to add attributes that span across related tables.

This creates [customized page views](https://administrate-prototype.herokuapp.com/customizing_page_views) that link to related snap applications.

Totally open to the idea that we may not want to solve in this way, but it works.

[#153521242]

screen shot 2017-12-11 at 3 52 13 pm
screen shot 2017-12-11 at 3 52 02 pm

### DIFF
--- a/app/views/admin/driver_errors/_collection.html.erb
+++ b/app/views/admin/driver_errors/_collection.html.erb
@@ -1,0 +1,106 @@
+<%#
+# Collection
+
+This partial is used on the `index` and `show` pages
+to display a collection of resources in an HTML table.
+
+## Local variables:
+
+- `collection_presenter`:
+  An instance of [Administrate::Page::Collection][1].
+  The table presenter uses `ResourceDashboard::COLLECTION_ATTRIBUTES` to determine
+  the columns displayed in the table
+- `resources`:
+  An ActiveModel::Relation collection of resources to be displayed in the table.
+  By default, the number of resources is limited by pagination
+  or by a hard limit to prevent excessive page load times
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Page/Collection
+%>
+
+<table aria-labelledby="<%= table_title %>">
+  <thead>
+    <tr>
+
+      <th class="cell-label cell-label--application_id">SNAP Application</th>
+
+      <% collection_presenter.attribute_types.each do |attr_name, attr_type| %>
+        <th class="cell-label
+        cell-label--<%= attr_type.html_class %>
+        cell-label--<%= collection_presenter.ordered_html_class(attr_name) %>"
+        scope="col"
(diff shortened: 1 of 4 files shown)


app/views/admin/driver_errors/_collection.html.erb
```diff
@@ -0,0 +1,106 @@
...
+          <% end %>
+          >
+
+        <td class="cell-data cell-data--application">
+          <%=  link_to(
+                   resource.driver_application.snap_application_id,
+                   [namespace, resource.driver_application.snap_application],
+           ) %>
+
+        </td>
```
bengolder: this is the new column data in the index view.

app/views/admin/driver_errors/_collection.html.erb
```diff
@@ -0,0 +1,106 @@
...
+  or by a hard limit to prevent excessive page load times
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Page/Collection
+%>
+
+<table aria-labelledby="<%= table_title %>">
+  <thead>
+    <tr>
+
+      <th class="cell-label cell-label--application_id">SNAP Application</th>
```
bengolder: This is the new column header in the index view. Nearly all other HTML was created by a generator.

app/views/admin/driver_errors/index.html.erb
```diff
@@ -0,0 +1,61 @@
+<%#
```
bengolder: This file was completely generated, but is necessary.

app/views/admin/driver_errors/show.html.erb
```diff
@@ -0,0 +1,58 @@
...
+  <dl>
+    <dt class="attribute-label" id="snap_application_id>">
+      SNAP Application
+    </dt>
+    <dd class="attribute-data attribute-data--snap_application_id">
+      <%=  link_to(
+               page.resource.driver_application.snap_application_id,
+               [namespace, page.resource.driver_application.snap_application],
+           ) %>
+    </dd>
```
bengolder: The above description term and description detail elements were added. All other HTML was generated.

spec/features/admin_dashboard_driver_errors_spec.rb
```diff
@@ -18,9 +18,19 @@
     visit admin_root_path
 
     click_on "Driver Errors"
+
+    expect(page).to have_link(
+      driver_error.driver_application.snap_application_id,
+    )
```
bengolder: I'm not sure if there is a more preferrable way to assert this link exists.

jessieay: this seems good to me!

PR #586, opened 2017-12-12
Ignore .DS_Store files

luigi:
Thanks macOS!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@
 mkmf.log
 Guardfile
 .idea
+.DS_Store


jessieay:
I usually put things like this, that is, machine-specific stuff, in my global `~/.gitignore` file. 

I've noticed that we added a few other gitignore things that are specific to people's setups (eg: rubymine) so adding this is totally consistent with that, just wanted to mention.

bengolder:
Talked more with Jessie, and we both agree it would ideal to have machine or editor configs be shared, but separate from project repo. But adding this only helps doesn't hurt, so I say 👍  with the hope we can identify more shared configs outside of repo and bring them in during setup.

PR #590, opened 2017-12-13
Chromedriver usage

hartsick:
Hey all!

This started with @luigi and I investigating those `checking for chrome... no / checking for chromedriver... yes` messages before running every command. It turns out that those are from Capybara registering our drivers while loading the Rails app and are unavoidable, but I took the opportunity while digging to do some moving around. Pushing this up to get your thoughts.

Previously, we stored all webdriving code in lib/ and then eager loaded the webdriving directories to make them available for the production application. In order to make lib available to the Rails application in all environments, we need to either manually require certain files or eager load them (production doesn't autoload).

While tracking down the chrome and chromedriver loading, we noticed that there were two places where we were trying to load the driving code previously--one in application.rb, which wasn't working correctly because it was autoloading (ignored in prod), and another in initializers/extensions.rb, which was loading the app correctly.

This PR moves the driving code to app/lib, where the files will be automatically auto-loaded, and removes the places where we were manually trying to load it before. It also manually requires files in tasks (like our summarizer) who relied on the lib/ directory being automatically loaded.

The boundary is kinda fuzzy, but I'd propose the following separation between lib/ and app/lib moving forward:

* `app/lib` any library-esque code the application relies on (now: webdriving, PDF filling)
* `lib` any library-esque code that our app doesn't rely on (now: rake tasks, Selenium monkeypatch used only in scripts)

Let me know what you think!


hartsick:
@jessieay I did check this with eagerload set to true in dev! All went off without a hitch. I'm a little worried about testing gaps, but did some spot checking in the app / webdriving / running our summarizer rake task and all looked good! Said this in Slack, but I was so grateful for your and @jayroh's super clear and helpful commit messages around that topic!

PR #597, opened 2017-12-14
Add <label>s to date selects for a11y!

jessieay:
* Monkey-patches build_selects_from_types because:
  * No default way to pass a label for a date
    selector in Rails
  * Could get around this using the date_separator
    option, but that didn’t work because
    date_separator would be same for every
    element and we need them to be dynamic to
    match the ID of the input.
  * Also, date_separator always skips the first
    date select field because date_separator is
    pre-pended to the field.

[#153646862]

### DIFF
--- a/app/helpers/mb_form_builder.rb
+++ b/app/helpers/mb_form_builder.rb
@@ -175,9 +175,11 @@ class MbFormBuilder < ActionView::Helpers::FormBuilder
       <fieldset class="form-group#{error_state(object, method)}">
         #{fieldset_label_contents(label_text, notes)}
         <div class="input-group--inline">
-          <div class="select">
-            #{date_select(method, { autofocus: autofocus, date_separator: '</div><div class="select">' }.merge(options), class: 'select__element')}
-          </div>
+          #{date_select(
+            method,
+            { autofocus: autofocus }.merge(options),
+            class: 'select__element',
+          )}
         </div>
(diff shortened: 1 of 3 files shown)


app/helpers/mb_form_builder.rb
```diff
@@ -397,3 +399,39 @@ def check_box_with_label(label_text, method, options = {})
...
+            "<div class='select'><label for='#{field_name}_3i' class='sr-only'>Day</label>"
+          when :hour
+            (@options[:discard_year] && @options[:discard_day]) ? "" : @options[:datetime_separator]
+          when :minute, :second
+            @options[:"discard_#{type}"] ? "" : @options[:time_separator]
+        end
+      end
+    end
+  end
+end
```
bengolder: Would it feel better if our `mb_date_select` just grabbed each select individually instead of monkey patching (`select_year`, `select_month`, `select_day`) (assuming it's possible to do so and get labels)?

jessieay: yeah we kinda looked into that and in the end it felt like it was going to take more time. *but* perhaps the time investment is worth it?

luigi: I'm gonna explore the individual `select_*` methods approach today.

jessieay:
Closing in favor of https://github.com/codeforamerica/michigan-benefits/pull/604

PR #599, opened 2017-12-15
Add `sr-only` labels for inputs without labels

jessieay:
* These labels are in the page header so they are visible to people
looking at a browser but will not be read by a screen reader.
* [#153646862]


jessieay:
@luigi finished up what we started! let me know what you think

luigi:
Ah, looks like there's a failing test though.

jessieay:
@luigi fixed at c2f30b18

PR #601, opened 2017-12-15
Require answer to authorized representative question

jessieay:
* Before, somehow could skip this question
* There was a related bug where if they answered "no" it would require
the auth rep name because "false" was being evaluated to a present value
in the step validation.
* Now, the bug is fixed and we have slightly different behavior too.
* [Finishes #153690105]

screen shot 2017-12-15 at 9 38 10 am

screen shot 2017-12-15 at 9 45 00 am

### DIFF
--- a/app/views/authorized_representative/edit.html.erb
+++ b/app/views/authorized_representative/edit.html.erb
@@ -17,12 +17,16 @@
       builder: MbFormBuilder,
       url: current_path,
       method: :put do |f| %>
-    <%= f.mb_radio_set(
-      :authorized_representative,
-      "Would you like to designate someone else to be your authorized representative?",
-      [{ value: :true, label: "Yes" }, { value: :false, label: "No" }],
-      layout: "inline") %>
-    <%= f.mb_input_field :authorized_representative_name, "What is their full legal name?" %>
+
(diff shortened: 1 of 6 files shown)


app/views/authorized_representative/edit.html.erb
```diff
@@ -17,12 +17,16 @@
...
-      "Would you like to designate someone else to be your authorized representative?",
-      [{ value: :true, label: "Yes" }, { value: :false, label: "No" }],
-      layout: "inline") %>
-    <%= f.mb_input_field :authorized_representative_name, "What is their full legal name?" %>
+
+    <div class="form-group">
+      <%= f.mb_radio_set(
+        :authorized_representative,
+        "Would you like to designate someone else to be your authorized representative?",
+        [{ value: :true, label: "Yes" }, { value: :false, label: "No" }],
```
luigi: If we change the symbols to booleans (`:true:` to `true`) and (`:false` to `false`), would that address the issue of checking for the string `"false"` in the step?

jessieay: the controller params will always be strings, as far as I an tell.

PR #604, opened 2017-12-16
Add a <label> to each component in a date select

luigi:
This is an alternative approach to #597. Instead of monkeypatching Rails, it:

* Drops `date_select` in favor of `select_month`,
  `select_day`, `select_year`
* Adds some helper methods to format `id=` and `name=`

It adds the same test as #597.

https://www.pivotaltracker.com/story/show/153646862


bengolder:
I like this solution because if we have to make HTML adjustments later (such as linking help text or validation errors using `aria-labelledby` or `aria-describedby`) it seems like it would be easier.

PR #609, opened 2017-12-19
Update legal agreement pages for a11y

jessieay:
* Fix invalid HTML as detected by WAVE
* Fix contrast issue
* Contrast issue brought up issue with our overrides, which was that we
load our overrides last in the manifest file so that they take
precedent, which means that our variable overrides were not loaded until
after all of the base styles. Now, we are loading our variable overrides
above the base styles so that the base styles that use variables are
using our overrides rather than the original colors.
* [Finishes #153764205]
* [Finishes #153764168]

screen shot 2017-12-19 at 9 14 25 am

### DIFF
--- a/app/views/shared/_terms.html.erb
+++ b/app/views/shared/_terms.html.erb
@@ -605,7 +605,7 @@ Other eligible members of your household will still be able to receive help.
 <ul>
   <li>
     Online at: <a
-      href="http://www.michigan.gov/domesticviolence">www.michigan.gov/domesticviolence"</a>
+      href="http://www.michigan.gov/domesticviolence">www.michigan.gov/domesticviolence</a>
   </li>
   <li>
     DHS Publication 859, Is Someone Hurting You or Your Children? (also
@@ -616,7 +616,7 @@ Other eligible members of your household will still be able to receive help.
 <ul>
   <li>
     Online at: <a
-      href="http://www.michigan.gov/dhs-publications">www.michigan.gov/dhs-publications"</a>
+      href="http://www.michigan.gov/dhs-publications">www.michigan.gov/dhs-publications</a>
   </li>
 </ul>
 
--- a/app/views/shared/_terms.html.erb
+++ b/app/views/shared/_terms.html.erb
@@ -520,73 +520,123 @@ http://www.michigan.gov/disabilityresources/
 >This form may be printed and sent to:</p><p
 >MDHHS, Americans with Disabilities Act
 Coordinator </p><p>P.O. Box 30037, Suite 708 </p><p>Lansing, MI 48909 </p><p>(855)
-275-6424</p><p
->Citizens and Non-Citizens /Social Security
-Numbers</p><p
->Social Security numbers and immigration papers are
-NOT required for a person who is :</p><ul
-><li>Not
-applying for help.</li><li>An undocumented
-non-citizen only applying for medical assistance for emergency services,
-pregnancy or child birth.</li><li>Only
-applying for child care. (You must give a Social Security number for the child
-and the child must be a U.S. citizen or show immigration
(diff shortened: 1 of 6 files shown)


app/views/shared/_terms.html.erb
```diff
@@ -724,20 +774,17 @@ meals.</p>
...
   <li>Applying for Assistance: <a href="http://www.michigan.gov/dhs-applicationprocess">www.michigan.gov/dhs-applicationprocess</a></li>
-  <li><a href="http://www.michigan.gov/dhs-applicationprocess"></a></li>
   <li>Cash Assistance: <a href="http://www.michigan.gov/dhs-cash">www.michigan.gov/dhs-cash</a></li>
-  <li ><a href="http://www.michigan.gov/dhs-cash"></a></li>
-    <li >Emergency Services: <a href="http://www.michigan.gov/dhs-ser">www.michigan.gov/dhs-ser</a></li><li
-><a href="http://www.michigan.gov/dhs-ser"></a></li><li
-    >Food Assistance: <a href="http://www.michigan.gov/foodstamps">www.michigan.gov/foodstamps"</a></li>
-    <li >Medical Services: <a href="http://www.michigan.gov/dhs-medical">www.michigan.gov/dhs-medical</a></li>
+  <li>Emergency Services: <a href="http://www.michigan.gov/dhs-ser">www.michigan.gov/dhs-ser</a></li>
+    <li>Food Assistance: <a href="http://www.michigan.gov/foodstamps">www.michigan.gov/foodstamps"</a></li>
```
luigi: Appears to have indented accidentally starting on this line. This line also has a dangling `"` before `</a>`.

app/views/shared/_terms.html.erb
```diff
@@ -520,73 +520,123 @@ http://www.michigan.gov/disabilityresources/
...
+</p>
 
 <p>Resources:</p>
 
 <ul>
-  <li >Online at: <a href="http://www.michigan.gov/domesticviolence>www.michigan.gov/domesticviolence"</a></li>
-  <li >DHS Publication 859, Is Someone Hurting You or Your Children? (also available in Spanish) -</li>
+  <li>
+    Online at: <a
+      href="http://www.michigan.gov/domesticviolence">www.michigan.gov/domesticviolence"</a>
```
luigi: Dangling `"` before `</a>`.

app/views/shared/_terms.html.erb
```diff
@@ -520,73 +520,123 @@ http://www.michigan.gov/disabilityresources/
...
+    DHS Publication 859, Is Someone Hurting You or Your Children? (also
+    available in Spanish)
+  </li>
 </ul>
 
 <ul>
-  <li>Online at: <a href="http://www.michigan.gov/dhs-publications>www.michigan.gov/dhs-publications"</a></li>
+  <li>
+    Online at: <a
+      href="http://www.michigan.gov/dhs-publications">www.michigan.gov/dhs-publications"</a>
```
luigi: Dangling `"` before `</a>`.

app/views/shared/_terms.html.erb
jessieay: oh yeah that's in an attempt to make the lines shorter :)

app/views/shared/_terms.html.erb
jessieay: same as before ;)

app/views/shared/_terms.html.erb
jessieay: good catch!

fixed at a3e34ad0

app/views/shared/_terms.html.erb
jessieay: woops, didnt read your comment correctly

fixed at 19db043f

PR #614, opened 2017-12-20
Run accessibility checks in feature specs using `axe-matchers`

jessieay:
* https://github.com/dequelabs/axe-matchers
* We are doing an accessibility check for each page using the
`proceed_with` method, which is really just an alias for `click_on` plus
the accessibility check.
* In some cases, doing the accessibility check scrolls the page in such
a way that the button is underneath the header element on the page,
which Capybara does not like. So, we need to scroll to the top on those
cases to bring the button to the foreground of the page.
* [Finishes #152982764]
https://www.pivotaltracker.com/story/show/152982764


luigi:
Just tested the whole suite locally: these checks add about 45 seconds to the total run.

jessieay:
oh yeah -- @bengolder @hartsick, to expand upon Luigi's comment about how this affects our test suite run time: Luigi and I paired on this but want feedback on from you both.

As I see it:

positives to this approach:
- do not need to write new feature specs to run axe-matchers
- have high confidence that we are hitting all or almost all pages

negatives to this approach:
- slows down our overall test suite (could write it another way where it is not run as part of `rake`)
- combines our accessibility tests with our feature tests, which would mean a failing test suite if we get an a11y error (or is this a positive?)

hartsick:
Yay! I think this is a great start. For suite timing, let's just try it out and see how we feel. If it's proving too slow or cumbersome, we can revisit. Thanks for completing this—so happy it's done!

PR #621, opened 2017-12-21
Add unit test for MiBridges driver

jessieay:
* This is mostly mocking and stubbing.
* [#153527657]

### DIFF
--- a/app/lib/mi_bridges/driver.rb
+++ b/app/lib/mi_bridges/driver.rb
@@ -129,15 +129,15 @@ module MiBridges
     def run_flow(flow)
       flow.each do |klass|
         begin
-          @page = klass.new(@snap_application, logger: logger)
-          @page.setup
-          @page.fill_in_required_fields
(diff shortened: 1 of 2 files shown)


app/lib/mi_bridges/driver.rb
```diff
@@ -129,15 +129,15 @@ def complete_application
     def run_flow(flow)
       flow.each do |klass|
         begin
-          @page = klass.new(@snap_application, logger: logger)
-          @page.setup
-          @page.fill_in_required_fields
-          @page.continue
+          mi_bridges_page = klass.new(snap_application, logger: logger)
```
jessieay: I renamed these to clarify intent -- below, we use `page`, in the `page_title` method...which is actually looking to to the capybara `page` method rather than the `page` that we instantiate here, but that was not clear with our previous naming

bengolder: good catch! Thank you

PR #623, opened 2017-12-22
Add current date and applicant name to attachment filename

jessieay:
* Previously, were using generic names
* This way, case workers can more easily identify which files are for
which applicants
* [Finishes #153763110]


jessieay:
NOTE: the code climate error is a false positive. will be fixed once this PR is merged/deployed: https://github.com/bbatsov/rubocop/pull/5230

PR #629, opened 2018-01-06
[PAIR] Apply for Medicaid on the homepage

racheledelman:
https://www.pivotaltracker.com/story/show/154034941
Added Medicaid button + copy changes to homepages.

_Note: this makes dual_index redundant. We will follow up this PR with another one that includes a refactor to remove it._

Office homepage:
image

Main homepage:
image


hartsick:
Quick note: Paras and I removed the dual-index as part of this PR

PR #651, opened 2018-01-25
Remove fax-sending code from codebase.

luigi:
* Keeps fax-related code related to viewing old
exports in the admin area.

[#154093350]


hartsick:
I LOVE DELETING CODE

PR #655, opened 2018-01-26
Send blank PDF for SNAP apps if feature flag enabled.

luigi:
For SNAP applications, if `NEW_FORM_ENABLED` is `true`, we build a barely-filled AssistanceApplicationForm with attached coversheet and verification documents. This will be included in office or applicant emails and downloadable in the admin.

Anything being assembled by the `ApplicationPdfAssembler` should implement a "pdf component" interface (described in a `spec/support/shared_examples/pdf_component`). `PdfConcatenator` will call the `.fill?` method to determine if the component needs to be filled, and will call `.output_file` to retrieve the file object associated with the component.

We updated some existing objects to implement the pdf component interface (`.fill?` and `.output_file`)

[Finishes #154566400]

### DIFF
--- a/app/models/dhs1171_pdf.rb
+++ b/app/models/dhs1171_pdf.rb
@@ -89,7 +89,7 @@ class Dhs1171Pdf
       VerificationDocument.new(
         url: document,
         benefit_application: snap_application,
-      ).file
+      ).output_file
     end.reject { |document| document.path.nil? }
   end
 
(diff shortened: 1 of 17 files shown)


luigi:
@hartsick Because this involved changing the interface for how `PdfConcatenator` distinguishes between fillable and non-fillable PDF components, we'd love your feedback.

hartsick:
this is so satisfying

PR #658, opened 2018-01-30
 Add `/numbers` dashboard with MichiganBenefits stats

hartsick:
We built our own dashboard (older GCF-style) after talking a bit with Lou. This solution was quicker and easier (and cheaper!) than Geckoboard since currently all of our data is located inside our application.

User + password info in LastPass. Used same info for staging + production since it's non-sensitive data.


hartsick:
Hm, code climate seems to be running Rubocop and its config is different than in our Rake task

PR #659, opened 2018-01-30
Correct incorrect citizen reporting

hartsick:
Previously in this step on the SNAP flow, when the client selected "everyone is a citizen", we only marked that on the application and didn't update individual members. As a result, when we fill in the PDF, the members were never marked as citizens on the PDF, because we only fill it out based on the member.

This PR fixes that by updating members-> citizen to true when 'everyone_a_citizen' is selected.

This PR also addresses the negative of that problem on the same step, which happens when someone marks 'anyone_blank' as false. Right now we aren't updating the members (as a result, the value for that field is nil). This PR marks those fields as false instead.

Lastly, we now update 'everyone_a_citizen' and similar fields to their proper values based on actual reported members (for example, we update to `true` if all members are selected, and `anyone_in_college` to `false` if no members are selected).

A problem: right now this says canonically that if e.g. no members are marked as disabled that there are no disabled members in the household. But, we specifically say that member selection is optional. Screenshot below.

screen shot 2018-01-30 at 2 07 44 pm

What do we want to do? It seems like there could be intentionally irreconcilable data, but we have no way of knowing whether that conflict is intentional (person just hasn't indicated who is disabled) or the result of user confusion (they would have gone back to mark that no one in household was disabled if given the chance).

### DIFF
--- a/app/controllers/household_more_info_controller.rb
+++ b/app/controllers/household_more_info_controller.rb
@@ -4,12 +4,26 @@ class HouseholdMoreInfoController < SnapStepsController
   def update_application
     super
 
-    current_application.members.each do |member|
-      member.update!(member_attrs)
+    if single_member_household?
+      current_application.primary_member.update!(member_attrs)
+    else
+      current_application.members.each do |member|
+        member.update!(multi_member_attrs)
+      end
     end
   end
 
   def member_attrs
+    {
+      citizen: step_params[:everyone_a_citizen],
+      disabled: step_params[:anyone_disabled],
+      new_mom: step_params[:anyone_new_mom],
+      in_college: step_params[:anyone_in_college],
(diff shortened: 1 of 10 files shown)


hartsick:
Update: as far as I can tell, we're never reporting the 'living elsewhere' field on member.

bengolder:
Thank you for adding such thorough tests!

hartsick:
> What do we want to do? It seems like there could be intentionally irreconcilable data, but we have no way of knowing whether that conflict is intentional (person just hasn't indicated who is disabled) or the result of user confusion (they would have gone back to mark that no one in household was disabled if given the chance).

Note: after talking to Alan, decided to remove the '(optional)' from fields we want them to fill.

PR #660, opened 2018-01-31
Avoid marking applications as homeless when stable_housing is nil.

luigi:
* On Medicaid applications, stable_housing might be nil.
* Rewrite stable_housing? to be true when it’s nil.
* Correct unstable_housing? to return false only on boolean false.

[#154776068]

### DIFF
--- a/app/models/concerns/common_benefit_application.rb
+++ b/app/models/concerns/common_benefit_application.rb
@@ -27,7 +27,11 @@ module CommonBenefitApplication
     addresses.where(mailing: true).first || NullAddress.new
   end
 
+  def stable_housing?
+    stable_housing == true || stable_housing == nil
+  end
+
   def unstable_housing?
-    !stable_housing?
+    stable_housing == false
   end
 end
--- a/spec/support/shared_examples/common_benefit_application.rb
+++ b/spec/support/shared_examples/common_benefit_application.rb
@@ -144,5 +144,39 @@ RSpec.shared_examples "common benefit application" do
         expect(subject.unstable_housing?).to eq(true)
       end
     end
+
+    context "when stable_housing is nil" do
+      it "returns false" do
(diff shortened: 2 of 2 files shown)


hartsick:
Great! I think it's likely we have this problem in a number of places. I'm planning to do an audit of the 1171 & 1426 PDFs soon to try to surface other inaccuracies like this.

app/models/concerns/common_benefit_application.rb
```diff
@@ -27,7 +27,11 @@ def mailing_address
     addresses.where(mailing: true).first || NullAddress.new
   end
 
+  def stable_housing?
+    stable_housing == true || stable_housing == nil
```
hartsick: Why would stable_housing be true or nil?

luigi: We didn't set a default value in the schema for it in `medicaid_applications`:

https://github.com/codeforamerica/michigan-benefits/blob/master/db/schema.rb#L171

Unlike for `snap_applications`, where we did:

https://github.com/codeforamerica/michigan-benefits/blob/master/db/schema.rb#L273

So during a Medicaid application, if the client never indicates they are homeless, the value is just never set to anything (which is `nil`). Since `nil` is falsey, that's why the checkbox was being filled on the 1426 PDF.

luigi:
@hartsick Wrote up a task to audit here: https://www.pivotaltracker.com/story/show/154788371

app/models/concerns/common_benefit_application.rb
hartsick: Gotcha. I feel like we might want to update this to true in the controller for now, until we move toward a better way of reflecting state. So, that way we have a sure way of telling whether someone has answered affirmatively or whether they haven't answered at all. But, I don't think that this should block the feature getting out.

PR #669, opened 2018-02-07
Add Olark live chat to some views

bengolder:
- Relies on a `LIVE_CHAT_ENABLED` feature flag
- Uses a shared erb snippet

[Finishes #154901171]


bengolder:
@hartsick yeah! I'll do that

PR #671, opened 2018-02-07
Ask if user has applied before

luigi:
- Use enum for unfilled/yes/no.
- Used backfill recommendation via Strong Migrations
https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value
- Updated yes/no JavaScript to support the enum "yes"/"no" values

[Finishes #154901167]

### DIFF
--- a/spec/controllers/personal_detail_controller_spec.rb
+++ b/spec/controllers/personal_detail_controller_spec.rb
@@ -36,7 +36,7 @@ RSpec.describe PersonalDetailController do
           expect(current_app_member.send(key)).to eq(value)
         end
 
-        expect(response).to redirect_to("/steps/introduction-applied-before")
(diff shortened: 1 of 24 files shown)


spec/controllers/personal_detail_controller_spec.rb
```diff
@@ -36,7 +36,7 @@
           expect(current_app_member.send(key)).to eq(value)
         end
 
-        expect(response).to redirect_to("/steps/household-members-overview")
+        expect(response).to redirect_to("/steps/introduction-applied-before")
```
hartsick: we might want to just make this `next_path`

hartsick:
Surfacing a convo Ben and I just had: for new flow, let's consider making the JS-forms just regular ol' HTML forms! And just generally make our HTML more consistent.

PR #687, opened 2018-02-28
Ask applicant whether they are a Michigan resident

hartsick:
And directs them to Healthcare.gov if they aren't.

Also creates 'ApplicationNavigator' object, which is intended to store application state and metadata. Should not be used to fill out final form. For now, we're trying out adding database defaults to direct application flow.

[#155521552]


app/views/integrated/_yes_no_form.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="form-card__content">
+  <footer class="form-card__buttons">
```
hartsick: @bengolder Check it out! No more Javascript 🎉

PR #688, opened 2018-02-28
Ask if applicant has previously received benefits

hartsick:
* Creates special Integrated::PdfAttributes for filling out the new 1171 (old was getting messy, and there's not a ton of overlap with the existing PdfAttributes class)
* Creates database defaults for enums, and sets them to 'unfilled'
* Fills out PDF radio only if field has been answered

[Finishes #155532806]

### DIFF
--- a/app/forms/introduce_yourself_form.rb
+++ b/app/forms/introduce_yourself_form.rb
@@ -1,12 +1,20 @@
 class IntroduceYourselfForm < Step
   include MultiparameterAttributeAssignment
 
-  form_attributes(
-    :first_name,
-    :last_name,
-    :birthday,
-    :sex,
-  )
+  def self.application_attributes
+    [:previously_received_assistance]
+  end
+
+  def self.member_attributes
+    %i[
+      first_name
+      last_name
+      birthday
+      sex
+    ]
+  end
+
+  form_attributes(*(member_attributes + application_attributes))
 
   validates :first_name,
     presence: { message: "Make sure to provide a first name" }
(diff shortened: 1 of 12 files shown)


app/forms/introduce_yourself_form.rb
```diff
@@ -1,12 +1,20 @@
...
+  def self.member_attributes
+    %i[
+      first_name
+      last_name
+      birthday
+      sex
+    ]
+  end
+
+  form_attributes(*(member_attributes + application_attributes))
```
hartsick: Particularly interested in thoughts about this approach—we've got some kind of awkward separation of attributes between different models so thought of this as one way to keep that knowledge of attribute names entirely inside the form model (we have put it also in the controller before, which I wanted to avoid if we could!)

luigi: I definitely like the reasoning behind it. It was non-obvious to me at first why `.application_attributes` and `.member_attributes` had to be class methods, but then I looked at `Step` and saw that `.form_attributes` was as well.

We always know which models will take in data from forms. So we could define class methods on `Step` specifically for attributes on each model, and use them in the form classes like so:

```ruby
application_attributes(:previously_received_assistance)

member_attributes(:first_name, :last_name)

navigator_attributes(:resides_in_state)

employment_attributes(:employer_name, :hours_per_week)
```

Then `Step.form_attributes` would combine them as appropriate. It would keep our form definitions succinct but expressive.

hartsick: Yeah, I like this suggestion! This also potentially handles some of the complexity of updating models—we can consider standardizing it in the controller using these methods. Let's tackle this refactor as part of the next feature.

PR #689, opened 2018-02-28
Indicate living situation

luigi:
image

### DIFF
--- a/app/forms/living_situation_form.rb
+++ b/app/forms/living_situation_form.rb
@@ -1,0 +1,12 @@
+class LivingSituationForm < Step
+  def self.application_attributes
+    [:living_situation]
+  end
+
+  form_attributes(*application_attributes)
+
+  validates :living_situation, inclusion: {
+    in: %w(stable_address temporary_address homeless),
+    message: "Make sure to answer this question",
+  }
+end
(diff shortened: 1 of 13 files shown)


app/models/common_application.rb
```diff
@@ -15,6 +15,8 @@ class CommonApplication < ApplicationRecord
   enum previously_received_assistance: { unfilled: 0, yes: 1, no: 2 },
        _prefix: :previously_received_assistance
 
+  enum living_situation: { unknown_living_situation: 0, stable_address: 1, temporary_address: 2, homeless: 3 }
```
luigi: I didn't add a `_prefix` here, as I felt the usage was better without it:

```
current_app.unknown_living_situation?
current_app.stable_address?
current_app.permanent_address?
current_app.homeless?
```

The downside I see is that we don't have a consistent usage with other enums.

hartsick: Makes sense to me. Reads well. If we feel pain around the inconsistency later, we can easily refactor!

app/forms/living_situation_form.rb
```diff
@@ -0,0 +1,12 @@
+class LivingSituationForm < Step
+  def self.application_attributes
+    [:living_situation]
+  end
+
+  form_attributes(*application_attributes)
```
hartsick: How would you feel about either not defining application_attributes (since we don't use in the controller) or using something like application_params (rather than form_params) to update the model in the controller?

luigi: I like using application_params in the controller. Just added that in 279c4aa3b5d02ae4be59a28b30afc017480da779.

It's more clear: "I'm updating the application, so I should use the application_params."

PR #691, opened 2018-03-02
Refactor 'member_attributes' and 'application_attributes'

hartsick:
Creates new 'form' class (based off of 'step') that uses a combination of simple 'member_attributes' and 'application_attributes' to construct one set of common form attributes. Getters for these methods can then be used to filter params as appropriate for their respective models.

For now, 'member_attributes' and 'application_attributes' only accept keys or strings as attributes, unlike 'step_attributes' method, which accepts hashes. We can add this in later as we need it.

[#155522274]

Note: this is step one of the feature (have more on my local machine that I'll push to a WIP on another branch), but thought it'd be useful to get your thoughts sooner rather than later!

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -91,10 +91,19 @@ class FormsController < ApplicationController
   end
 
   def member_params
-    form_params.except(*form_class.application_attributes)
+    filter_params(form_class.member_attributes)
   end
 
   def application_params
-    form_params.slice(*form_class.application_attributes)
+    filter_params(form_class.application_attributes)
+  end
+
+  def filter_params(attributes)
+    return unless attributes.present?
+    form_params.select do |key, _|
+      attributes.any? do |attr|
+        key.match(/^#{attr}(\(.+\))?\z/) # Match single and multi-attribute keys
+      end
+    end
   end
 end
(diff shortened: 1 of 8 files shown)


hartsick:
Aaand, here's the WIP branch with more actual feature work: https://github.com/codeforamerica/michigan-benefits/tree/integrated_add_members_wip

app/controllers/forms_controller.rb
```diff
@@ -91,10 +91,19 @@ def form_navigation
...
-    form_params.except(*form_class.application_attributes)
+    filter_for_single_and_multiattribute_params(form_class.member_attributes)
   end
 
   def application_params
-    form_params.slice(*form_class.application_attributes)
+    filter_for_single_and_multiattribute_params(form_class.application_attributes)
+  end
+
+  def filter_for_single_and_multiattribute_params(attributes)
```
hartsick: maybe just name 'filter_params'?

app/controllers/forms_controller.rb
```diff
@@ -91,10 +91,19 @@ def form_navigation
...
+    return unless attributes.present?
+    form_params.select do |key, _|
+      attributes.any? do |attr|
+        key.match(/^#{attr}(\(.+\))?\z/)
```
hartsick: Add comment about use

hartsick:
Reviewed this with Luigi and added comments above where we discussed them! Just addressed the comments (and the failing test) in the amended commit.

PR #700, opened 2018-03-09
Choose household members requesting food assistance.

luigi:
your household 2018-03-09 18-48-49

When unstable housing, the checkbox screen is skipped, and
all members of the "household" are set to request food.

[#155522792]

Signed-off-by: Christa Harstock <christa@codeforamerica.org>
Co-authored-by: Christa Harstock <christa@codeforamerica.org>

### DIFF
--- a/app/controllers/integrated/food_assistance_controller.rb
+++ b/app/controllers/integrated/food_assistance_controller.rb
@@ -1,0 +1,37 @@
+module Integrated
+  class FoodAssistanceController < FormsController
+    def self.skip?(current_application)
+      if current_application.members.count == 1 || current_application.unstable_housing?
+        current_application.members.each { |member| member.update(requesting_food: "yes") }
+        true
+      else
+        false
+      end
+    end
+
+    def edit
+      if members.first.requesting_food_unfilled?
+        members.first.update(requesting_food: "yes")
+      end
+      @form = form_class.new(members: members)
+    end
+
+    def update_models
+      members.each do |member|
+        attrs = params.dig(:form, :members, member.to_param)
+        member.assign_attributes(attrs.permit(form_class.member_attributes))
+      end
+      ActiveRecord::Base.transaction { members.each(&:save!) }
+    end
+
+    private
+
+    def members
+      @_members ||= current_application.members
+    end
+
+    def form_params
+      params.fetch(:form, {}).permit(members: {})
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180305192048) do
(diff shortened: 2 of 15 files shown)


app/controllers/integrated/food_assistance_controller.rb
```diff
@@ -0,0 +1,37 @@
+module Integrated
+  class FoodAssistanceController < FormsController
+    def self.skip?(current_application)
+      if current_application.members.count == 1 || current_application.unstable_housing?
+        current_application.members.each { |member| member.update(requesting_food: "yes") }
```
hartsick: I feel a little hesitant to do database updates in the skip? method since it's a class method called during the course of navigation, rather than once we visit the controller. I think it should only get called when we're trying to figure out if we should visit that page, but I think that we can't guarantee that's true with the existing interface. but, it's also nice to not have to tackle this update in an unrelated controller. would love to talk about this when we implement the next time!

db/schema.rb
```diff
@@ -10,9 +10,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180305192048) do
+ActiveRecord::Schema.define(version: 20180308214556) do
 
   # These are extensions that must be enabled in order to support this database
+  enable_extension "pg_stat_statements"
```
hartsick: this is because of crap in my local base, need to reset

db/schema.rb
```diff
@@ -66,6 +67,27 @@
     t.index ["common_application_id"], name: "index_application_navigators_on_common_application_id"
   end
 
+  create_table "apps", id: :serial, force: :cascade do |t|
```
hartsick: also local db crap

db/schema.rb
```diff
@@ -323,6 +346,18 @@
     t.boolean "vehicle_income"
   end
 
+  create_table "users", id: :serial, force: :cascade do |t|
```
hartsick: also local db crap

db/schema.rb
```diff
@@ -323,6 +346,18 @@
...
+    t.datetime "created_at"
+    t.string "crypted_password"
+    t.string "email", null: false
+    t.string "name"
+    t.string "salt"
+    t.datetime "updated_at"
+    t.index ["email"], name: "index_users_on_email", unique: true
+  end
+
+  add_foreign_key "apps", "users"
```
hartsick: also local db crap

app/forms/form.rb
```diff
@@ -29,7 +29,9 @@ def member_attributes
     def form_attributes(attribute_names)
       self.attribute_names = (application_attributes + member_attributes)
 
-      attribute_strings = attribute_names.map(&:to_s)
+      attribute_strings = Step::Attributes.
```
hartsick: would like to add in tests around hash support in Form

PR #702, opened 2018-03-12
Enter birthday as text instead of select dropdown

bengolder:
- adds new spec helper `include_html`
- only changes inputs for *integrated* flow
- uses new `mb_date_input` method to `MbFormBuilder`
- renders form errors as spans instead of divs

### _Date input:_
screen shot 2018-03-12 at 4 11 15 pm

### _Date input with errors:_
screen shot 2018-03-12 at 4 12 14 pm

### DIFF
--- a/app/controllers/integrated/introduce_yourself_controller.rb
+++ b/app/controllers/integrated/introduce_yourself_controller.rb
@@ -8,7 +8,7 @@ module Integrated
         requesting_food: "yes",
         buy_and_prepare_food_together: "yes",
       )
-
+      combine_birthday_fields(member_data)
       if current_application
         current_application.primary_member.update(member_data)
         current_application.update(application_params)
@@ -21,10 +21,21 @@ module Integrated
 
     private
 
+    def combine_birthday_fields(data)
+      data[:birthday] = DateTime.new(
+        data.delete(:birthday_year).to_i,
+        data.delete(:birthday_month).to_i,
+        data.delete(:birthday_day).to_i,
+      )
+    end
+
     def existing_attributes
       if current_application
         attributes = current_application.attributes.
           merge(current_application.primary_member.attributes)
+        %i[year month day].each do |sym|
+          attributes["birthday_#{sym}"] = current_application.primary_member.birthday.try(sym)
+        end
         HashWithIndifferentAccess.new(attributes)
       else
         {}
--- a/app/controllers/integrated/introduce_yourself_controller.rb
+++ b/app/controllers/integrated/introduce_yourself_controller.rb
@@ -21,14 +21,6 @@ module Integrated
 
     private
 
-    def combine_birthday_fields(data)
-      data[:birthday] = DateTime.new(
-        data.delete(:birthday_year).to_i,
-        data.delete(:birthday_month).to_i,
-        data.delete(:birthday_day).to_i,
-      )
-    end
-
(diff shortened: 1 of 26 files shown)


app/controllers/integrated/introduce_yourself_controller.rb
```diff
@@ -17,10 +17,21 @@ def update_models
 
     private
 
+    def combine_birthday_fields(data)
+      data[:birthday] = DateTime.new(
+        data.delete(:birthday_year).to_i,
+        data.delete(:birthday_month).to_i,
+        data.delete(:birthday_day).to_i,
+      )
+    end
```
luigi: At first I was wondering if we could continue to use Rails' built-in `(2i)`, `(3i)` facility. But this is probably more clear. 👍

spec/support/include_html.rb
```diff
@@ -0,0 +1,18 @@
...
+  match do |actual|
+    @actual = clean_html(actual)
+    @expected = clean_html(expected)
+    @actual.include? @expected
+  end
+
+  failure_message do |_actual|
+    "EXPECTED:\n#{@actual}\n\nTO INCLUDE:\n#{@expected}"
+  end
+end
```
luigi: :metal:

app/forms/introduce_yourself_form.rb
```diff
@@ -21,11 +21,18 @@ class IntroduceYourselfForm < Form
...
   }
 
-  validates :birthday,
-    presence: { message: "Make sure to provide a birthday" }
+  validate :birthday_must_be_present_and_valid_date
 
-  # https://github.com/rails/rails/pull/8189#issuecomment-10329403
-  def class_for_attribute(attr)
-    return Date if attr == "birthday"
+  def birthday_must_be_present_and_valid_date
```
hartsick: It looks like some of this logic is duplicated at the model and controller level. Do you think we could extract this into a shared helper and potentially move the logic that's currently into the controller into a getter/setter for birthday on the model (/in the shared helper)?

bengolder:
@hartsick thank you! @luigi & I refactored the validation.

Regarding the error location, great call out. I did put it on top in response to observing screen reader usage, which is not consistent with the rest of our fields. I will loop back around with Rachel to check if that was a sound decision, and if I should change in other places as well.

One other caveat I should mention: This does not accept a partial birthday for other household members. It's all or nothing. If only year is filled out, it would save `nil`. I'm not sure how we can correctly handle that situation without storing month, day, and year in separate db fields.

hartsick:
> One other caveat I should mention: This does not accept a partial birthday for other household members. It's all or nothing. If only year is filled out, it would save nil. I'm not sure how we can correctly handle that situation without storing month, day, and year in separate db fields.

That seems fine to me! It think partial dates are something we'd have to handle specially in the PDF, anyway, so worth thinking about some more.

@racheledelman Could you take a look at Ben's comments above on the validation placement and screenreaders? If you have questions, plz let me know! Going to merge this in for now, and if we want to change placement of validations (here or elsewhere), we can do in another branch.

bengolder:
Talked with @racheledelman. The verdict is to put the errors at the end of fields consistently and wait for further screenreader testing to indicate whether it needs changing.

luigi:
Thanks @bengolder. @hartsick and I will take that on.

PR #705, opened 2018-03-15
Indicate household members who buy and prepare food separately

luigi:
your household 2018-03-14 17-35-07
your household 2018-03-14 17-35-30
pdf 2018-03-14 17-36-32
pdf 2018-03-14 17-36-50

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180314145539) do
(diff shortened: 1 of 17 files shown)


db/schema.rb
```diff
@@ -325,6 +326,11 @@
     t.boolean "vehicle_income"
   end
 
+  create_table "snap_supplements", force: :cascade do |t|
```
hartsick: Is this leftover from the sketch I was doing? not finding anywhere where we're using it, may just be leftover in a local DB

PR #709, opened 2018-03-20
Apply for healthcare

hartsick:
Allows user to select who in their household is applying for healthcare, and defaults them to applying for healthcare.

Also deletes some integration tests that seemed duplicative.

screen shot 2018-03-20 at 11 28 04 am

### DIFF
--- a/app/models/common_application.rb
+++ b/app/models/common_application.rb
@@ -20,6 +20,13 @@ class CommonApplication < ApplicationRecord
     class_name: "HouseholdMember",
     foreign_key: "common_application_id"
 
+  has_many :healthcare_applying_members,
+    -> {
+      requesting_healthcare
+    },
+    class_name: "HouseholdMember",
+    foreign_key: "common_application_id"
+
   has_many :snap_household_members,
     -> {
       requesting_food.buy_and_prepare_food_together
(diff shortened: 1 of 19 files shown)


app/models/common_application.rb
```diff
@@ -20,6 +20,13 @@ class CommonApplication < ApplicationRecord
...
     foreign_key: "common_application_id"
 
+  has_many :healthcare_applying_members,
+    -> {
+      requesting_healthcare
+    },
+    class_name: "HouseholdMember",
+    foreign_key: "common_application_id"
+
   has_many :snap_household_members,
```
luigi: I wonder if we should change `snap_applying_members` and `snap_household_members` to something else. Like `food_applying_members` and `food_household_members`? I like the general-ness of referring to it as `healthcare` rather then `medicaid`. Likewise `household_member`, I don't like that it's `requesting_food` when it's really `requesting_food_assistance`. But `food_assistance_applying_members` is just way too long.

hartsick: yeah I like food_applying_members / food_household_members, too. 'healthcare' is really 'healthcare_coverage', so the shortening scheme would be consistent. gonna incorporate these changes into the next PR

PR #715, opened 2018-03-22
Review, add, and remove members from tax household

luigi:
slack - code for america 2018-03-22 15-51-45
slack - code for america 2018-03-22 15-52-03
pdf 2018-03-22 15-53-00

Married filing separately:

image

image

https://www.pivotaltracker.com/story/show/155716708
https://www.pivotaltracker.com/story/show/155716923

PR #717, opened 2018-03-23
Integrated/auto tab date

hartsick:
When inputting numbers in our new DOB field, we should automatically tab to the next input after the required number of characters has been inputted. 

Based off of https://github.com/codeforamerica/michigan-benefits/pull/715 to make sure the birthday field on the new add a tax member page is included.

mar-22-2018 17-08-09


app/assets/javascripts/application.js
```diff
@@ -103,10 +103,27 @@ var yesNoEnumButtons = (function () {
   }
 })()
 
+var autoAdvanceTelInputs = (function() {
+  var autoAdvance = {
+    init: function () {
+      $(".form-group > input[type='tel']").keyup(function () {
+        if (this.value.length == this.maxLength) {
```
luigi: Small note that I prefer `===` for strict equality, since `==` does type conversion in  sometimes surprising ways.

PR #721, opened 2018-03-26
Make “relationship” a required field for household members.

luigi:
screenshot

- Adds a `skip_relationship_validation?` method for the `IntroduceYourselfController`.

[#156283599]


luigi:
@hartsick Good idea. We have only sparingly used hidden form fields so far in the codebase, so I was unsure of starting that pattern. But it does mean that we can remove the bits about skipping validation. Implemented in the latest commit.

PR #722, opened 2018-03-27
Integrated/married

hartsick:
Does a couple of feature things:

1) If user has added a member and that member is their spouse, mark them as married (on `household_member`) and as someone in the household being married (on `navigator`).
1) *In single-member household*: asks user if they're married
1) *In multi-member household with no known spouses*: asks user if anyone in household is married.
1) *In multi-member household with known married people*: asks user who is married. On this screen, they can't deselect members that have indicated they're married by relation "spouse", and they must select at least one married person.

Commits have a decent descriptions of other things done along the way.

screen shot 2018-03-26 at 5 51 36 pm
screen shot 2018-03-26 at 6 13 15 pm
screen shot 2018-03-26 at 5 51 51 pm

There ended up being a good amount of logical complexity in here. I started to feel the pain of two sources of truth (navigator's `anyone_married` vs member's `married`), but ultimately think they serve different purposes in this PR.

I'd still like to refactor the 'update' logic for multiple members—will tag a note with the relevant lines to get your thoughts. We had talked about refactoring this the next time it came up and I wanted to do so, but I ran out of time and wanted to get this up because my calendar is packed tomorrow. Some clear patterns have emerged, though!

### DIFF
--- a/app/controllers/integrated/remove_household_member_controller.rb
+++ b/app/controllers/integrated/remove_household_member_controller.rb
@@ -3,6 +3,10 @@ module Integrated
     def update_models
       if member.is_spouse?
         current_application.primary_member.update!(married: "no")
+
+        if (current_application.members - [member]).none?(&:married_yes?)
+          current_application.navigator.update!(anyone_married: false)
+        end
       end
 
       flash[:notice] = if member && current_application.members.delete(member)
--- a/app/controllers/integrated/remove_household_member_controller.rb
+++ b/app/controllers/integrated/remove_household_member_controller.rb
@@ -1,6 +1,10 @@
 module Integrated
   class RemoveHouseholdMemberController < RemoveMemberController
     def update_models
+      if member.is_spouse?
+        current_application.primary_member.update!(married: "no")
+      end
+
       flash[:notice] = if member && current_application.members.delete(member)
                          "Removed the household member."
                        else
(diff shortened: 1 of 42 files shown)


app/controllers/integrated/who_is_married_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    def self.skip?(application)
+      return true if application.single_member_household?
+      return true unless application.navigator.anyone_married?
+    end
+
+    def edit
+      @form = form_class.new(members: current_application.members)
+    end
+
+    def update
```
hartsick: Here's the relevant logic that I think we could tackle refactoring now (I can tackle between meetings tomorrow) or in the future. It's really similar to the update method in `DescribeTaxRelationshipsController`

app/controllers/integrated/remove_household_member_controller.rb
```diff
@@ -1,6 +1,14 @@
 module Integrated
   class RemoveHouseholdMemberController < RemoveMemberController
     def update_models
+      if member.is_spouse?
```
hartsick: just pointing out something I've been trying—updating models and reconciling the sources of truth wherever we can in the relevant controller, and keeping updates out of the skip.

luigi: Yup, I agree with this!

hartsick: sweet

PR #725, opened 2018-03-29
Reveal tax relationships based on relationship to member when adding a new member

luigi:
- Uses GCF’s Follow Up molecule.
- To be accessible, creates a distinct `tax_relationship_spouse` radio set which then gets set to `tax_relationship` in the controller.

https://www.pivotaltracker.com/story/show/156242462

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -24,7 +24,7 @@ var textWell = (function () {
       })
     }
   }
-})()
+})();
 
 var radioSelector = (function () {
   var rs = {
@@ -40,11 +40,11 @@ var radioSelector = (function () {
         })
       })
     }
-  }
+  };
   return {
     init: rs.init
   }
-})()
+})();
 
 var checkboxSelector = (function () {
   var cs = {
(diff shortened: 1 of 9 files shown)


app/controllers/integrated/add_tax_member_controller.rb
```diff
@@ -6,6 +6,18 @@ def additional_model_attributes
       }
     end
 
+    def update_models
+      member_data = combined_attributes(member_params)
+      if member_data[:relationship] == "spouse"
+        member_data[:married] = "yes"
+        current_application.navigator.update!(anyone_married: true)
+        current_application.primary_member.update!(married: "yes")
+        member_data[:tax_relationship] = member_data[:tax_relationship_spouse]
```
luigi: This is icky, but I couldn't think of an alternative way to keep the two checkbox sets distinct for accessibility purposes while still using the Follow Up pattern that GCF established.

app/assets/javascripts/application.js
```diff
@@ -113,17 +113,40 @@ var autoAdvanceTelInputs = (function() {
         }
       })
     }
-  }
+  };
```
hartsick: I appreciate these semicolons

app/controllers/integrated/add_tax_member_controller.rb
hartsick: yeah, that's tricky! I think it's easy to read, though, and the tests add some clarity.

PR #742, opened 2018-04-12
Reorder integrated flow

luigi:
[#156677052]

### DIFF
--- a/app/controllers/integrated/introduce_yourself_controller.rb
+++ b/app/controllers/integrated/introduce_yourself_controller.rb
@@ -12,20 +12,19 @@ module Integrated
                            month: member_data.delete(:birthday_month),
                            year: member_data.delete(:birthday_year),
       ))
-      if current_application.primary_member.is_a?(HouseholdMember)
+      if current_application.primary_member
         current_application.primary_member.update(member_data)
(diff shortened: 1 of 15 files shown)


app/controllers/integrated/introduce_yourself_controller.rb
```diff
@@ -14,20 +12,18 @@ def update_models
...
-        session[:current_application_id] = application.id
+        current_application.members.create(member_data)
       end
     end
 
     private
 
     def existing_attributes
-      if current_application
+      if current_application.primary_member.is_a?(HouseholdMember)
```
luigi: I feel iffy about this. Because `current_application.primary_member` will return a `NullMember`, this check (or something like it) is needed.

app/forms/form_navigation.rb
```diff
@@ -61,14 +56,20 @@ class FormNavigation
...
+      Integrated::ImmigrationInfoController,
+    ],
+    "Housing Expenses" => [
+
+    ],
+    "Dependent Expenses" => [
+
+    ],
+    "Student Loans" => [
+
```
luigi: I left some placeholders so we can see where we are and where we're going.

PR #756, opened 2018-04-20
Integrated app: Authorized Representative

luigi:
auth-rep

https://www.pivotaltracker.com/story/show/156390111


db/schema.rb
```diff
@@ -84,6 +84,9 @@
   end
 
   create_table "common_applications", force: :cascade do |t|
+    t.boolean "authorized_representative", default: false
```
hartsick: should we do this as an enum? right now it breaks the pattern of not stating defaults on anything but the navigator

luigi: Thanks for catching this. I've converted the field to an enum in 440e6cd.

PR #765, opened 2018-04-24
Integrated app: provide your social security number

luigi:


app/forms/provide_ssn_form.rb
```diff
@@ -0,0 +1,6 @@
+class ProvideSsnForm < Form
+  extend AutoStripAttributes
```
hartsick: Cool—I didn't realize we had AutoStripAttributes available! Should we use elsewhere, instead of redefining setters like we're doing on first_name/last_name? (see HouseholdMember for example)

PR #766, opened 2018-04-24
Ask applicant if they own property

hartsick:
Also:
* Modifies form_attributes to accept array (e.g. `application_attributes(dogs: [])`)
* Creates new form builder method to use with the checkbox + none pattern.
* Expands support in checkbox_set to accept a simple array, rather than to assign methods/attributes directly to form (e.g. will result in post like `["foo", "bar"]` rather than `{ foo: "1", bar: "0" }`. This is useful for updating simple array attributes on one model, rather than updating a collection of models

screen shot 2018-04-24 at 11 07 41 am

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -91,24 +91,15 @@ class FormsController < ApplicationController
   end
 
   def member_params
-    filter_params(form_class.member_attributes)
+    filtered_params(form_class.member_attributes)
   end
 
   def application_params
-    filter_params(form_class.application_attributes)
+    filtered_params(form_class.application_attributes)
   end
 
   def navigator_params
-    filter_params(form_class.navigator_attributes)
-  end
-
-  def filter_params(attributes)
-    return unless attributes.present?
-    form_params.select do |key, _|
-      attributes.any? do |attr|
-        key.match(/^#{attr}(\(.+\))?\z/) # Match single and multi-attribute keys
-      end
-    end
+    filtered_params(form_class.navigator_attributes)
   end
 
   def combined_birthday_fields(day: nil, month: nil, year: nil)
@@ -119,5 +110,9 @@ class FormsController < ApplicationController
     else
       {}
     end
+  end
+
+  def filtered_params(attrs)
+    form_params.slice(*Step::Attributes.new(attrs).to_s)
   end
 end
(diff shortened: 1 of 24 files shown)


hartsick:
Two pain points currently that I want to flag: how we designate attributes (member_attributes, application_attributes, navigator_attributes), and the form builder. Going to try to address some of latter in follow-up commit!

hartsick:
I didn't tackle a larger refactor of form builder, though I *so* want to, but did carve out a little method for the pattern we've been using (checkbox collection with 'none' option).

app/controllers/forms_controller.rb
```diff
@@ -91,24 +91,15 @@ def form_navigation
...
   def application_params
-    filter_params(form_class.application_attributes)
+    filtered_params(form_class.application_attributes)
   end
 
   def navigator_params
-    filter_params(form_class.navigator_attributes)
-  end
-
-  def filter_params(attributes)
```
hartsick: Note: we don't need this anymore, since we no longer send birthdays as multi-part forms

app/forms/form.rb
```diff
@@ -43,7 +43,9 @@ def form_attributes(attribute_names)
       attributes = (application_attributes + member_attributes + navigator_attributes)
       self.attribute_names = attributes
 
-      attribute_strings = attribute_names.map(&:to_s)
```
hartsick: This adds support for `member_attributes(foo: [])` rather than just `member_attributes(:foo)`

app/views/integrated/property/edit.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :header_title, "Assets" %>
+<% content_for :form_card_title, t(".title", count: current_application.members.count)  %>
+<% content_for :form_card_help, "You should include your primary home, if you own it. But don't worry — it won't be counted against you." %>
+
+<% content_for :form_card_body do %>
+    <%= fields_for(:form, @form, builder: MbFormBuilder) do |f| %>
+
+        <%= f.mb_checkbox_set_with_none(:properties,
```
hartsick: This example would post something like `{ form: { properties: ['foo', 'bar', '0'] } }` where 0 is an unchecked value, that we'd need to filter out. The `value_is_array: true` argument is the key to this.

PR #772, opened 2018-05-09
Require zip code be a 5-digit number

hartsick:
Previously, we only required that zip code be 5 digits. This makes sure that those 5 digits are all numbers!

Also adds a few tests and makes a small copy update.

screen shot 2018-05-08 at 6 22 59 pm

screen shot 2018-05-08 at 6 23 30 pm

[Fixes #156390099]
[Fixes #156390093]


hartsick:
no problem! looked like a fun little ticket and I had a little window

PR #773, opened 2018-05-11
Check for mailing zip if there is no residential zip

luigi:
[#156968837]

### DIFF
--- a/app/models/office_recipient.rb
+++ b/app/models/office_recipient.rb
@@ -24,13 +24,7 @@ class OfficeRecipient
   end
 
   def residential_or_mailing_zip
-    if benefit_application.respond_to?(:residential_zip)
-      benefit_application.residential_zip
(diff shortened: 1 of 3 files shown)


app/models/office_recipient.rb
```diff
@@ -19,7 +19,17 @@ def office
     if office_location.present?
       self.class.offices[office_location]
     else
-      self.class.find_by(zip: residential_address_zip)
+      self.class.find_by(zip: residential_or_mailing_zip)
+    end
+  end
+
+  def residential_or_mailing_zip
+    if benefit_application.respond_to?(:residential_zip)
```
hartsick: I think we don't use `residential_zip` anymore—seems to be legacy that no longer is meaningfully in the code base. can we remove entirely?

luigi:
@hartsick Thanks for pointing that out! Excision successful. :+1:

PR #775, opened 2018-05-16
Integrated flow: Additional Income Details (amounts)

luigi:
Includes the following refactorings:
- Rename `Income` to `AdditionalIncome`
- In form models, replace `set_foo_attributes` with `set_attributes_for :foo`
- In form controllers, replace `foo_params` with `params_for(:foo)`
- Introduce `MemberPerPageForm`, which performs validations on the current member

https://www.pivotaltracker.com/story/show/156390045

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -115,7 +115,7 @@ class FormsController < ApplicationController
   end
 
   def income_params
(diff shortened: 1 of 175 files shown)


luigi:
Merging now since this was a product of pairing, but feel free to leave suggestions @hartsick!

PR #779, opened 2018-05-17
Ask about dependent care expenses

hartsick:
screen shot 2018-05-17 at 4 05 29 pm

[Finishes #156961635]


app/controllers/concerns/expenses_details.rb
```diff
@@ -0,0 +1,35 @@
+module ExpensesDetails
+  extend ActiveSupport::Concern
```
luigi: Nice!

PR #793, opened 2018-05-29
Transition: Tell us about your household

bengolder:
Uses a partial to render a placeholder progress bar

screen shot 2018-05-29 at 2 58 41 pm
screen shot 2018-05-29 at 2 58 24 pm

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -73,6 +73,7 @@
 @import "michigan-benefits/atoms/form-elements";
 @import "michigan-benefits/molecules/form-molecules";
 @import "michigan-benefits/molecules/line-item";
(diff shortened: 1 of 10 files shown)


hartsick:
Just surfacing conversation Ben and I had! cc @luigi 

He's going to continue working on the partial on his branch to get it styled all pretty and reading well on a screenreader (goodbye, yellow and magenta), but he pushed this commit to unblock the other transition page stories.

When we pick up those stories, we should base off of this work and use the progress bar like so:
```
<%= render partial: "integrated/progress_bar", locals: {current_step: 1} %>
```
And we should not modify the partial at all.

bengolder:
Updated with proper styling

hartsick:
Look at THAT

PR #804, opened 2018-06-05
Ask applicants if there's anything else

bengolder:
screen shot 2018-06-05 at 12 15 45 pm
screen shot 2018-06-05 at 12 30 46 pm

### DIFF
--- a/app/controllers/integrated/anything_else_controller.rb
+++ b/app/controllers/integrated/anything_else_controller.rb
@@ -1,0 +1,11 @@
+module Integrated
+  class AnythingElseController < FormsController
+    def existing_attributes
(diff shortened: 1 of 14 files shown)


app/controllers/integrated/anything_else_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Integrated
+  class AnythingElseController < FormsController
+    def self.skip?(_application)
```
luigi: We should just not define this method rather than including it with `false`.

bengolder: fixed! Thanks!

PR #805, opened 2018-06-05
Remove questions for Medicaid-only flow 

luigi:
- Adds two new `SkipRule`s
- Adds member traits for `_food`
- Adds Medicaid-only spec

### DIFF
--- a/app/controllers/integrated/childcare_expenses_controller.rb
+++ b/app/controllers/integrated/childcare_expenses_controller.rb
@@ -1,10 +1,13 @@
 module Integrated
   class ChildcareExpensesController < FormsController
-    def self.skip?(application)
-      application.single_member_household?
-    end
-
     include SingleExpense
+
+    def self.skip_rule_sets(application)
+      [
+        SkipRules.must_be_applying_for_food_assistance(application),
+        SkipRules.multi_member_only(application),
+      ]
+    end
 
     def expense_type
       :childcare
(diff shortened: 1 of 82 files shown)


app/controllers/integrated/childcare_expenses_controller.rb
```diff
@@ -1,11 +1,14 @@
...
-    def self.skip?(application)
-      application.single_member_household?
-    end
-
     include SingleExpense
 
+    def self.skip_rule_sets(application)
+      [
+        SkipRules.must_be_applying_for_food_assistance(application),
+        SkipRules.multi_member_only(application),
```
bengolder: this one seems like it should be using the shovel since `multi_member_only` would already be added by `SingleExpense`

app/controllers/integrated/dependent_care_expenses_controller.rb
```diff
@@ -1,11 +1,14 @@
...
-    def self.skip?(application)
-      application.single_member_household?
-    end
-
     include SingleExpense
 
+    def self.skip_rule_sets(application)
+      [
+        SkipRules.must_be_applying_for_food_assistance(application),
+        SkipRules.multi_member_only(application),
```
bengolder: same as `ChildcareExpensesController`, it should probably just use `super <<`

app/controllers/integrated/childcare_expenses_controller.rb
luigi: I had to change `SingleExpense` to no longer always include the `multi_member_only` restriction. There are controllers that use it that should also show for single members.

PR #807, opened 2018-06-05
Calculate progress bar for transition pages

bengolder:
screen shot 2018-06-05 at 3 17 31 pm

### DIFF
--- a/app/controllers/integrated/finishing_up_controller.rb
+++ b/app/controllers/integrated/finishing_up_controller.rb
@@ -1,7 +1,7 @@
 module Integrated
-  class FinishingUpController < FormsController
-    def form_class
-      NullStep
+  class FinishingUpController < TransitionController
+    def current_step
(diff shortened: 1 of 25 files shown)


luigi:
@bengolder @hartsick 

Just to explain myself for the above commits: By default, CircleCI times out after 10 minutes of no output. Our test suite can now run over 10 minutes, and on CircleCI was piping all the output as a JUnit XML.

I first tried to increase the timeout from 10 to 20 minutes in the config `circle.yml`. Despite the [docs](https://circleci.com/docs/1.0/configuration/#modifiers) and some Googling, I couldn't get the config right.

So I removed the JUnit output directive (9682775b37132c8a66087e46eca47e70d6357e7f) so that suite would output those little dots as it ran, and avoid the timeout. I'm not sure how the JUnit output was used, if at all. That line was last modified in 2016(!).

PR #829, opened 2018-06-18
Add 2fa demo env

hartsick:
Currently, we can't set up 2FA for the demo environment because it has the same OTP name as production. This is because we use the RAILS_ENV to set the name, and they both are production.

This PR does a couple of things:
* Creates a new environment variable specifically for OTP name, since we'll need to have it different across every environment we might want to set up 2FA for: staging, demo, production
* Removes any reference to Rails.env.staging? with the intention to change any deployment environment to RAILS_ENV=production, so that we have more consistency between our staging and production environments (for testing)
* Stops 2FA from asking if we want to set it up in development 🗡 

This will require a few changes once merged/deployed to each env:
* Update `RAILS_ENV` to "production" on our staging env
* Add `OTP_ENV` as environment variable and set to either "staging", "demo", or "production", depending on the environment

This should not require any changes to 2FA codes if you already have them in an app like Google Authenticator, but it's hard to verify this is true before trying it out on staging.

Where a user's application is sent won't be affected by this, because it is controlled by the APP_RELEASE_STAGE variable.


hartsick:
@luigi yeah great point! I'll tackle this now

PR #847, opened 2018-06-28
Various copy changes

luigi:
[Finishes #158671305]

### DIFF
--- a/app/models/household_member.rb
+++ b/app/models/household_member.rb
@@ -92,13 +92,6 @@ class HouseholdMember < ApplicationRecord
     ["Other", "other_relation"],
   ].freeze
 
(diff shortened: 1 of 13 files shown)


app/models/household_member.rb
```diff
@@ -92,13 +92,6 @@ class HouseholdMember < ApplicationRecord
     ["Other", "other_relation"],
   ].freeze
 
-  TAX_RELATIONSHIP_LABELS_AND_KEYS = [
```
hartsick: huh, were these just not used anymore?

luigi: Yup, originally these were used in a dropdown, but we changed the UI later to use radio button sets that discern between spouses and dependent/other.

PR #848, opened 2018-06-29
Update Rails to 5.2

hartsick:
Oy, this was fun.

This PR:
* Updates gems
* Updates Rails from 5.1 -> 5.2
* Updates our factories and nested factory build strategy for failing tests

Here's more on that last point:

In v5, Rails began validating presence of a belongs_to association by default. This means any Member or HouseholdMember requires an application to be valid, etc. Because of this, many of our leaf factories (e.g. Member, Employment) were invalid. Also, because of Factory Bot's build strategy for nested factories (e.g. `create(:member, employments: build_list(:employment, 1))`), the associations were invalid—even though we're saying to build the nested association, Factory Bot will create the association (employments) before creating the parent (member).

For reasons I still don't understand, upgrading to Rails 5.2 brought this to light, even though it should have been present since Rails 5.1.

To fix the failing tests, I added default associations for belongs_to factories, and changed some of the way we build nested associations in the failing tests.


luigi:
The test suite is also really slow on Circle, so something must be up: https://circleci.com/gh/codeforamerica/michigan-benefits/3924

Gemfile
```diff
@@ -15,7 +15,7 @@ gem "bourbon", "~> 4.2.0" # to keep in sync with getcalfresh
 gem "browser"
 gem "capybara"
 gem "delayed_job_active_record"
-gem "delayed_job_web", "~> 1.4.2"
+gem "delayed_job_web", ">= 1.4.3"
```
luigi: Why the greater-than-equal-to operator rather than the pessimistic operator?

bin/rake
```diff
@@ -4,8 +4,7 @@ begin
 rescue LoadError => e
   raise unless e.message.include?('spring')
 end
-# frozen_string_literal: true
 
-require_relative "../config/boot"
-require "rake"
+require_relative '../config/boot'
+require 'rake'
```
luigi: I think Rubocop won't like the single quotes.

Gemfile
hartsick: It didn't seem like there was a reason to have a pessimistic operator—and I tried to remove that from a number of our gems that didn't need them, because I think we had accidentally let things get out of date—, but we want to make sure it doesn't regress to 1.4.1, which had the security vulnerability.

bin/rake
hartsick: I did run rubocop and it was surprisingly fine with this! I tried to keep whatever files were generated as-is to reduce future diffs, because the slight modifications that had been done to generated files in the past (removing comments, changing quote types) made it hard to see what was actually changed during the upgrade this time around.

hartsick:
I think I'm going to look into how to speed up our tests with any downtime I have today. Part of what was painful about this PR was a ton of failing tests, and the length of time it took to run the whole suite.

PR #850, opened 2018-06-29
Add missing SMTP config for production env

hartsick:
I missed requiring the SMTP config for production env when upgrading Rails to 5.2, which resulted in an error on Sentry in our staging env. This should fix that—we'll want to get in before we do the next production deploy.


hartsick:
I wonder why this test recently got really flakey? I'm going to merge this to get it onto master, and look into that test independently.

PR #853, opened 2018-07-02
Reduce test time

luigi:
This contains Rubocop fixes (are my and @hartsick's Rubocop config different?) and sets `default_max_wait_time` to as low as it can be to pass the JS-dependent feature specs.


hartsick:
Cool! CI is down to 11 minutes for a test run. Seems like a good start!

As for Rubocop—I had those changes locally, but thought they had been taken care of in another branch so didn't commit them, to avoid conflicts.

I've sometimes been skipping a full rake before pushing because local tests are so slow, which is why rubocop stuff is getting through more often—hoping this helps reduce time for specs locally, so it's not so painful.

PR #867, opened 2018-08-16
Update success page

luigi:
application submitted 2018-08-16 16-09-53

application submitted 2018-08-16 16-10-16

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -237,6 +237,21 @@ var noneOfTheAbove = (function() {
   }
 })();
 
+var feedbackForm = (function() {
+  var feedback = {
+    init: function () {
+      $('body').on('ajax:success', function(event, xhr, status, error) {
+        var success = '<span class="radio-button__image emoji emoji--med emoji--blush">Thank you</span>' +
+          '<h3>Thanks for your feedback! It will help improve this application for people in the future.</h3>';
+        $('#feedback-form').html(success);
+      });
+    }
+  };
+  return {
+    init: feedback.init
+  }
+})();
+
 $(document).ready(function () {
   radioSelector.init();
   checkboxSelector.init();
@@ -247,4 +262,5 @@ $(document).ready(function () {
   followUpQuestion.init();
   incrementer.init();
   noneOfTheAbove.init();
+  feedbackForm.init();
 })
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -245,6 +245,19 @@ var feedbackForm = (function() {
           '<h3>Thanks for your feedback! It will help improve this application for people in the future.</h3>';
         $('#feedback-form').html(success);
       });
+      $('body').on('ajax:error', function(event, xhr, status, error) {
+        var feedbackRatingErrors = xhr.responseJSON.errors["feedback_rating"];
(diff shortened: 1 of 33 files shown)


app/views/integrated/application_submitted/edit.html.erb
```diff
@@ -76,8 +74,21 @@
         </div>
       </div>
     </div>
+    <%= form_for @feedback_form, url: feedback_path, builder: MbFormBuilder, remote: true do |f| %>
```
luigi: Using `form_with` didn't seem to AJAX-ify the form, but passing `remote: true` to `form_for` worked correctly.

app/assets/javascripts/application.js
```diff
@@ -237,6 +237,21 @@ var noneOfTheAbove = (function() {
   }
 })();
 
+var feedbackForm = (function() {
+  var feedback = {
+    init: function () {
+      $('body').on('ajax:success', function(event, xhr, status, error) {
+        var success = '<span class="radio-button__image emoji emoji--med emoji--blush">Thank you</span>' +
+          '<h3>Thanks for your feedback! It will help improve this application for people in the future.</h3>';
+        $('#feedback-form').html(success);
```
luigi: Fully replacing the HTML with JS.

spec/features/integrated_application/integrated_application_with_single_member_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.feature "Integrated application" do
-  scenario "with one member", :a11y do
+  scenario "with one member", :js do
```
luigi: Remove `:a11y` means this feature spec doesn't check for accessibility. I don't think that's ideal, but it's more ideal than making a duplicate spec that does.

hartsick: yeah not ideal, but I think better than the alternative

hartsick:
Unfortunately, I didn't end up finishing this—ran into a couple of snags:

I found out that the current way we're doing this doesn't allow a user to keyboard navigate to the fields. I think the culprit is the input[type="radio"] { display: none }. This came up when I was trying to make accessible error messages, which is a bit painful because we've never done this with Javascript before and our form builder methods aren't set up for doing this easily.

I looked at GCF's code for their feedback survey, and saw a few differences. They:
1. Don't validate a rating (so no errors, period).
1. Don't submit with JS, and instead re-render the page with a different message.
1. Can't access the emoji radio buttons via tab (not accessible).

I talked with Andrew about the last bullet, and it sounds like none of their radio buttons are accessible by tabbing; it's on their todos and they're interested in if we were able to come up with a pattern they could use.

So, for this story, I think we have a couple of options:
1. Move forward knowing it's inaccessible (with or without errors).
1. Use our JS approach, but work to make it accessible (change the way we render buttons and errors, if we move forward with errors)
1. Switch to re-rendering the HTML (maybe by having two forms). We could use our standard form builder to render errors if we were to move forward.

I'm personally pro #2 with removing any validation that we have on the feedback form object.

I'd love to hear your thoughts!

hartsick:
I'm going to close this PR, since it's been handled in https://github.com/codeforamerica/michigan-benefits/pull/868 & https://github.com/codeforamerica/michigan-benefits/pull/869

PR #869, opened 2018-08-20
Feedback form accessible

hartsick:
Renders an accessible radio button set and displays errors using our form builder.

I decided not to show a visible error if there's no application, since it seemed confusing to the user. Just fails silently.

Also fixes a related bug, where our radio button set didn't properly show when buttons are selected after an initial error submission (go to page, submit an invalid state, then try to select the button again to see this—the button won't be visibly selected, but it will submit the right value).


hartsick:
the tiniest of pixels

PR #872, opened 2018-08-29
Require that any feedback be entered, not just rating

hartsick:
If user enters a comment or a rating, feedback submission is considered valid.

Also includes some minor styling changes.

[Fixes #159703918]

screen shot 2018-08-29 at 2 46 29 pm


hartsick:
Just realized this is going to fail because of the CVE related to Selenium (when we run `bundle-audit`).  Going to see if we can remove selenium for now to address

hartsick:
Tackled webdriving PR (that moves selenium to test env only) as part of https://github.com/codeforamerica/michigan-benefits/pull/873. Checks will still fail, since we include selenium, but we can safely ignore them.

PR #937, opened 2018-11-16
Remove demo site Warning and About screens, replace with About screen as homepage

luigi:
We've decided to change how the demo site operates. Old flow:

1. Homepage with happy faces
2. Demo Warning screen
3. Demo About screen
4. On with the flow

The new simpler flow:

1. Demo About screen, with smaller warning
2. Homepage with happy faces
3. On with the flow

The first two steps are both the homepage. To make the new step 2 feel like a distinct screen, I made the homepage available at `/start` in addition to `/`, with different display behavior for each.

Here's the new first page:

image


wschaefer-cfa (approved):
🚢

